Issue #42, opened 2013-08-15
Startup of Eureka in local environment takes 3-5 minutes


jfenner:
When starting Eureka in a dev environment with no other Eureka servers available, the startup times takes between 3 and 5 minutes.  I was that setting the following parameters would bypass this:

eureka.enableSelfPreservation=false
eureka.registration.enabled=false

However, it still is taking 3-5 minutes.  Is there something else that needs to be set for this to work?

tbak:
To eliminate syncUp just set ```eureka.numberRegistrySyncRetries=0```. Once it is set, the server starts up instantly.
I propose to close this issue, as there was no follow up on it for some time, and there is alternative approach to achieve this goal.

PR #119, opened 2014-05-01
Instantly start up if there are no peers to sync with

bpollack:
Please feel free to reject this; it's as much an RFC on how to address #42 as anything else.  Instead of introducing new settings for instant startup in development environments, I'm instead simply bypassing the wait part of the cycle if there are no peers to sync with.  This *might* have negative implications in a production environment, but I'm struggling to come up with one; it only shows up if you launch Eurekas a second before you populate DNS or the like.  I believe it ought to be safe.

### DIFF
--- a/eureka-client/src/test/java/com/netflix/discovery/BackUpRegistryTest.java
+++ b/eureka-client/src/test/java/com/netflix/discovery/BackUpRegistryTest.java
@@ -8,7 +8,6 @@ import com.netflix.discovery.shared.Application;
 import com.netflix.discovery.shared.Applications;
 import org.junit.After;
 import org.junit.Assert;
-import org.junit.Before;
 import org.junit.Test;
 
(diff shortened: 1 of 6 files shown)


qiangdavidliu:
@pgkelley4 the associated issue for this pull (#42) has been closed and it seems the settled upon way to remove the initial start up wait is just to set property `eureka.numberRegistrySyncRetries=0`. If you still want to continue working on this PR that would be more than welcome. Otherwise, we can just close this one out.

pgkelley4:
I am fine with this being closed @qiangdavidliu. I do still think that it should be documented in the wiki if the code isn't changed though.

qiangdavidliu:
Thanks. Have added a paragraph for this in https://github.com/Netflix/eureka/wiki/Configuring-Eureka#configuration-1, under Configuring for local development.

PR #220, opened 2014-10-10
More consistent health check behavior

elandau:
HealthCheck status is currently tracked via a mutable master InstanceInfo which can be set manually via ApplicationInfoManager or automatically through HealthCheckHandler.  This design gives rise to several race conditions since there is a circular dependency in setting the status.

This PR changes the behavior of InstanceInfo so that its status is only set via the InstanceInfoReplicator.  The status for ApplicationInfoManager is tracked internally instead of through the master InstanceInfo.  Using this new method the health check is only consulted if ApplicationInfoManager.getInstanceStatus() is UP.  Otherwise the status in ApplicationInfoManager is used. 

There still exists the possibility for instance status info to be 'inconsistent' if client code sets the master InstanceInfo.setStatus() manually.  This is unavoidable given the current design without a more significant refactoring.


qiangdavidliu:
closing this for now.

PR #234, opened 2014-10-28
Provide support for SystemSocketFactory for https client connections

keithbranton:
Use SystemSocketFactory for https client connections if system property com.netflix.eureka.shouldSSLConnectionsUseSystemSocketFactory is set to
true


bitsofinfo:
@keithbranton  
Is this supposed to be working w/ the discovery client or no (per the previous comments)?

keithbranton:
@bitsofinfo It's been a long time and I haven't used this since I contributed it, but looking at the patch I think it should be working with the DiscoveryClient.

bitsofinfo:
@keithbranton k thanks, yeah it seems to be broken: https://github.com/Netflix/eureka/issues/812

PR #320, opened 2015-01-04
2.x Replacing SelfRegistrationService impls with new HealthService

qiangdavidliu:


qiangdavidliu:
The healthCheck is now a component resolver

PR #326, opened 2015-01-08
Build changes to gradle 2.2.1 and new netflixoss plugin

rspieldenner:


rspieldenner:
This will create jdk7 bytecode. We can add a sourceCompatibility line if you prefer jdk6.

I removed the ide lines because the nebula.netflixoss plugin auto applies those.

PR #330, opened 2015-01-10
2.x merge register update

qiangdavidliu:
Mainly fixes in AbstractClientHandler and BaseMessageConnection to ensure proper ordering of events.
Refactered server side registrationChannel with more clear state handling.
Basic new version of versioning for registration that can handle async inbound requests and registry updates. This should go away once we implement the proper multi-local-source version.


eureka-client/src/main/java/com/netflix/eureka2/client/registration/RegistrationHandlerImpl.java
```diff
@@ -59,7 +59,7 @@ public RegistrationHandlerImpl(ClientChannelFactory channelFactory) {
             logger.info("Instance: {} is not registered. Ignoring unregister", instanceInfo);
             return Observable.empty(); // Be more acceptable to errors from user as unregister for non-existent instance is a no-op.
         }
-        return registrationChannel.unregister().doOnTerminate(new Action0() {
```
tbak: What will happen now when there is an error during unregistration? Shouldn't we close the channel as well?

eureka-client/src/test/java/com/netflix/eureka2/client/channel/RegistrationChannelTest.java
```diff
@@ -0,0 +1,123 @@
...
+                onErrorLatch.countDown();
+            }
+
+            @Override
+            public void onNext(Void aVoid) {
+
+            }
+        });
+
+        assertTrue(onErrorLatch.await(10, TimeUnit.SECONDS));
```
tbak: Can we use here TestScheduler, and TestSubscriber and eliminate the latch?

eureka-core/src/main/java/com/netflix/eureka2/channel/AbstractClientChannel.java
```diff
@@ -23,25 +19,25 @@
...
+        singleConnection = client.connect()
+                .take(1)
+                .map(new Func1<MessageConnection, MessageConnection>() {
+                    @Override
+                    public MessageConnection call(MessageConnection serverConnection) {
+                        connectionIfConnected = serverConnection;
+                        return connectionIfConnected;
+                    }
+                })
+                .replay()
```
tbak: As this is ConnectableObservable, shouldn't we call connect on it? According to the documentation, this observable should not emit any items without this call.

eureka-core/src/main/java/com/netflix/eureka2/registry/SourcedEurekaRegistryImpl.java
```diff
@@ -160,7 +161,7 @@ public void call(MultiSourcedDataHolder.Status status) {
      * TODO: do we have to eagerly subscribe? This code is inefficient.
      */
     private static Observable<Boolean> subscribeToUpdateResult(Observable<MultiSourcedDataHolder.Status> status) {
-        final ReplaySubject<Boolean> result = ReplaySubject.create();
+        final AsyncSubject<Boolean> result = AsyncSubject.create();  // use an async subject as we only need the last result
```
tbak: Is it possible to have more than one item in this stream?

eureka-write-server/src/main/java/com/netflix/eureka2/server/channel/RegistrationChannelImpl.java
```diff
@@ -77,97 +112,145 @@ public void onNext(Void aVoid) {
...
+                        });
+                    }
+                });
     }
 
+    /**
+     * Cases:
+     * 1. channel state is Registered. Call unregister on the registry, and
+     *   1a. if successful, ack on the channel and close the channel regardless of the ack result
+     *   1b. if unsuccessful, send error on the transport. TODO should we optimize and close the channel here?
```
tbak: There is no concept of sending an error on the transport. I think we should close eagerly here.

eureka-client/src/main/java/com/netflix/eureka2/client/registration/RegistrationHandlerImpl.java
qiangdavidliu: It's preferable that we do retries for client side unregister failures instead of just walking away from the connection. Otherwise, the instance will be unregistered via the eviction mechanism which can be slow.

eureka-client/src/test/java/com/netflix/eureka2/client/channel/RegistrationChannelTest.java
qiangdavidliu: Not easily, there's a few layers of plumbing that we might need to do. The latch here executes very quickly as the underlying channels are mocked and just returns Observable.empty().

eureka-core/src/main/java/com/netflix/eureka2/channel/AbstractClientChannel.java
qiangdavidliu: the refCount() calls connect().

eureka-core/src/main/java/com/netflix/eureka2/registry/SourcedEurekaRegistryImpl.java
qiangdavidliu: No

eureka-write-server/src/main/java/com/netflix/eureka2/server/channel/RegistrationChannelImpl.java
qiangdavidliu: Currently, sendErrorOnTransport() closes the channel implicitly so this is the current behaviour. We should discuss whether we want this to be the standard behaviour on all the channels however.

PR #331, opened 2015-01-12
2.x index compaction, and fix concurrency issues

tbak:


eureka-core/src/main/java/com/netflix/eureka2/interests/InstanceInfoInitStateHolder.java
```diff
@@ -1,40 +1,76 @@
...
         }
     }
 
     @Override
     protected void addNotification(ChangeNotification<InstanceInfo> notification) {
-        notifications.add(notification);
+        String id = notification.getData().getId();
+        ChangeNotification<InstanceInfo> current = notificationMap.get(id);
+
+        ChangeNotification<InstanceInfo> updated = processNext(current, notification);
```
qiangdavidliu: This code assumes single threaded calls of addNotification(). Are we guaranteed that?

tbak: Updates are serialized and come from the registry. Also for concurrent updates, it is thread safe. The issue might be only when the same instance info is updated concurrently from two different threads. In such case we would end up with the problem we have in the registry. But fortunately this is not the case.

eureka-core/src/main/java/com/netflix/eureka2/interests/NotificationsSubject.java
```diff
@@ -101,6 +100,13 @@ public ResumeResult resume() {
         }
     }
 
+    protected void drainBuffer() {
+        ChangeNotification<T> nextPolled;
+        while ((nextPolled = notificationsWhenPaused.poll()) != null) { // drain pending queue.
+            notificationSubject.onNext(nextPolled); // Since pause flag is not yet unset, don't call this.onNext()
```
qiangdavidliu: This comment is outdated now that this is in a separate method. Are we only ever allowed to drainBuffer() as part of resume or if we are not currently paused? Can we add a comment here to details these restrictions?

PR #336, opened 2015-01-13
Removing EurekaRegistry interface

qiangdavidliu:
Also refactored Source and SourceMatcher to be more explicit in usage.


NiteshKant:
What was the motivation of removing this interface?

PR #345, opened 2015-01-16
Adding random lifecycle test suite

qiangdavidliu:


eureka2-integration/src/main/java/com/netflix/eureka2/integration/IntegrationTestClient.java
```diff
@@ -0,0 +1,139 @@
...
+    public List<ChangeNotification<InstanceInfo>> getExpectedLifecycle() {
+        return lifecycle;
+    }
+
+    public List<ChangeNotification<InstanceInfo>> playLifecycle() {
+        final ChangeNotification<InstanceInfo> expectedEnd = lifecycle.get(lifecycle.size() - 1);
+
+        final CountDownLatch expectedEndLatch = new CountDownLatch(1);
+        Subscription subscription = readClient.forApplication(appName).subscribe(
+                new SerializedSubscriber<>(
```
tbak: Why do we have to enforce Rx contract here? Notifications should be serialized, if not it is a bug.

eureka2-integration/src/main/java/com/netflix/eureka2/integration/IntegrationTestClient.java
```diff
@@ -0,0 +1,139 @@
...
+                            @Override
+                            public void onCompleted() {
+                            }
+
+                            @Override
+                            public void onError(Throwable e) {
+                            }
+
+                            @Override
+                            public void onNext(ChangeNotification<InstanceInfo> notification) {
```
tbak: We can use takeWhile + TestSubscriber or ExtTestSubscriber. The code is more compact this way.

eureka2-integration/src/main/java/com/netflix/eureka2/integration/IntegrationTestClient.java
```diff
@@ -0,0 +1,139 @@
...
+
+        for (ChangeNotification<InstanceInfo> notification : lifecycle) {
+            if (notification.getKind() == ChangeNotification.Kind.Delete) {
+                writeClient.unregister(notification.getData()).retry(3).subscribe();
+            } else {
+                writeClient.register(notification.getData()).retry(3).subscribe();
+            }
+
+            try {
+                Thread.sleep(gapWaitMs);
```
tbak: Is there a way to avoid explicit sleeps?

eureka2-integration/src/main/java/com/netflix/eureka2/integration/IntegrationTestClient.java
qiangdavidliu: I deliberately tried to avoid test code in the client so we can re-use it as a standalone elsewhere.

eureka2-integration/src/main/java/com/netflix/eureka2/integration/IntegrationTestClient.java
qiangdavidliu: We can do Observable.from(list).interval()

eureka2-integration/src/main/java/com/netflix/eureka2/integration/IntegrationTestClient.java
qiangdavidliu: Yeah not needed. I think this might have been autocomplete.

PR #346, opened 2015-01-16
Fixes for clean connection close in AmazonInfo when reading from the AWS

qiangdavidliu:
metadata url.


scotte:
Looks good, thanks for fixing that! :boat:

PR #347, opened 2015-01-16
2.x various fixes

qiangdavidliu:


eureka2-core/src/main/java/com/netflix/eureka2/utils/rx/RetryStrategyFunc.java
```diff
@@ -0,0 +1,48 @@
...
+import rx.functions.Func2;
+
+import java.util.concurrent.TimeUnit;
+
+/**
+ * A Func1 with retry options for use in .retryWhen(new RetryStrategyFunc())
+ *
+ * @author David Liu
+ */
+public class RetryStrategyFunc implements Func1<Observable<? extends Throwable>, Observable<Long>> {
```
tbak: Have you seen ExponentialBackoff or Retrys from ocelli? There are many standard algorithms implemented that we could reuse.

qiangdavidliu: The Ocelli version is nicer, but we don't want to have dependency on ocelli for all the packages that needs retry. It feels like this is something useful to put in a more basic layer (rxjava/rxnetty)

tbak: Is it visible in the API? If not, depending on ocelli is ok.

PR #349, opened 2015-01-20
2.x Eager cache within Retryable Interest and Registration channels

qiangdavidliu:


qiangdavidliu:
More changes TBD. Close for now.

PR #350, opened 2015-01-21
2.x Self-unregister of Eureka server during shutdown process.

tbak:
Self-unregister of Eureka server during shutdown process.
Additional unit/integration tests.


qiangdavidliu:
Looks good!

PR #351, opened 2015-01-22
2.x Extract API for metrics. Move from servo to spectator. WIP

tbak:
API is extracted and a project reorganized accordingly. A new no-op implementation is provided as a default.
Spectator binding is not complete yet and not verified.


eureka2-core/src/main/java/com/netflix/eureka2/channel/AbstractServiceChannel.java
```diff
@@ -60,4 +66,27 @@ public void onNext(T message) {
...
+                }
+            }
+            return true;
+        }
+        return false;
+    }
+
+    protected boolean moveToState(STATE to) {
+        if (state.get() == to) {
+            return false;
```
qiangdavidliu: What is the contract for the return of this method? If we are already at the state that we are trying to move to, it could be said that the moveToState is successful (i.e. true).

eureka2-metrics-spectator/src/main/java/com/netflix/eureka2/metric/SpectatorEurekaRegistryMetrics.java
```diff
@@ -0,0 +1,124 @@
...
+        updatesTotal = newCounter("updatesTotal");
+
+        unregistrationsLocal = newCounter("unregistrationsLocal");
+        unregistrationsReplicated = newCounter("unregistrationsReplicated");
+        unregistrationsTotal = newCounter("unregistrationsTotal");
+    }
+
+    @Override
+    public void incrementRegistrationCounter(String origin) {
+        switch (origin) {
```
qiangdavidliu: Why not just use the enum available in Source.Origin directly here?

eureka2-metrics-spectator/src/main/java/com/netflix/eureka2/metric/server/SpectatorBridgeChannelMetrics.java
```diff
@@ -0,0 +1,45 @@
...
+
+import com.netflix.eureka2.channel.BridgeChannel.STATE;
+import com.netflix.eureka2.metric.AbstractStateMachineMetrics;
+import com.netflix.spectator.api.ExtendedRegistry;
+
+/**
+ * @author Tomasz Bak
+ */
+public class SpectatorBridgeChannelMetrics extends AbstractStateMachineMetrics<STATE> implements BridgeChannelMetrics {
+//    private final LongGauge totalCount;
```
qiangdavidliu: Why commented out?

eureka2-metrics-spectator/src/main/java/com/netflix/eureka2/metric/server/SpectatorReplicationChannelMetrics.java
```diff
@@ -14,17 +14,18 @@
...
 /**
  * @author Tomasz Bak
  */
-public class ReplicationChannelMetrics extends AbstractStateMachineMetrics<STATES> {
+public class SpectatorReplicationChannelMetrics extends AbstractStateMachineMetrics<STATE> implements ReplicationChannelMetrics {
 
-    public ReplicationChannelMetrics() {
-        super("server", STATES.class);
+    public SpectatorReplicationChannelMetrics(ExtendedRegistry registry, String server) {
+        super(registry, "server", STATE.class);
```
qiangdavidliu: Should use the server string instead of "server" here.

eureka2-core/src/main/java/com/netflix/eureka2/channel/AbstractServiceChannel.java
```diff
@@ -60,4 +66,27 @@ public void onNext(T message) {
             }
         });
     }
+
+    protected boolean moveToState(STATE from, STATE to) {
```
qiangdavidliu: How does this work with the (now) overridden moveToState in the specific channelImpls?

eureka2-core/src/main/java/com/netflix/eureka2/channel/AbstractServiceChannel.java
tbak: In this case there would be no point to return the status, as it always would be true. Now we return true only when the state changes. This is useful when we move to Close state, which guards cleanup code. We want to run it only once.

eureka2-metrics-spectator/src/main/java/com/netflix/eureka2/metric/SpectatorEurekaRegistryMetrics.java
tbak: Originally metrics API were in a separate package and using Origin directly was not possible. I will fix that.

eureka2-metrics-spectator/src/main/java/com/netflix/eureka2/metric/server/SpectatorBridgeChannelMetrics.java
tbak: It is still work in progress. I need to fix/test all spectator metrics classes.

eureka2-metrics-spectator/src/main/java/com/netflix/eureka2/metric/server/SpectatorReplicationChannelMetrics.java
tbak: I need to cleanup metric names. Probably centralizing all these names in one class would be best.

eureka2-core/src/main/java/com/netflix/eureka2/channel/AbstractServiceChannel.java
tbak: Thanks for spotting this. It is a bug. I wanted to provide common implementation that both handles state transitions and related state metrics. I need to remove old implementation.

PR #354, opened 2015-01-25
2.x startup/shutdown integ test fix

qiangdavidliu:


eureka2-integration/src/test/java/com/netflix/eureka2/integration/startup/FileBasedServerStartupAndShutdownIntegrationTest.java
```diff
@@ -0,0 +1,23 @@
...
+import org.junit.Test;
+
+/**
+ * Since this test relies on Archaius which has some global settings, we can't have different tests for both read
+ * and write servers under the same JVM. As the tests are testing startup and shutdown behaviour that is common to
+ * both (within AbstractEurekaServer), we just pick one of the two server types (write servers) to test this.
+ *
+ * @author David Liu
+ */
+public class FileBasedServerStartupAndShutdownIntegrationTest extends WriteServerStartupAndShutdownIntegrationTest {
```
tbak: Why this test is extract in a separate class, and cannot be part of WriteServerStartupAndShutdownIntegrationTest?

qiangdavidliu: It's in a separate class for clearer organisation. If we added it to the write test class chances are someone in the future might see the "missing" complementary test in the read class and add a read version back, which will then get us back to the same problem as now.

tbak: But we need this test for read server as well. We should be able to
configure write cluster addresses in read server properties file, and
bootstrap it accordingly.

PR #355, opened 2015-01-25
2.x Metrics update + metrics unit tests

tbak:


eureka2-server/src/main/java/com/netflix/eureka2/server/channel/InterestChannelImpl.java
```diff
@@ -65,13 +64,11 @@ public void call(Object message) {
...
                         case Idle:
-                            state.set(STATE.Open);
-                            metrics.stateTransition(STATE.Idle, STATE.Open);
+                            moveToState(STATE.Idle, STATE.Open);
 
                             sendSnapshot(((SnapshotRegistration) message).getInterests());
 
-                            state.set(STATE.Closed);
-                            metrics.stateTransition(STATE.Idle, STATE.Closed);
+                            moveToState(STATE.Idle);
```
qiangdavidliu: Back to Idle or to Closed?

eureka2-write-server/src/main/java/com/netflix/eureka2/server/channel/SenderReplicationChannel.java
```diff
@@ -43,11 +43,8 @@ public SenderReplicationChannel(TransportClient client, ReplicationChannelMetric
 
     @Override
     public Observable<ReplicationHelloReply> hello(final ReplicationHello hello) {
-        if (state.get() == STATE.Closed) {
-            return Observable.error(CHANNEL_CLOSED_EXCEPTION);
-        }
-        if (!state.compareAndSet(STATE.Idle, STATE.Handshake)) {
-            return Observable.error(HANDSHAKE_FINISHED_EXCEPTION);
+        if (!moveToState(STATE.Idle, STATE.Handshake)) {
```
qiangdavidliu: Is it possible for the handshake processing to fail? If so, this channel will be stuck in state handshake and have no transition out. We should either let handshake -> handshake to be valid or better yet, close this channel if the handshake fails so we can retry.

eureka2-core/src/main/java/com/netflix/eureka2/channel/AbstractServiceChannel.java
```diff
@@ -76,6 +76,8 @@ protected boolean moveToState(STATE from, STATE to) {
                     metrics.incrementStateCounter(to);
```
qiangdavidliu: With this line replacing all the metrics.incrementStateCount(Idle) lines in the concrete classes' constructors and the metric less moveToState(STATE to) option, can we possibly lose metrics when channels are created and are then immediately closed via a call to moveToState(Closed)?

eureka2-core/src/main/java/com/netflix/eureka2/transport/base/BaseMessageConnection.java
```diff
@@ -174,6 +177,11 @@ public Boolean call(Object message) {
             public void call(Object o) {
                 metrics.incrementIncomingMessageCounter(o.getClass(), 1);
             }
+        }).doOnTerminate(new Action0() {
+            @Override
+            public void call() {
+                shutdown();
```
qiangdavidliu: Do we have verification tests to ensure that onComplete and onError terminates will trigger the proper shutdown and send the correct onComplete or onError on the lifecycle observable?

qiangdavidliu:
Much kudos on all the metrics tests!

eureka2-server/src/main/java/com/netflix/eureka2/server/channel/InterestChannelImpl.java
tbak: Closed. I have fixed that.

eureka2-write-server/src/main/java/com/netflix/eureka2/server/channel/SenderReplicationChannel.java
tbak: The state transition is driven by the client. This guard enforces single handshake that happens in the very beginning. We could add explicit channel close in case of failure, but we do not do this on the client side, and let the client handle the errors. We drift towards more "standardized" state machine in the channel. Lets revisit how we should handle it, including lifecycle/shutdown handling.

eureka2-core/src/main/java/com/netflix/eureka2/channel/AbstractServiceChannel.java
tbak: I have deliberately removed Idle state tracking, as it has unclear semantic. There was no connection attempt yet, just pure object construction. It is easy to put it back if we see value of it.

eureka2-core/src/main/java/com/netflix/eureka2/transport/base/BaseMessageConnection.java
tbak: I have added this action, to handle remote endpoint disconnects. Handling of onComplete and onError is up to the client.
I will add however a test that check, that in case of remote server disconnect, we detect if immediately and close the channel.

PR #356, opened 2015-01-25
2.x Eureka API update (Eureka class cleanup).

tbak:


qiangdavidliu:
Ship it. 

On an unrelated note, we use the terms interest and discovery for the same meaning interchangeably in our code. We should do a sweep through at the end and stick to one or the other.

tbak:
I agree, we should stick to one term.

PR #379, opened 2015-01-30
2.x Use TestScheduler in InterestChannelImplTest for better control over the testing process.

tbak:


qiangdavidliu:
Is this the fix for the intermittent interestChannel test failures?

tbak:
I cannot replicate this error in my dev environment. With this change I want to narrow down the possible culprits.

PR #382, opened 2015-01-31
2.x Client channel major refactor

qiangdavidliu:
- more consistent channel lifecycles and onComplete/onError
- old Retryable channels are not RetryableConnections that can be .retry()-ed on the lifecycle.
  for each retry, new channels are created and the old channel shutdown
- registrationHandler now takes a stream of InstanceInfos
- registrationTracker that bridges between the old register/unregister API and the new
  stream based API. EurekaClient API change to the new form is not done as part of this pull
  to minimize the change list. It will change in a subsequent pull.


eureka2-server/src/main/java/com/netflix/eureka2/server/service/SelfRegistrationService.java
```diff
@@ -77,8 +77,10 @@ public void call() {
                         public void call(Throwable throwable) {
                             logger.warn("Error during self unregistration process: {}", latestSelfInfo);
                         }
-                    }).materialize().toBlocking().firstOrDefault(null);
+                    }).materialize().toBlocking().firstOrDefault(null);  // FIXME why materialize and block here?
```
tbak: We cannot/shouldn't proceed with shutdown until unregister is completed. We cannot change shutdown() to return observable, as it is a pre-destroy hook.

eureka2-server/src/main/java/com/netflix/eureka2/server/service/SelfRegistrationService.java
```diff
@@ -86,6 +88,8 @@ public void call(Throwable throwable) {
         return resolver.resolve();
     }
 
+    public abstract void cleanUpResources();
```
tbak: Why it is not in shutdown hook override?

eureka2-server/src/test/java/com/netflix/eureka2/server/channel/AbstractHandlerChannelTest.java
```diff
@@ -42,15 +42,15 @@ public void setUp() {
     public void testCloseChannelOnSendError() throws Exception {
         when(transport.submit(anyObject())).thenReturn(Observable.<Void>error(new Exception("msg")));
 
-        final CountDownLatch onCompletedLatch = new CountDownLatch(1);
+        final CountDownLatch onErrorLatch = new CountDownLatch(1);
```
tbak: We can use TestSubscriber for that purpose.

eureka2-client/src/main/java/com/netflix/eureka2/client/registration/RegistrationTracker.java
```diff
@@ -0,0 +1,107 @@
...
+        protected RegistrationBridge(BehaviorSubject<InstanceInfo> subject) {
+            this.subject = subject;
+            this.response = handler.register(subject.asObservable());
+            this.subscriber = new NoOpSubscriber<>();
+
+            // it's fine to eager subscribe here, the actual channel creation is not triggered until the first onNext
+            response.subscribe(subscriber);
+        }
+
+        public void register(InstanceInfo instanceInfo) {
```
tbak: This can be called concurrently, so we need thread safe subject.

eureka2-client/src/main/java/com/netflix/eureka2/client/registration/RegistrationTracker.java
```diff
@@ -0,0 +1,107 @@
...
+/**
+ * TODO deprecate me
+ *
+ * This is a bridging class that bridges the registration behaviour of the current eurekaClient (individual register()
+ * and unregister() calls) to the new behaviour in the handler (a stream of instanceInfos). Going forward, this class
+ * should go away as we change over the client api.
+ *
+ * @author David Liu
+ */
+public class RegistrationTracker {
```
tbak: In the current model we never onError to the client. It is fine, assuming we will refactor client API, and expose different result type.

eureka2-client/src/main/java/com/netflix/eureka2/client/channel/RetryableConnectionFactory.java
```diff
@@ -0,0 +1,128 @@
...
+
+                            @Override
+                            public void onNext(Void aVoid) {
+                            }
+                        });
+
+                        return channel.asLifecycleObservable();
+                    }
+                })
+                .flatMap(new Func1<Observable<Void>, Observable<Void>>() {
```
tbak: Why is it here?

eureka2-client/src/main/java/com/netflix/eureka2/client/interest/InterestHandlerImpl.java
```diff
@@ -17,33 +19,106 @@
...
+
+        Observable<Void> appendInterest = Observable.create(new Observable.OnSubscribe<Void>() {
+            @Override
+            public void call(Subscriber<? super Void> subscriber) {
+                interestTracker.appendInterest(interest);
+                subscriber.onCompleted();
+            }
+        });
+
+        Observable toReturn = appendInterest
```
tbak: Should we merge this with lifecycle, so in case retries fail, ultimately we notify the client about an error. The only other way to do that would be to shutdown registry with an error, but we do not want to do that.

eureka2-client/src/test/java/com/netflix/eureka2/client/channel/RetryableConnectionFactoryTest.java
```diff
@@ -0,0 +1,468 @@
...
+            return channelLifecycle.asObservable();
+        }
+
+        @Override
+        public String toString() {
+            return "channel:" + id;
+        }
+    }
+
+    static class TestChannelFactory implements ChannelFactory<TestServiceChannel> {
```
tbak: We have TestChannelFactory in test-utils. Cannot we reuse or extend that one?

eureka2-server/src/main/java/com/netflix/eureka2/server/service/SelfRegistrationService.java
qiangdavidliu: I see. But shouldn't we let unregisters be best effort? Server side we are ok with best effort unregisters, and we would not want a bad unregister to block server shutdown.

eureka2-server/src/main/java/com/netflix/eureka2/server/service/SelfRegistrationService.java
qiangdavidliu: the typical pattern for shutdown overrides are 
void shutdown() {
   // do child shutdown
   super.shutdown().
}

In this case we want the cleanUp action to happen after the super.shutdown(), so this is a safer pattern.

eureka2-client/src/main/java/com/netflix/eureka2/client/registration/RegistrationTracker.java
qiangdavidliu: fixed

eureka2-server/src/test/java/com/netflix/eureka2/server/channel/AbstractHandlerChannelTest.java
qiangdavidliu: changed

eureka2-client/src/main/java/com/netflix/eureka2/client/registration/RegistrationTracker.java
qiangdavidliu: Yep. Going forward the RegistrationResult object should be what will be exposed to users.

eureka2-client/src/main/java/com/netflix/eureka2/client/channel/RetryableConnectionFactory.java
qiangdavidliu: The combineLatest does not compose without this.

eureka2-client/src/main/java/com/netflix/eureka2/client/interest/InterestHandlerImpl.java
qiangdavidliu: For the interest side, since all interest subscriptions share the same underlying channel, it makes more sense to have final channel failures propagate the error to the registry, and then have the registry onError the clients.

eureka2-client/src/test/java/com/netflix/eureka2/client/channel/RetryableConnectionFactoryTest.java
qiangdavidliu: Ahh yes. This TestChannelFactory was created before that one. Indeed they can be merged. Will do that in a separate pull.

PR #388, opened 2015-02-05
2.x RetryableConnection for SenderReplication

qiangdavidliu:
Using RetryableConnection for sender replication.
Connecting connection lifecycle to channel lifecycle in AbstractClientChannel.


eureka2-core/src/main/java/com/netflix/eureka2/connection/RetryableConnectionFactory.java
```diff
@@ -14,50 +14,55 @@
...
     public RetryableConnectionFactory(final ChannelFactory<CHANNEL> channelFactory) {
         this.channelFactory = channelFactory;
     }
 
     /**
+     * @param executeOnChannel a func1 that describes a nullary call for the channel
+     * @return a {@link RetryableConnection} that contains several observables. This lifecycle observable provided
+     * can be retried on.
+     */
+    public RetryableConnection<CHANNEL> nullaryConnection(final Func1<CHANNEL, Observable<Void>> executeOnChannel) {
```
tbak: I understand that this is type of connection where a client does not provide any operations. Can we pickup a better name for it, and also for unaryConnection? Some documentation would be useful, as well.

eureka2-core/src/main/java/com/netflix/eureka2/connection/RetryableConnectionFactory.java
```diff
@@ -124,5 +129,23 @@ public void call(CHANNEL channel) {
         );
     }
 
-    protected abstract Observable<Void> executeOnChannel(CHANNEL channel, OP op);
```
tbak: I have added in my branch 
```java
protected abstract Observable<INPUT> connectToChannelInput(CHANNEL channel);
```
How can we expose channel input with this implementation?

eureka2-write-server/src/test/java/com/netflix/eureka2/server/service/replication/ReplicationHandlerTest.java
```diff
@@ -0,0 +1,414 @@
...
+        MessageConnection messageConnection = newMockMessageConnection();
+        when(messageConnection.submit(anyObject())).thenReturn(Observable.<Void>empty());
+        when(messageConnection.incoming()).thenReturn(Observable.<Object>just(HELLO_REPLY));
+
+        TransportClient transportClient = mock(TransportClient.class);
+        when(transportClient.connect()).thenReturn(Observable.just(messageConnection));
+
+        final ReplicationChannel channel = new SenderReplicationChannel(transportClient, mock(ReplicationChannelMetrics.class));
+
+        Observable.empty().delay(failAfterMillis, TimeUnit.MILLISECONDS).subscribe(new Subscriber<Object>() {
```
tbak: Is it real time, or from TestScheduler? Why not schedule this as a task?

eureka2-core/src/main/java/com/netflix/eureka2/connection/RetryableConnectionFactory.java
qiangdavidliu: nullary and unary are pretty standard mathematical names for operand sizes: http://en.wikipedia.org/wiki/Arity.

eureka2-write-server/src/test/java/com/netflix/eureka2/server/service/replication/ReplicationHandlerTest.java
qiangdavidliu: This is intended as a realtime task as it need to sync with some other steps in the test. It'll be difficult to do as a testScheduler unfortunately.

eureka2-core/src/main/java/com/netflix/eureka2/connection/RetryableConnectionFactory.java
qiangdavidliu: the abstract executeOnChannel() functionality is now provide as a func1/func2 that are passed in, so RetryableConnectionFactory is no longer abstract.

eureka2-core/src/main/java/com/netflix/eureka2/connection/RetryableConnectionFactory.java
tbak: I understand that. But in nullaryConnection method name, it implies that the connection itself is nullary, and we actually mean here the operator. Maybe connectionWithNullaryOp?

eureka2-core/src/main/java/com/netflix/eureka2/connection/RetryableConnectionFactory.java
tbak: That does not answer the question, how we can now pass channel input. Shall I keep the abstract method or add another function argument. For consistency the latter is better, but the number of arguments becomes too big. We might be force to introduce a builder for more clarity.

PR #404, opened 2015-02-07
2.x Adding ability to configure some transport properties

qiangdavidliu:


eureka2-core/src/main/java/com/netflix/eureka2/config/BasicEurekaRegistryConfig.java
```diff
@@ -9,32 +11,36 @@
...
+    public static final String EVICTION_STRATEGY_TYPE = StrategyType.PercentageDrop.name();
+    public static final String EVICTION_STRATEGY_VALUE = "20";
 
-    private String evictionTimeoutMs = System.getProperty("eureka.registry.evictionTimeoutMs", EVICTION_TIMEOUT_MS);
-    private String evictionStrategyType = System.getProperty("eureka.registry.evictionStrategy.type", EVICTION_STRATEGY_TYPE);
-    private String evictionStrategyValue = System.getProperty("eureka.registry.evictionStrategy.value", EVICTION_STRATEGY_VALUE);
+    private Long evictionTimeoutMs = Long.parseLong(
+            System.getProperty(evictionTimeoutMsName, EVICTION_TIMEOUT_MS)
+    );
+    private StrategyType evictionStrategyType = StrategyType.valueOf(
```
tbak: We should handle more gracefully errors here, as exceptions are very cryptic when class initialization fails. If property value cannot be parsed, we should log a warning and use a default.

eureka2-core/src/main/java/com/netflix/eureka2/config/ConfigNameStrings.java
```diff
@@ -0,0 +1,25 @@
+package com.netflix.eureka2.config;
+
+/**
+ * use camel case for better readability
+ *
+ * @author David Liu
+ */
+public final class ConfigNameStrings {
```
tbak: Or just ConfigurationNames?

eureka2-core/src/main/java/com/netflix/eureka2/config/ConfigNameStrings.java
```diff
@@ -0,0 +1,25 @@
...
+    private static final String EUREKA_PREFIX = "eurekas.";
+
+    public final class Transport {
```
tbak: TansportNames? So we can do static import, and have a proper context in the class name.

eureka2-core/src/main/java/com/netflix/eureka2/config/EurekaTransportConfig.java
```diff
@@ -0,0 +1,13 @@
+package com.netflix.eureka2.config;
+
+import com.netflix.eureka2.transport.EurekaTransports;
+
+/**
+ * @author David Liu
+ */
+public interface EurekaTransportConfig {
```
tbak: We need a builder here, that plays well with EurekaBuilder.
For example:
```java
newEurekaBuilder().
    .withTransport(newTransportConfig().withCodec(Codec.Avro))
   .withXXX
```
EurekaBuilder should accept both EurekaTransportConfig and EurekaTransportConfigBuilder for client convenience.

eureka2-external-server/src/main/java/com/netflix/eureka2/server/EurekaClientProvider.java
```diff
@@ -69,7 +69,6 @@ public Integer call(ServerBootstrap server) {
                 }
         );
         return Eureka.newClientBuilder(discoveryResolver, registrationResolver)
-                .withCodec(config.getCodec())
```
tbak: Can we still run EurekaRead server with arbitrary codec? JSON is very useful for debugging purposes.

eureka2-core/src/main/java/com/netflix/eureka2/config/BasicEurekaRegistryConfig.java
qiangdavidliu: Will fix.

eureka2-core/src/main/java/com/netflix/eureka2/config/ConfigNameStrings.java
qiangdavidliu: Good suggestion

eureka2-core/src/main/java/com/netflix/eureka2/config/EurekaTransportConfig.java
qiangdavidliu: There are only two variables right now. We can add a builder if this config grows in the future.

eureka2-external-server/src/main/java/com/netflix/eureka2/server/EurekaClientProvider.java
qiangdavidliu: We should still be able to do so, the config just provides a way to specify the codec and defaults to avro, it does not force it to avro.

eureka2-core/src/main/java/com/netflix/eureka2/config/EurekaTransportConfig.java
tbak: It is more about client API usability. With builder we can have more fluent
object construction.

PR #405, opened 2015-02-08
Fix InstanceInfo.Builder.setLeaseInstance

shaie:
InstanceInfo.Builder.setLeaseInstance broke the builder pattern by declaring to return void. This simple commit fixes it to return Builder.


qiangdavidliu:
Thanks for the fix.

PR #406, opened 2015-02-08
Don't use String.toUpperCase()

shaie:
InstanceInfo uses String.toUpperCase() without passing a Locale. This
is very dangerous as it relies on the default locale which might not
match the server's Locale. It is one of the APIs that are discouraged
for use, and it is better if it always passes a Locale.ROOT.


qiangdavidliu:
Make sense, thanks for the pull.

PR #407, opened 2015-02-08
Check for right logging level

shaie:
The code checked if isInfoEnabled() but then called logger.debug().


qiangdavidliu:
thanks for the fix.

Issue #402, opened 2015-02-06
Eureka 1.x - add additional guard in InstanceInfo.getHealthCheckUrls


tbak:
Currently this method will return Set object with null values in it, if enabled port(s) has no health check URL provided.

tbak:
Fixed by PR #412

Issue #389, opened 2015-02-05
Eureka 1.x - there is no mechanism provided to remove overridden status


tbak:
Eureka 1 allows one to override an instance status via issuing the following REST call:
```PUT /eureka/v2/apps/appID/instanceID/status?value=OUT_OF_SERVICE```
This is useful if we want to take otherwise healthy instance out of service. The instance info data contain afterwards two status related fields:
```
<status>OUT_OF_SERVICE</status>
<overriddenstatus>OUT_OF_SERVICE</overriddenstatus>
```
To mark instance as up we can modify again the overriddenstatus field by calling:
```PUT /eureka/v2/apps/appID/instanceID/status?value=UP```
If however the instance real state change(s), that is hidden by the enforced value.

We can address this issue in two ways:
* support DELETE HTTP method to remove the overridden status
* merge the two states (see table below)

Status | Overridden status | Merge
---------|-------------------------|---------
DOWN | DOWN | DOWN
DOWN | * | DOWN
UP | OUT_OF_SERVICE | OUT_OF_SERVICE
UP | UP | UP

NiteshKant:
I would prefer the support for DELETE overridden status for the following reason:

The merge algorithm addresses the "most common" usecase for today (for Netflix). I can think of corner cases when someone wants to override DOWN status from the instance with UP, may be in the case the healthcheck has an issue or it is flapping. 

There is value in being consistent with the definition of "overridden status" that it __always__ overrides the status with no special cases. If the override does not apply anymore, then it should be deleted.

tbak:
This issue is a duplicate of #89.

tbak:
Fixed by #412.
Overridden status can be removed via DELETE REST call.

PR #412, opened 2015-02-09
Fixes for #389, #402

tbak:


qiangdavidliu:
The code looks good, but a more consistent behaviour could be to fix the update code so that the (non-) overridden status is not updated when an override is specified, and hence at deletion time we'll only need to remove the overridden status. Then on the InstanceInfo getter side, we can return either the base status or the override status based on whether the override is present.

tbak:
As currently setting status override modifies the status fields itself, this change would require bigger code refactoring. Instead, the status override DELETE will accept optional query parameter to set the desired status. If not set it will default to UNKNOWN.

PR #413, opened 2015-02-10
2.x client fixes

qiangdavidliu:
- client interestChannel now evict older entries once new channel is available
- AbstractClientChannel properly handle connection lifecycle

### DIFF
--- a/eureka2-core/src/main/java/com/netflix/eureka2/registry/SourcedEurekaRegistry.java
+++ b/eureka2-core/src/main/java/com/netflix/eureka2/registry/SourcedEurekaRegistry.java
@@ -48,16 +48,10 @@ public interface SourcedEurekaRegistry<T> {
     Observable<ChangeNotification<T>> forInterest(Interest<T> interest, Source.Matcher sourceMatcher);
 
     /**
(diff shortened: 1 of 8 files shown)


eureka2-core/src/main/java/com/netflix/eureka2/registry/SourcedEurekaRegistry.java
```diff
@@ -48,16 +48,10 @@
...
      * @return an observable of long denoting the number of holder items touched for the eviction
      */
-    Observable<Long> evictAll(Source source);
-
-    /**
-     * Evict all registry info for all sources
-     * @return an observable of long denoting the number of holder items touched for the eviction
-     */
-    Observable<Long> evictAll();
+    Observable<Long> evictAllExcept(Source source);
```
tbak: How do we evict now replication channels? We need to put on the eviction queue a specific source.

qiangdavidliu: Good catch. Looks like the replication eviction was lost as well during the refactoring just like for the interestChannel. Will fix.

PR #418, opened 2015-02-11
Objects of different types compared with equals method.

C0rWin:
Compare data center name object with its name rather than instance of data center object itself.


qiangdavidliu:
Hi @C0rWin , thanks for the pull and the fix. We took another look at the logic around the fix, and it looks like the fixes will actually enable some code paths that have been dormant all this time (and eureka has been working in production for a long time). A better fix would be to remove both checks and any associated code paths that were never exercised instead, i.e. remove both conditional statements and their inner executions. Feel free to resubmit your pull with the change if you'd like, otherwise we can make the fix also. Thanks.

C0rWin:
Ok, I'll take another look and will remove code as you suggested.

qiangdavidliu:
Thanks for the contrib @C0rWin

PR #420, opened 2015-02-12
2.x BaseMessageConnection fix

qiangdavidliu:


eureka2-core/src/main/java/com/netflix/eureka2/transport/base/BaseMessageConnection.java
```diff
@@ -180,10 +181,11 @@ public void call(Object o) {
         }).doOnTerminate(new Action0() {
             @Override
             public void call() {
-                shutdown();
+                // always close with an exception here as the underlying connection never onError
+                shutdown(CONNECTION_INPUT_ONCOMPLETE_EXCEPTION);
```
tbak: Isn't it strange to always terminate a stream with an error? What was the original reason for this change?

PR #423, opened 2015-02-13
2.x multiple updates related to new interest stream implementation with buffer hints

tbak:


eureka2-client/src/main/java/com/netflix/eureka2/client/resolver/EurekaServerResolver.java
```diff
@@ -119,20 +131,34 @@ public EurekaServerResolver(
...
-                        Server newServer = new Server(socketAddress.getHostString(), socketAddress.getPort());
-                        servers.add(newServer);
-                        return servers;
-                    }
-                })
-                .retryWhen(new RetryStrategyFunc(500, 2, true))
-                .concatWith(Observable.just(Collections.<Server>emptySet()))
-                .take(1);
+        final EurekaClient eurekaClient = eurekaClientBuilder.build();
+        try {
```
qiangdavidliu: Why do we need to try catch here? It's all rx so the doOnTerminate (we can be even safer with a finallyDo) should handle the client closure.

eureka2-client/src/main/java/com/netflix/eureka2/client/resolver/ServerResolvers.java
```diff
@@ -43,16 +47,18 @@ public static ServerResolver forDnsName(String dnsName, int port) {
...
     public static ServerResolver fromWriteServer(ServerResolver writeServerResolver,
                                                  String readClusterVip,
-                                                 EurekaClientMetricFactory metricFactory) {
+                                                 EurekaClientMetricFactory clientMetricFactory,
+                                                 EurekaRegistryMetricFactory registryMetricFactory) {
         return new EurekaServerResolverBuilder()
                 .withBootstrapResolver(writeServerResolver)
                 .withReadServerInterest(Interests.forVips(readClusterVip))
-                .withMetricFactory(metricFactory)
+                .withMetricFactory(clientMetricFactory)
```
qiangdavidliu: the withMetricFactory overload seems confusing, we can be more specific with the type of metric factories?

eureka2-client/src/main/java/com/netflix/eureka2/client/resolver/EurekaServerResolver.java
```diff
@@ -219,14 +250,20 @@ public EurekaServerResolver build() {
...
             }
 
-            TcpDiscoveryClient transportClient =
-                    (TcpDiscoveryClient) TransportClients.newTcpDiscoveryClient(transportConfig, bootstrapResolver,metricFactory);
-            SnapshotInterestChannel snapshotInterestChannel = new SnapshotInterestChannelImpl(transportClient);
-            return new EurekaServerResolver(snapshotInterestChannel, readServerInterest, serviceSelector, loadBalancerBuilder);
+            EurekaClientBuilder eurekaClientBuilder = EurekaClientBuilder.newBuilder()
+                    .withMetricFactory(clientMetricFactory)
+                    .withMetricFactory(registryMetricFactory)
+                    .withWriteServerResolver(ServerResolvers.just("localhost", 0)) // TODO We have provide some resolver, otherwise validation fails
```
qiangdavidliu: At part of the client reorganisation, the final EurekaClient should be provided as a wrapper of both a RegistrationClient and an InterestClient. When that is available, we can just use a raw InterestClient here.

eureka2-client/src/main/java/com/netflix/eureka2/client/functions/ChangeNotificationFunctions.java
```diff
@@ -0,0 +1,88 @@
...
+    private ChangeNotificationFunctions() {
+    }
+
+    /**
+     * Convert change notification stream with buffering sentinels into stream of lists, where each
+     * list element contains a batch of data delineated by the markers.
+     *
+     * @return observable of non-empty list objects
+     */
+    public static <T> Transformer<ChangeNotification<T>, List<ChangeNotification<T>>> buffers() {
```
qiangdavidliu: What is the contract for this operations if we received two successive BufferSentinels from the stream? Should we emit empty lists or emit nothing (what the code seem to do right now)? We should document the expected behaviour in this case.

eureka2-client/src/test/java/com/netflix/eureka2/client/functions/ChangeNotificationFunctionsTest.java
```diff
@@ -0,0 +1,67 @@
...
+    public void testSnapshotFunctionGeneratesDistinctValueSets() throws Exception {
+        PublishSubject<List<ChangeNotification<String>>> notificationSubject = PublishSubject.create();
+
+        ExtTestSubscriber<Set<String>> testSubscriber = new ExtTestSubscriber<>();
+        notificationSubject.compose(ChangeNotificationFunctions.<String>snapshots()).subscribe(testSubscriber);
+
+        notificationSubject.onNext(FIRST_BATCH);
+        assertThat(testSubscriber.takeNext(), is(equalTo((Set) asSet("A", "B"))));
+        notificationSubject.onNext(SECOND_BATCH);
+        assertThat(testSubscriber.takeNext(), is(equalTo((Set) asSet("B", "C"))));
```
qiangdavidliu: Can we add an extra test here to validate that a third batch that does not change the snapshot == no emits?

eureka2-client/src/main/java/com/netflix/eureka2/client/resolver/EurekaServerResolver.java
tbak: Try/catch is a safeguard here if the observable creation fails for whatever reason. In such case the doOnTerminate will never be executed, as this observable do not exist yet nor is subscribed to.
It is to guarantee that we release resources (shutdown eureka client).

eureka2-client/src/main/java/com/netflix/eureka2/client/resolver/ServerResolvers.java
tbak: I agree. We will have to align this in other parts of code.

eureka2-client/src/main/java/com/netflix/eureka2/client/resolver/EurekaServerResolver.java
tbak: Absolutely. We should refactor it once new API is available.

eureka2-client/src/main/java/com/netflix/eureka2/client/functions/ChangeNotificationFunctions.java
tbak: We do not emit empty lists, which is what we would have to do if we have consecutive buffer sentinels issued. This is documented in @return description, but I will make it more explicit.

eureka2-client/src/test/java/com/netflix/eureka2/client/functions/ChangeNotificationFunctionsTest.java
tbak: Done

PR #424, opened 2015-02-14
2.x server address fixes

qiangdavidliu:
- Moving Server to core/utils for more generic use.
- Removing usage of InetSocketAddress.
- Replication peer resolve to dns name only (except for A records)


eureka2-client/src/main/java/com/netflix/eureka2/client/Eureka.java
```diff
@@ -17,9 +17,9 @@
 package com.netflix.eureka2.client;
 
 import com.netflix.eureka2.client.resolver.ServerResolver;
-import com.netflix.eureka2.client.resolver.ServerResolver.Server;
 import com.netflix.eureka2.client.resolver.ServerResolvers;
 import com.netflix.eureka2.transport.EurekaTransports;
+import com.netflix.eureka2.utils.Server;
```
tbak: As ```Server``` is part of public API, it should not reside in utils package. Can we move it to some other place (com.netflix.eureka2?)

qiangdavidliu: Good point, moved to com.netflix.eureka2

PR #434, opened 2015-02-24
1.x Cache string values to reduce memory footprint.

tbak:


eureka-client/src/main/java/com/netflix/discovery/converters/StringCache.java
```diff
@@ -0,0 +1,38 @@
+package com.netflix.discovery.converters;
+
+import java.util.Map;
+import java.util.WeakHashMap;
+
+/**
+* @author Tomasz Bak
+*/
+public class StringCache {
```
NiteshKant: Any reason why you are not using String.intern() simply?

eureka-client/src/main/java/com/netflix/discovery/converters/StringCache.java
```diff
@@ -0,0 +1,38 @@
...
+
+    public static final int LENGTH_LIMIT = 38;
```
NiteshKant: What is the reasoning for this length limitation?

eureka-client/src/main/java/com/netflix/discovery/converters/StringCache.java
tbak: Using String.intern() was technically forbidden in the past for string pool usage. This implementation of StringCache follows xstream's StringConverter which I tried to use, but since we have custom serializer it is never invoked.
I checked again, and actually in Java7 the String.intern() implementation is fixed, and it is better alternative than having own string pools, but unless running on 7u40, it requires setting -XX:StringTableSize=N to proper value (before 7u40 default = 1009, later default = 60013 which is better but we might need even higher value). Check this article: http://java-performance.info/string-intern-in-java-6-7-8/
Given that we do not control which Java7 JVM versions clients run, and we cannot enforce proper JVM configuration, sticking to own String pool implementation seems to be better option short term. Medium/long term, if we want to make similar optimization in Eureka 2.x I would go for String.intern.

eureka-client/src/main/java/com/netflix/discovery/converters/StringCache.java
tbak: StringCache implementation follows xstream's StringConverter which has this optimization. Processing long strings is expensive, and probability of having duplicates decreases with increased string length.
This PR was a quick prototype to check how big is the benefit of pooling strings. We should revise it further if we agree we want this change.

eureka-client/src/main/java/com/netflix/discovery/converters/Converters.java
```diff
@@ -366,13 +377,13 @@ public Object unmarshal(HierarchicalStreamReader reader,
...
 
                 if (ELEM_HOST.equals(nodeName)) {
-                    builder.setHostName(reader.getValue());
+                    builder.setHostName(cache.cachedValueOf(reader.getValue()));
                 } else if (ELEM_APP.equals(nodeName)) {
-                    builder.setAppName(reader.getValue());
+                    builder.setAppName(cache.cachedValueOf(reader.getValue()));
                 } else if (ELEM_IP.equals(nodeName)) {
-                    builder.setIPAddr(reader.getValue());
+                    builder.setIPAddr(cache.cachedValueOf(reader.getValue()));
```
qiangdavidliu: Are these the best fields for caching? IP for example would be pretty unique for all instances.

qiangdavidliu:
Rather than implementing a StringCache independent of xstream, can't we just register a StringConverter with high priority when we register the specific converters for the JsonXStream and XmlXStream?

We may also want to look at implementing an alternative to the default StringConverter similar to http://javadoc.jenkins-ci.org/hudson/util/HeapSpaceStringConverter.html.

tbak:
I tried to setup StringConverter, and my custom converter implementation, but it never got triggered. Our custom serializers do not delegate processing further once they got basic types. If you know how to enforce delegation to StringConverter it would be much nicer implementation.

eureka-client/src/main/java/com/netflix/discovery/converters/Converters.java
tbak: Good catch. I will remove it.

PR #435, opened 2015-02-24
2.x Implement custom EurekaInterestClient for Eureka Read server.

tbak:
The custom client uses single subscription (fullRegistryFetch), and propagates properly channel level buffer markers to interest subscribers.

### DIFF
--- a/eureka2-client/src/main/java/com/netflix/eureka2/client/interest/EurekaInterestClientImpl.java
+++ b/eureka2-client/src/main/java/com/netflix/eureka2/client/interest/EurekaInterestClientImpl.java
@@ -1,7 +1,6 @@
 package com.netflix.eureka2.client.interest;
 
 import javax.inject.Inject;
-import java.util.concurrent.atomic.AtomicBoolean;
 
 import com.netflix.eureka2.channel.ChannelFactory;
 import com.netflix.eureka2.channel.InterestChannel;
@@ -9,17 +8,11 @@ import com.netflix.eureka2.connection.RetryableConnection;
 import com.netflix.eureka2.connection.RetryableConnectionFactory;
 import com.netflix.eureka2.interests.ChangeNotification;
 import com.netflix.eureka2.interests.Interest;
-import com.netflix.eureka2.registry.Source;
-import com.netflix.eureka2.registry.Sourced;
 import com.netflix.eureka2.registry.SourcedEurekaRegistry;
 import com.netflix.eureka2.registry.instance.InstanceInfo;
-import com.netflix.eureka2.utils.rx.RetryStrategyFunc;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 import rx.Observable;
 import rx.Subscriber;
 import rx.functions.Action0;
-import rx.functions.Func1;
 import rx.functions.Func2;
 
 /**
@@ -33,15 +26,9 @@ import rx.functions.Func2;
  *
  * @author David Liu
  */
-public class EurekaInterestClientImpl implements EurekaInterestClient {
-    private static final Logger logger = LoggerFactory.getLogger(EurekaInterestClientImpl.class);
+public class EurekaInterestClientImpl extends AbstractInterestClient {
 
-    private static final int DEFAULT_RETRY_WAIT_MILLIS = 500;
-
-    private final AtomicBoolean isShutdown;
-    private final SourcedEurekaRegistry<InstanceInfo> registry;
     private final InterestTracker interestTracker;
-
     private final RetryableConnection<InterestChannel> retryableConnection;
 
     @Inject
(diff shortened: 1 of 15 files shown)


eureka2-client/src/main/java/com/netflix/eureka2/client/interest/EurekaInterestClientImpl.java
```diff
@@ -71,54 +57,8 @@ public EurekaInterestClientImpl(SourcedEurekaRegistry<InstanceInfo> registry,
 
         this.retryableConnection = retryableConnectionFactory.singleOpConnection(opStream, executeOnChannel);
 
-        // subscribe to the base interest channels to do cleanup on every channel refresh.
-        retryableConnection.getChannelObservable()
```
qiangdavidliu: The original intent for having these in the constructor was to ensure that they are not multi-subscribed to, and also that they are always eagerly subscribed to. Moving these to methods in the abstract client allows breakage of the above assumptions for future implementors. From what I can see the two implementing clients always call
```java
registryEvictionSubscribe(retryableConnection);
lifecycleSubscribe(retryableConnection);
````
anyway in the constructor, so can we just move this block to the constructor for the new Abstract class instead of making them methods?

File: eureka2-read-server/src/main/java/com/netflix/eureka2/server/registry/EurekaReadServerRegistry.java [hunk 2]
```diff
@@ -125,6 +119,40 @@ public int size() {
...
+                if (buffer.isEmpty()) {
+                    return Observable.empty();
+                }
+                Observable<ChangeNotification<InstanceInfo>> result;
+                if (buffer.size() == 1) {
+                    result = Observable.just(buffer.get(0));
+                } else {
+                    List<ChangeNotification<InstanceInfo>> batch = new ArrayList<>(2 + buffer.size());
+                    batch.add(bufferStartNotification);
+                    for (ChangeNotification<InstanceInfo> item : buffer) {
```
qiangdavidliu: can just do an addAll(buffer) here

eureka2-read-server/src/main/java/com/netflix/eureka2/server/registry/EurekaReadServerRegistry.java
```diff
@@ -83,19 +87,9 @@ public int size() {
...
-        return eurekaClient.forInterest(interest).map(new Func1<ChangeNotification<InstanceInfo>, ChangeNotification<InstanceInfo>>() {
-            @Override
-            public ChangeNotification<InstanceInfo> call(ChangeNotification<InstanceInfo> notification) {
-                if (notification.isDataNotification()) {
-                    return notification;
-                }
-                return null;
-            }
-        }).filter(filterNullValuesFunc());
+        return interestClient.forInterest(interest).flatMap(bufferStartEndDelineateFun(interest));
```
qiangdavidliu: So the change in behaviour here is that the read registry will always buffer inbetween each BufferSentinel?

eureka2-read-server/src/main/java/com/netflix/eureka2/server/interest/FullFetchInterestClient.java
```diff
@@ -0,0 +1,70 @@
...
+import rx.functions.Func1;
+
+/**
+ * {@link EurekaInterestClient} implementation with single full registry fetch subscription.
+ * Client interest subscriptions are not propagated to the channel, as the registry is already eagerly
+ * subscribed to the full content.
+ *
+ * @author Tomasz Bak
+ */
+public class FullFetchInterestClient extends AbstractInterestClient {
```
qiangdavidliu: Seems like the standard InterestClient should be able to be used for the same purpose as this special client, as long as we do an eager forInterest(FullRegistryInterest) from the ReadRegistry?

eureka2-client/src/main/java/com/netflix/eureka2/client/interest/EurekaInterestClientImpl.java
tbak: Unfortunately we cannot do that, as ```retryableConnection``` is not constructed yet. But as these methods are protected, the risk of misuse is not big.

eureka2-read-server/src/main/java/com/netflix/eureka2/server/registry/EurekaReadServerRegistry.java
tbak: Sure, I will fix it.

eureka2-read-server/src/main/java/com/netflix/eureka2/server/registry/EurekaReadServerRegistry.java
tbak: I have added a comment. It would be actually good to have some flexibility in choosing which markers we prefer on registry/client API as technically we have both implementations. Possibly improvement in the future.

eureka2-read-server/src/main/java/com/netflix/eureka2/server/interest/FullFetchInterestClient.java
tbak: This client does more. It does not allow for any other interest subscriptions.

PR #436, opened 2015-02-25
2.x Changes to MultiSourcedDataHolder implementation

qiangdavidliu:
- updates for the same origin:name Sources replace each other regardless of source.id
- deletes only execute when origin:name:id equals for the sources

### DIFF
--- a/eureka2-client/src/main/java/com/netflix/eureka2/client/interest/EurekaInterestClientImpl.java
+++ b/eureka2-client/src/main/java/com/netflix/eureka2/client/interest/EurekaInterestClientImpl.java
@@ -78,7 +78,7 @@ public class EurekaInterestClientImpl implements EurekaInterestClient {
                     public Observable<Long> call(InterestChannel interestChannel) {
                         if (interestChannel instanceof Sourced) {
                             Source toRetain = ((Sourced) interestChannel).getSource();
-                            return registry.evictAllExcept(toRetain);
+                            return registry.evictAllExcept(Source.matcherFor(toRetain));
                         }
                         return Observable.empty();
                     }
--- a/eureka2-core/src/main/java/com/netflix/eureka2/registry/NotifyingInstanceInfoHolder.java
+++ b/eureka2-core/src/main/java/com/netflix/eureka2/registry/NotifyingInstanceInfoHolder.java
@@ -1,6 +1,7 @@
 package com.netflix.eureka2.registry;
 
 import java.util.Collection;
+import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Set;
@@ -37,7 +38,7 @@ public class NotifyingInstanceInfoHolder implements MultiSourcedDataHolder<Insta
     private final NotificationsSubject<InstanceInfo> notificationSubject;  // subject for all changes in the registry
 
     private final HolderStoreAccessor<NotifyingInstanceInfoHolder> holderStoreAccessor;
-    private final LinkedHashMap<Source, InstanceInfo> dataMap;  // for order
+    private final DataStore dataStore;
     private final NotificationTaskInvoker invoker;
     private final String id;
     private Snapshot<InstanceInfo> snapshot;
@@ -52,7 +53,7 @@ public class NotifyingInstanceInfoHolder implements MultiSourcedDataHolder<Insta
         this.notificationSubject = notificationSubject;
         this.invoker = invoker;
         this.id = id;
-        this.dataMap = new LinkedHashMap<>();
+        this.dataStore = new DataStore();
     }
 
     @Override
@@ -62,7 +63,7 @@ public class NotifyingInstanceInfoHolder implements MultiSourcedDataHolder<Insta
 
     @Override
     public int size() {
-        return dataMap.size();
+        return dataStore.size();
     }
 
     @Override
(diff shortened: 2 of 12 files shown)


eureka2-client/src/main/java/com/netflix/eureka2/client/interest/EurekaInterestClientImpl.java
```diff
@@ -78,7 +78,7 @@ public EurekaInterestClientImpl(SourcedEurekaRegistry<InstanceInfo> registry,
                     public Observable<Long> call(InterestChannel interestChannel) {
                         if (interestChannel instanceof Sourced) {
                             Source toRetain = ((Sourced) interestChannel).getSource();
-                            return registry.evictAllExcept(toRetain);
+                            return registry.evictAllExcept(Source.matcherFor(toRetain));
```
tbak: Can we rename Matcher to SourceMatcher? This name is used by many libraries.

eureka2-core/src/main/java/com/netflix/eureka2/registry/NotifyingInstanceInfoHolder.java
```diff
@@ -227,19 +230,35 @@ public void call(Throwable throwable) {
...
                 ", id='" + id + '\'' +
                 ", snapshot=" + snapshot +
                 "} " + super.toString();
     }
 
+    /**
+     * @return true if the two sources have the same origin and name. Don't use a matcher from source
+     *         to avoid the object creation.
+     */
+    private boolean matches(Source one, Source two) {
```
tbak: Can we move this method to Source class? Might be useful in other places.

eureka2-core/src/main/java/com/netflix/eureka2/registry/NotifyingInstanceInfoHolder.java
```diff
@@ -255,4 +274,94 @@ public void shutdown() {
...
+         * - If a matching source already exist in the sourceMap, remove from the dataStore first before add (as the
+         *   curr source may not equal due to a different source id).
+         * - Otherwise, just add to the dataStore
+         * - finally, add the source and sourceKey to the sourceMap
+         */
+        public void put(Source source, InstanceInfo instanceInfo) {
+            String sourceKey = sourceKey(source);
+            Source currIfExist = sourceMap.get(sourceKey);
+            if (currIfExist != null) {
+                dataMap.remove(currIfExist);
```
tbak: An update on existing entry will push it at the back of the queue, which I think is not what we want, especially for head.
If we do not remove this entry, the order will be preserved.

eureka2-core/src/main/java/com/netflix/eureka2/registry/NotifyingInstanceInfoHolder.java
```diff
@@ -255,4 +274,94 @@ public void shutdown() {
...
+            }
+            return null;
+        }
+
+        public InstanceInfo getExact(Source source) {
+            return dataMap.get(source);
+        }
+
+        /**
+         * Matches sources on origin:name only.
```
tbak: Origin, name, id?

eureka2-core/src/main/java/com/netflix/eureka2/registry/NotifyingInstanceInfoHolder.java
```diff
@@ -255,4 +274,94 @@ public void shutdown() {
             super.shutdown();
         }
     }
+
+    /**
+     * Assume access to this is synchronized.
+     * All access to the dataStore first route through the sourceMap for the authoritative source to use
+     */
+    /* visible for testing */ static class DataStore {
+        protected final Map<String, Source> sourceMap = new HashMap<>();
```
tbak: Can we use custom comparator (by origin/name) and eliminate this data structure?

eureka2-client/src/main/java/com/netflix/eureka2/client/interest/EurekaInterestClientImpl.java
qiangdavidliu: sure

eureka2-core/src/main/java/com/netflix/eureka2/registry/NotifyingInstanceInfoHolder.java
qiangdavidliu: This is a somewhat misleading match specific to the holder, which is why it is here as a private method. The Source class already have matcher methods that does similar with a more sensible API.

eureka2-core/src/main/java/com/netflix/eureka2/registry/NotifyingInstanceInfoHolder.java
qiangdavidliu: Unfortunately this is not trivial unless we override Source's equals and hashcode to ignore the id, which come with its own set of issues. The whole ordering of the copies is just an optimisation anyway, so I see no strong need for it to be enforced.

eureka2-core/src/main/java/com/netflix/eureka2/registry/NotifyingInstanceInfoHolder.java
qiangdavidliu: updated comment

eureka2-core/src/main/java/com/netflix/eureka2/registry/NotifyingInstanceInfoHolder.java
qiangdavidliu: We can use a treeMap, but we'd still have to override Source's equals and hashcode to ignore id.

Issue #439, opened 2015-02-26
2.x Eureka server health monitoring


tbak:
Eureka 2.0 health monitoring

# Concept

Each Eureka component manages its own health information, which is aggregated in one place where overall system health status is derived.

A component that wishes to provide its health status must implement:
```java
public interface HealthStatusProvider<SUBSYSTEM> {

   Observable<HealthStatusUpdate<SUBSYSTEM>> healthStatus();
}
```
```healthStatus``` method returns an observable of health status changes. The returned observable shall never emit an error, and should complete when the subsystem is shut down. Status value represents internal state of the component.  The health status information is captured in ```HealthStatusUpdate``` object:

```java
public class HealthStatusUpdate<SUBSYSTEM> {

    public HealthStatusUpdate(Status status,
                              SubsystemDescriptor<SUBSYSTEM> description) {
      ...
    }

    public STATUS getStatus() { }

    public SubsystemDescriptor<SUBSYSTEM> getDescriptor() { }
}
```

```SubsystemDescriptor``` JavaBean provides static, supplementary information primary for UIs.

The aggregation logic is captured in ```HealthStatusAggregator``` class, which itself extends ```HealthStatusProvider```. Apart for returning the aggregated state, it provides access to its underlying components:
```java
public interface HealthStatusAggregator<SUBSYSTEM extends HealthStatusAggregator<SUBSYSTEM>>
        extends HealthStatusProvider<SUBSYSTEM> {
    List<HealthStatusProvider<?>> components();
}
```

# Eureka server self registration

Eureka self registration service subscribes to ```HealthStatusAggregator```, and updates its own ```InstanceInfo.status``` accordingly, which gets immediately propagated to Eureka registry.

# External Health check API 

## REST API

Current Eureka health status can be accessed via REST GET method call:
```
curl -X GET http://eureka.host/eureka/health

HTTP/1.1 200 OK
HEALTHY
```

## WebSockets API

Clients wishing to receive live updates about Eureka health status, can connect to WebSocket endpoint:
```
open ws://eureka.host/eureka/ws/health

text-message: {
    "descriptor":{
        "title":"Client interest channel",
        "description":"...",
        "type":"InterestChannel"
    },
    "status":"DOWN"
}
text-message: {
	"descriptor":{
		"title":"Aggregated health check status",
		"description":"...",
		"type":"EurekaHealthStatusAggregator"
	},
	"status":"DOWN",
}
...
text-message: {
	"descriptor": {
[...]

PR #442, opened 2015-02-27
2.x Eureka health status monitoring

tbak:
See issue #439


eureka2-core/src/main/java/com/netflix/eureka2/health/AbstractHealthStatusProvider.java
```diff
@@ -0,0 +1,39 @@
...
+    @Override
+    public Observable<HealthStatusUpdate<SUBSYSTEM>> healthStatus() {
+        return healthSubject;
+    }
+
+    public boolean moveHealthTo(Status newStatus) {
+        if (currentStatus.getAndSet(newStatus) == newStatus) {
+            return false;
+        }
+        healthSubject.onNext(new HealthStatusUpdate<SUBSYSTEM>(newStatus, descriptor));
```
qiangdavidliu: Is concurrency control important here? There is still no concurrency guard against two threads that execute in the following order:
Thread1               Thread2
S1 -> S2              
                            S2 -> S3
                            onNext S3
onNext S2

tbak: This method should not be called concurrently. If there might be concurrent
updates, it is up to the component itself to synchronize access.

eureka2-core/src/main/java/com/netflix/eureka2/registry/PreservableEurekaRegistry.java
```diff
@@ -247,7 +271,7 @@ public void call() {
             } else {
                 selfPreservation.set(true);
                 metrics.setSelfPreservation(true);
-                logger.info("Entering self preservation mode");
+                logger.info("Entering self preserv4ation mode");
```
qiangdavidliu: typo

eureka2-core/src/main/java/com/netflix/eureka2/registry/PreservableEurekaRegistry.java
```diff
@@ -194,17 +209,26 @@ public void call() {
...
+    }
+
     /**
      * FIXME eviction strategies need a rethink
      * >= 0 as when both sizes are equal, we still allow eviction to happen as they may be stale copies
      */
     private boolean allowedToEvict() {
-        return evictionStrategy.allowedToEvict(expectedRegistrySize, eurekaRegistry.size()) >= 0;
+        boolean allowed = evictionStrategy.allowedToEvict(expectedRegistrySize, eurekaRegistry.size()) >= 0;
+        moveHealthTo(allowed ? Status.UP : Status.DOWN);
```
qiangdavidliu: Per our discussion, we should not mark servers as UP even when they are in selfPreservation. Otherwise we may redirect all traffic to a (possibly) much smaller set of UP servers and take them down.

eureka2-server/src/main/java/com/netflix/eureka2/server/health/EurekaHealthStatusAggregator.java
```diff
@@ -0,0 +1,115 @@
...
+                                    if (STATUS_PRIORITY_MAP.get(next) < STATUS_PRIORITY_MAP.get(aggregate)) {
+                                        aggregate = next;
+                                    }
+                                }
+                                return aggregate;
+                            }
+                        });
+                    }
+                })
+                .distinct()
```
qiangdavidliu: Do you want distinct or distinctUntilChanged?

eureka2-server/src/main/java/com/netflix/eureka2/server/http/HealthConnectionHandler.java
```diff
@@ -0,0 +1,72 @@
...
+                        updatesObservables.add(aggregatedObservable);
+                        return Observable.merge(updatesObservables);
+                    }
+                });
+    }
+
+    @Override
+    public Observable<Void> handle(final ObservableConnection<WebSocketFrame, WebSocketFrame> connection) {
+        // Ignore messages from client
+        connection.getInput().subscribe();
```
qiangdavidliu: Would we subscribe each time we call handle here?

eureka2-server/src/main/java/com/netflix/eureka2/server/health/EurekaHealthStatusAggregator.java
tbak: My mistake. Should be distinctUntilChanged

eureka2-server/src/main/java/com/netflix/eureka2/server/http/HealthConnectionHandler.java
tbak: Yes, it is RxNetty ObservableConnection handling code.

PR #454, opened 2015-03-05
2.x Eureka 1.x backwards compatible API (query operations)

tbak:


eureka2-core/src/main/java/com/netflix/eureka2/interests/ApplicationInterest.java
```diff
@@ -2,82 +2,23 @@
...
-        return true;
-    }
-
-    @Override
-    public int hashCode() {
-        int result = applicationName != null ? applicationName.hashCode() : 0;
-        result = 31 * result + (operator != null ? operator.hashCode() : 0);
-        return result;
+    public ApplicationInterest(String secureVip, Operator operator) {
+        super(secureVip, operator);
```
qiangdavidliu: why secureVip?

eureka2-core/src/main/java/com/netflix/eureka2/interests/ChangeNotifications.java
```diff
@@ -67,4 +75,83 @@ private ChangeNotifications() {
...
+
+    /**
+     * Convert change notification stream with buffering start/end markers into stream of lists, where each
+     * list element contains a batch of data delineated by the markers. Only non-empty lists are
+     * issued, which means that for two successive BufferSentinels from the stream, the second
+     * one will be swallowed.
+     *
+     * @return observable of non-empty list objects
+     */
+    public static <T> Transformer<ChangeNotification<T>, List<ChangeNotification<T>>> delineatedBuffers() {
```
qiangdavidliu: Are these functions just moved over from ChangeNotificationFunctions.java or are they changed as well?

eureka2-core/src/main/java/com/netflix/eureka2/interests/InstanceInterest.java
```diff
@@ -2,82 +2,23 @@
...
-        return true;
-    }
-
-    @Override
-    public int hashCode() {
-        int result = instanceId != null ? instanceId.hashCode() : 0;
-        result = 31 * result + (operator != null ? operator.hashCode() : 0);
-        return result;
+    public InstanceInterest(String secureVip, Operator operator) {
+        super(secureVip, operator);
```
qiangdavidliu: Again secureVip? Do you mean instanceId?

eureka2-ext/eureka2-eureka1x-rest-api/build.gradle
```diff
@@ -0,0 +1,32 @@
+/*
```
qiangdavidliu: Can we standardise our naming here? we refer to 2.0 as eureka2 (no "x"), but then refer to eureka1 as eureka1x. May be more clear to change eureka1x to eureka1.

eureka2-server/src/main/java/com/netflix/eureka2/server/spi/ExtAbstractModule.java
```diff
@@ -27,14 +27,15 @@
...
      * Module will be added to the container if it is runnable. It is up to the
      * module implementation to decide what are the conditions for that.
-     * As a minimum modules shall verify that its configuration is complete.
-     * This allows to failover to default service implementation, if available.
-     * @param extensionContext
+     *
+     * @param serverType running server type
      */
-    public boolean isRunnable(ExtensionContext extensionContext) {
+    public boolean isRunnable(ServerType serverType) {
```
qiangdavidliu: More a question on implementation, do they all a "default" type handle so that if we do add new serverTypes in the future we won't need to change code for all current server types?

eureka2-core/src/main/java/com/netflix/eureka2/interests/ApplicationInterest.java
tbak: copy-paste error
fixed

eureka2-core/src/main/java/com/netflix/eureka2/interests/ChangeNotifications.java
tbak: delineatedBuffers is different, as it is using buffer start/end marker

eureka2-core/src/main/java/com/netflix/eureka2/interests/InstanceInterest.java
tbak: copy-paste error again

eureka2-server/src/main/java/com/netflix/eureka2/server/spi/ExtAbstractModule.java
tbak: We need to revisit the extensibility API, as it is very rigid now. If we keep ServerType, and keep it as an enum, the module themselves will have to handle "default" or unrecognized values properly.

eureka2-ext/eureka2-eureka1x-rest-api/build.gradle
tbak: I agree. I will rename all "1x"es to just "1"

eureka2-ext/eureka2-eureka1x-rest-api/src/main/java/com/netflix/eureka2/eureka1x/rest/AbstractEureka1xRequestHandler.java
```diff
@@ -0,0 +1,128 @@
...
+import rx.functions.Func2;
+
+/**
+ * @author Tomasz Bak
+ */
+public abstract class AbstractEureka1xRequestHandler implements RequestHandler<ByteBuf, ByteBuf> {
+
+    private static final Logger logger = LoggerFactory.getLogger(AbstractEureka1xRequestHandler.class);
+
+    public static final String ROOT_PATH = "/eureka1x/v2";
```
qiangdavidliu: Can we make this ROOT_PATH configurable? E.g. for backwards compatibility it does not align with current eureka1 root path.

eureka2-ext/eureka2-eureka1x-rest-api/src/main/java/com/netflix/eureka2/eureka1x/rest/Eureka1xRedirectRequestHandler.java
```diff
@@ -0,0 +1,123 @@
...
+        return Observable.empty();
+    }
+
+    private Observable<Void> redirectTo(InstanceInfo readServerInfo,
+                                        HttpServerRequest<ByteBuf> request,
+                                        HttpServerResponse<ByteBuf> response) {
+        ServiceEndpoint serviceEndpoint = HTTP_SERVICE_SELECTOR.returnServiceEndpoint(readServerInfo);
+
+        StringBuilder redirectBuilder = new StringBuilder("http://")
+                .append(serviceEndpoint.getAddress().getIpAddress())
```
qiangdavidliu: Do we have the choice here to redirect to either public or private ips? (For running in AWS Classic).

eureka2-ext/eureka2-eureka1x-rest-api/src/main/java/com/netflix/eureka2/eureka1x/rest/Eureka1xRegistrationRequestHandler.java
```diff
@@ -0,0 +1,129 @@
...
+
+/**
+ * @author Tomasz Bak
+ */
+public class Eureka1xRegistrationRequestHandler extends AbstractEureka1xRequestHandler {
+
+    private final Eureka1xRegistryProxy registryProxy;
+
+    /* For testing */
+    public Eureka1xRegistrationRequestHandler(Eureka1xRegistryProxy registryProxy,
```
qiangdavidliu: remove public?

tbak: Fixed

eureka2-ext/eureka2-eureka1x-rest-api/src/main/java/com/netflix/eureka2/eureka1x/rest/Eureka1xRedirectRequestHandler.java
tbak: It is a bit tricky here, as we redirect clients which might be inside or outside.
Probably best default would be:
1. public hostname
2. if not available private ip
This would work best in AWS, but for generic cloud deployment might be too restrictive

Possibly we need some extensibility here, where the actual choice could be based on cloud type, and client IP address.

tbak: So here is the new code:
```java
        ServiceEndpoint serviceEndpoint = HTTP_PUBLIC_SERVICE_SELECTOR.returnServiceEndpoint(readServerInfo);
        String redirectHost = serviceEndpoint.getAddress().getHostName();
        if(redirectHost == null) {
            serviceEndpoint = HTTP_PRIVATE_SERVICE_SELECTOR.returnServiceEndpoint(readServerInfo);
            redirectHost = serviceEndpoint.getAddress().getIpAddress();
        }
```

eureka2-ext/eureka2-eureka1x-rest-api/src/main/java/com/netflix/eureka2/eureka1x/rest/query/AbstractEureka2RegistryView.java
```diff
@@ -0,0 +1,61 @@
...
+        this.notifications = notifications;
+        this.refreshIntervalMs = refreshIntervalMs;
+        this.scheduler = scheduler;
+    }
+
+    /**
+     * Connect happens immediately after child object is constructed.
+     */
+    protected void connect() {
+        subscription = notifications
```
qiangdavidliu: Does this mean that all the sub classes all have their individual subscriptions to the eureka2 registry? Would that introduce synchronisation differences between the different views? What about just having a single converter that converts from eureka2 registry interest to a eureka1 registry (store in app groups) at a regular interval, then feel all queries from there?

tbak: These all views are connected to the same local registry (which is reads server registry). We do not do any remote calls here.
We could still save some cycles by doing conversion from v1 to v2 once, but I intentionally kept it simple. We do 30 sec batching (by default), so the load from it will be minimal anyway.

eureka2-ext/eureka2-eureka1x-rest-api/src/main/java/com/netflix/eureka2/eureka1x/rest/registry/Eureka1xRegistryProxyImpl.java
```diff
@@ -0,0 +1,123 @@
...
+        return Result.NotFound;
+    }
+
+    @Override
+    public void unregister(RegistrationHandler cleanupHandler, long expiryTime) {
+        synchronized (handlers) {
+            String instanceId = cleanupHandler.getV1InstanceInfo().getId();
+            RegistrationHandler activeHandler = handlers.get(instanceId);
+            if (cleanupHandler == activeHandler && activeHandler.getExpiryTime() <= expiryTime) {
+                activeHandler.unregister();
```
qiangdavidliu: remove first from map then call unregister for better synchronisation

eureka2-ext/eureka2-eureka1x-rest-api/src/main/java/com/netflix/eureka2/eureka1x/rest/registry/Eureka1xRegistryProxyImpl.java
```diff
@@ -0,0 +1,123 @@
...
+        synchronized (handlers) {
+            RegistrationHandler handler = handlers.get(instanceId);
+            if (handler != null) {
+                InstanceInfo v1InstanceInfo = handler.getV1InstanceInfo();
+                if (appName.equalsIgnoreCase(v1InstanceInfo.getAppName())) {
+                    handler.renew();
+                    expiryQueue.enqueue(handler);
+                    return Result.Ok;
+                }
+                logger.warn("Application name in the request, does not match the one in the instance info ({} != {}, instanceId={}",
```
qiangdavidliu: for this case we may want to do an unregister (from old) and re-register (with new).

eureka2-ext/eureka2-eureka1x-rest-api/src/main/java/com/netflix/eureka2/eureka1x/rest/registry/Eureka1xRegistryProxyImpl.java
tbak: It really does not matter here as the handler is used internally here and
no client depends on it. Synchronization will force any pending new
registrations/unregistrations to wait.

eureka2-ext/eureka2-eureka1x-rest-api/src/main/java/com/netflix/eureka2/eureka1x/rest/registry/Eureka1xRegistryProxyImpl.java
tbak: InstanceInfo state does not change. Lease information is handled
exclusively in the holder class. Unregister would close Eureka 2
registration session, and we do not want it.

qiangdavidliu:
Rethinking this a little bit from a broader perspective, if we want to provide backwards compatibility with Eureka1, it might be easier to dependent on eureka 1.x server modules and reuse the same server side core components. The the work necessary will only be:
- add the necessary rxNetty REST resources to talk to the 1.x server components
- bridge interest from 2.x registry to 1.x registry
- bridge registrations to 1.x registry to the 2.x registry

tbak:
I thought first about the same, but:
1. the proxy code is very simple, and more efficient than maintaining
parallel registry
2. we would have to replicate 1.x registry to 2.x because of registering
1.x clients
3. we do not want to reuse replication mechanism from 1.x
4. cross region replication should be bridged directly to Eureka 2 cross
region replication mechanism (whatever it will be)

I think we would get ultimately in a lot of accidental complexity by
bringing in all Eureka 1 server.
But data mode, codecs are reused in this implementation.

PR #456, opened 2015-03-11
update dependency versions

brharrington:
The primary one that matters to me is ribbon. The older
version is still pulling in 'com.netflix.rxjava:rxjava-core'
which conflicts with classes in 'io.reactivex:rxjava'.

This change also changes the aws dependency to only pull
in the parts that are actually used by eureka client.


qiangdavidliu:
Thanks @brharrington

PR #457, opened 2015-03-12
2.x Adding basic randomisation to SelfClosingConnection

qiangdavidliu:


eureka2-core/src/main/java/com/netflix/eureka2/transport/base/SelfClosingConnection.java
```diff
@@ -44,7 +47,7 @@ public SelfClosingConnection(MessageConnection delegate, long lifecycleDurationM
 
         terminationWorker = terminationScheduler.createWorker();
         if (lifecycleDurationMs > 0) {
-            terminationWorker.schedule(selfTerminateTask, lifecycleDurationMs, TimeUnit.MILLISECONDS);
+            terminationWorker.schedule(selfTerminateTask, randomizeLifecycleDuration(lifecycleDurationMs), TimeUnit.MILLISECONDS);
```
tbak: Implementation looks good. What about unit test to verify it?

qiangdavidliu: It's just a mathematical formula so is provable that the result will be in the range of [lifecycleDuration/2, lifecycleDuration + lifecycleDuration/2]. As well, the random nature of the method make it hard to write an fixed unit test.

PR #466, opened 2015-03-17
2.x Rx-fy Index onSubscribe to be tolerant to backpressure

qiangdavidliu:


eureka2-core/src/main/java/com/netflix/eureka2/interests/Index.java
```diff
@@ -77,20 +78,20 @@ public void call(Subscriber<? super ChangeNotification<T>> subscriber) {
...
-                realTimeSubject.subscribe(subscriber);
 
-                // Reply initial state
-                for (ChangeNotification<T> notification : initStateHolder) {
-                    subscriber.onNext(notification);
-                }
-
-                // Now open real time source
-                realTimeSubject.resume();
+                realTimeSubject.mergeWith(Observable.from(initStateHolder).doOnCompleted(new Action0() {
```
NiteshKant: Is concating the two streams what we need here instead of merge?
I understand that this pauses the emission of the realtime stream but the use of merge is counter-intuitive. 
Also, is the notification subject Backpressure aware?

qiangdavidliu: we need to merge here instead of concat (which would be Ob.from(init).concatWith(realTimeSubject) ) as we need to ensure that the realTimeSubject is subscribed to before resume() is called on it. Otherwise, we would lose buffered messages that are stored while paused.

NiteshKant: I see. Since, `NotificationSubject` was written before backpressure in RxJava, it may be a good time now to have a second look at it.
I can see this being implemented as an operator that does not request upstream till the init state finish, which naturally buffers in the subject.

qiangdavidliu: Agree on that. Since this is a pretty central component the rewrite need to take a bit more care. As it stands right now, it works well enough for the RC1 release.

PR #469, opened 2015-03-18
2.x NotifyingInstanceInfoHolder eagerly promote updates from local source to snapshot if current snapshot is not from local source

qiangdavidliu:
This is done to fix the following race condition:
1. Write Servers A and B replicate to each other
2. Instance X registers with server A (A -> X local, B -> X replicated)
3. Instance X registration channel close and reconnect to server B
3a. A put X local into eviction queue
3b. B put X replicated into eviction queue
4. X register an unchanged instanceInfo to server B
4a. B's holder add X as a second copy as the original replicated copy is yet to be evicted
4b. Because the B holder's copies diff to empty delta, a CN with source local is never published
5. B's senderReplicationChannel disconnects and reconnects to A
5a. B's initState never sees the new X local as the CN is never published, resulting in X
not replicating to server A.


qiangdavidliu:
Actually rethinking about this a similar issue can also occur on the removal side as well, which will lead to stale instances. This will need a deeper think.

PR #470, opened 2015-03-19
1.x Use Jackson JSON serializer instead of XStream to improve serialization performance/memory usage.

tbak:
XML serialization is still done by XStream.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -11,6 +11,7 @@ ext {
     archaiusVersion='0.6.5'
     ribbonVersion='2.0.0'
     blitzVersion='1.34'
+    jacksonVersion='1.9.13'
 }
 
 subprojects {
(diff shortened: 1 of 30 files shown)


aspyker:
The only further enhancements I can see:

Of the 543M that is allocate with the new code:

InstanceInfo$Builder.setVIPAddress  10M
- this is due to the regex matching that I think is intended for eureka registration
- this could benefit from a deser specific implementation

All of the other allocations are either for the String cache or HashMaps used by Jackson.  My only further questions if what the memory allocation would be if we used Jackson's native de-serialization.  I might ask the developers to see if they expect more or less allocation with their deserializers.  I think it might be likely as now all of our allocation is in HashMap allocation.  I can't see how they need to long term create HashMaps if we're directly creating POJO's.

qiangdavidliu:
Looks good!

tbak:
>InstanceInfo$Builder.setVIPAddress 10M

I have added two more setters to ```InstanceInfo``` builder for vip/secure vip, that do not do macro expansion.

PR #473, opened 2015-03-20
2.x Eureka 1.x backwards compatible API with delta fetch support.

tbak:


eureka2-core/src/main/java/com/netflix/eureka2/interests/ChangeNotifications.java
```diff
@@ -55,9 +67,20 @@ public Boolean call(ChangeNotification<?> notification) {
...
+        @Override
+        public int compare(InstanceInfo o1, InstanceInfo o2) {
+            return o1.getId().compareTo(o2.getId());
+        }
+    };
+
     private ChangeNotifications() {
     }
 
+    public static Comparator<InstanceInfo> instanceInfoIdentityComparator() {
```
qiangdavidliu: Why not just make the static comparator public?

eureka2-ext/eureka2-eureka1-rest-api/src/main/java/com/netflix/eureka2/eureka1/rest/query/Eureka2RegistryViewCache.java
```diff
@@ -92,6 +87,20 @@ public Applications findApplicationsBySecureVip(final String secureVipAddress) {
...
+            Eureka2FullFetchWithDeltaView newView = new Eureka2FullFetchWithDeltaView(
+                    registry.forInterest(Interests.forFullRegistry()),
+                    refreshIntervalMs,
+                    scheduler
+            );
+            if (!fullFetchWithDeltaView.compareAndSet(null, newView)) {
+                newView.close();
+            }
+        }
+        return fullFetchWithDeltaView.get().latestCopy().timeout(queryTimeoutMs, TimeUnit.MILLISECONDS).take(1).toBlocking().first();
```
qiangdavidliu: Why toBlocking here? Also, if the timeout occurs and there is no element returned, this would result in an NPE?

eureka2-ext/eureka2-eureka1-rest-api/src/main/java/com/netflix/eureka2/eureka1/rest/query/Eureka2RegistryViewCache.java
```diff
@@ -92,6 +87,20 @@ public Applications findApplicationsBySecureVip(final String secureVipAddress) {
         return findApplicationsBy(secureVipAddress, Interests.forSecureVips(secureVipAddress), secureVipMap);
     }
 
+    private RegistryFetch getLatestFullFetch() {
+        if (fullFetchWithDeltaView.get() == null) {
```
qiangdavidliu: Why not inititalize this hot or just use synchronized? This seems more complex than needed.

eureka2-core/src/main/java/com/netflix/eureka2/interests/ChangeNotifications.java
```diff
@@ -158,4 +196,157 @@ private ChangeNotifications() {
...
+            final Comparator<T> identityComparator, final long interval, final TimeUnit timeUnit) {
+        return aggregateChanges(identityComparator, interval, timeUnit, Schedulers.computation());
+    }
+
+    /**
+     * It is a version of {@link #aggregateChanges} method, where the first item is emitted eagerly, and aggregation
+     * is applied only afterwards. This fits the Eureka registration model, where first batch will contain the full
+     * current registry content, which is followed by live updates.
+     */
+    public static <T> Transformer<List<ChangeNotification<T>>, List<ChangeNotification<T>>> emitAndAggregateChanges(
```
qiangdavidliu: Are all the layers of aggregateChanges and collapse necessary? It seems emitAndAggregateChanges() is the only function actually used. externally.

eureka2-core/src/main/java/com/netflix/eureka2/interests/ChangeNotifications.java
tbak: To align it with rest of the methods in this class. They are all factories for functions.

eureka2-ext/eureka2-eureka1-rest-api/src/main/java/com/netflix/eureka2/eureka1/rest/query/Eureka2RegistryViewCache.java
tbak: The values are returned from cache, but the initial query will actually block, which is bad. Thanks for spotting this.
We have two choices here:
1. populate view object with initial empty value or NO_VALUE marker
2. Rx-fy Eureka2RegistryViewCache API
I prefer the second route, and I will fix this accordingly.

eureka2-ext/eureka2-eureka1-rest-api/src/main/java/com/netflix/eureka2/eureka1/rest/query/Eureka2RegistryViewCache.java
tbak: I agree. All other methods use synchronize which is good enough here.

eureka2-core/src/main/java/com/netflix/eureka2/interests/ChangeNotifications.java
tbak: They build logically on each other: collapse -> aggregate -> aggregate with emit
There is also evaluate method in this spirit for list of change notifications. These functions are very useful when processing change notification streams, and I think we will have to provide them in some for as part of client API.

qiangdavidliu: If these are to be public API functions, then please make sure that they have more descriptive names and better java doc (especially the differences between the many flavours of functions that currently have the same name).

eureka2-core/src/main/java/com/netflix/eureka2/interests/ChangeNotifications.java
```diff
@@ -158,4 +201,153 @@ private ChangeNotifications() {
...
+    private static <T> List<ChangeNotification<T>> collapse(List<ChangeNotification<T>> notifications, Comparator<T> identityComparator) {
+        List<ChangeNotification<T>> result = new ArrayList<>();
+        collapse(notifications, new TreeMap<T, Integer>(identityComparator), result);
+        Collections.reverse(result);
+        return result;
+    }
+
+    private static <T> void collapse(List<ChangeNotification<T>> notifications, Map<T, Integer> markers, List<ChangeNotification<T>> result) {
+        for (int i = notifications.size() - 1; i >= 0; i--) {
+            ChangeNotification<T> next = notifications.get(i);
```
qiangdavidliu: Can we add a guard to ensure only data notifications are processed?

tbak: Added

eureka2-core/src/test/java/com/netflix/eureka2/interests/ChangeNotificationsTest.java
```diff
@@ -21,15 +28,42 @@
...
     private static final ChangeNotification<String> ADD_B = new ChangeNotification<>(Kind.Add, "B");
     private static final ChangeNotification<String> ADD_C = new ChangeNotification<>(Kind.Add, "C");
+    private static final ChangeNotification<String> MODIFY_D = new ChangeNotification<>(Kind.Modify, "D");
+    private static final ChangeNotification<String> DELETE_E = new ChangeNotification<>(Kind.Delete, "E");
+
+    private static final List<ChangeNotification<String>> FIRST_BATCH = asList(ADD_A, ADD_B);
+    private static final List<ChangeNotification<String>> SECOND_BATCH = asList(DELETE_A, ADD_C);
+
+    private static final List<ChangeNotification<String>> NOTIFICATION_LIST_1 = asList(
+            ADD_A, DELETE_A, ADD_A, ADD_B, MODIFY_A, MODIFY_D, DELETE_E
```
qiangdavidliu: For item A the modify here is not really tested as there is no real data modification, so collapse for meaningful modifies are not actually tested. Can we explicitly test for that?

eureka2-ext/eureka2-eureka1-rest-api/src/main/java/com/netflix/eureka2/eureka1/rest/Eureka1QueryRequestHandler.java
```diff
@@ -81,57 +85,88 @@ public Eureka1QueryRequestHandler(Eureka1Configuration config, ExtensionContext
...
-        }
-        return encodeResponse(format, gzip, response, v1InstanceInfo);
+    private Observable<Void> instanceGetByAppAndInstanceId(final String appName, final String instanceId, final EncodingFormat format,
+                                                           final boolean gzip, final HttpServerResponse<ByteBuf> response) {
+        return registryViewCache.findInstance(instanceId).materialize().toList().flatMap(
+                new Func1<List<Notification<InstanceInfo>>, Observable<Void>>() {
+                    @Override
+                    public Observable<Void> call(List<Notification<InstanceInfo>> notifications) {
+                        // If completed with error propagate it
+                        Notification<InstanceInfo> lastNotification = notifications.get(notifications.size() - 1);
```
qiangdavidliu: Are there any NPE potential here? Also, instead of doing a materialize().toList(), can't we just do a
```java
findInstance()
                .filter(new Func1<InstanceInfo, Boolean>() {
                    @Override
                    public Boolean call(InstanceInfo instanceInfo) {
                        // find and return true for the instance.
                    }
                })
                .materialize()
                .map(new Func1<Notification<InstanceInfo>, Void>() {
                    @Override
                    public Void call(Notification<InstanceInfo> instanceInfoNotification) {
                        // check for the different cases and do appropriate returns
                        return null;
                    }
                })
                .take(1);
```

eureka2-ext/eureka2-eureka1-rest-api/src/main/java/com/netflix/eureka2/eureka1/rest/query/Eureka2FullFetchWithDeltaView.java
```diff
@@ -0,0 +1,162 @@
...
+                        refreshIntervalMs, TimeUnit.MILLISECONDS,
+                        scheduler
+                ))
+                .map(new Func1<List<ChangeNotification<InstanceInfo>>, RegistryFetch>() {
+                    @Override
+                    public RegistryFetch call(List<ChangeNotification<InstanceInfo>> instanceInfos) {
+                        return updateSnapshot(instanceInfos);
+                    }
+                })
+                .subscribe(latestCopySubject);
```
qiangdavidliu: I think .replay(1).share() can work here too without having to use a subject.

eureka2-ext/eureka2-eureka1-rest-api/src/main/java/com/netflix/eureka2/eureka1/rest/query/Eureka2RegistryViewCache.java
```diff
@@ -1,124 +1,158 @@
...
-        return allApplicationsView.get().latestCopy().timeout(queryTimeoutMs, TimeUnit.MILLISECONDS).take(1).toBlocking().first();
+    public Observable<Applications> findAllApplications() {
+        return getLatestFullFetch().map(REGISTRY_FETCH_TO_APPLICATIONS_FUN);
+    }
+
+    public Observable<Applications> findAllApplicationsDelta() {
+        return getLatestFullFetch().map(REGISTRY_FETCH_TO_DELTA_FUN);
     }
 
-    public Application findApplication(String appName) {
```
qiangdavidliu: We should be able to get all the non-fullFetch variants (app, vip etc) just by going to the fullFetchView, and not need to cache.

eureka2-ext/eureka2-eureka1-rest-api/src/test/java/com/netflix/eureka2/eureka1/rest/query/Eureka2FullFetchWithDeltaViewTest.java
```diff
@@ -0,0 +1,138 @@
...
+        Applications delta = secondCopy.getDeltaChanges();
+
+        verifyApplicationsPresent(fullFetch, webAppName, backendName);
+        verifyApplicationsPresent(delta, backendName);
+
+        verifyHashCodes(fullFetch, delta);
+    }
+
+    @Test
+    public void testRemovedElementsDroppedFromApplicationsAndPresentInDelta() throws Exception {
```
qiangdavidliu: Is there a modify equivalent of this test (or is that tested together with one of the current testcases?)

eureka2-core/src/test/java/com/netflix/eureka2/interests/ChangeNotificationsTest.java
tbak: Added more data to cover this case

eureka2-ext/eureka2-eureka1-rest-api/src/main/java/com/netflix/eureka2/eureka1/rest/Eureka1QueryRequestHandler.java
tbak: We will always have at least one item. We have to do this that way for proper onComplete handling.

eureka2-ext/eureka2-eureka1-rest-api/src/main/java/com/netflix/eureka2/eureka1/rest/query/Eureka2FullFetchWithDeltaView.java
tbak: This will unsubscribe if there is no subscriber, and we want to cache it all the time, as building this view is very expensive.

eureka2-ext/eureka2-eureka1-rest-api/src/main/java/com/netflix/eureka2/eureka1/rest/query/Eureka2RegistryViewCache.java
tbak: Yes, having separate application views is not needed here.

eureka2-ext/eureka2-eureka1-rest-api/src/test/java/com/netflix/eureka2/eureka1/rest/query/Eureka2FullFetchWithDeltaViewTest.java
tbak: Modify is not explicitly tested. I will add test for that.

Issue #455, opened 2015-03-06
1.x discovery client treats HTTP redirects as errors


tbak:
Although discovery client properly handles HTTP request redirects, every returned status code other than 200 is regarded as an error.

PR #486, opened 2015-03-27
1.x Pin client to the redirected address.

tbak:
Initial implementation for discussion. Unit tests missing.

### DIFF
--- a/eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
+++ b/eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
@@ -112,7 +112,7 @@ public class DiscoveryClient implements LookupService {
 
     // Constants
     public static final int MAX_FOLLOWED_REDIRECTS = 10;
-    public static final String HTTP_X_NETFLIX_DISCOVERY_ALLOW_REDIRECT = "X-Netflix-Discovery-AllowRedirect";
+    public static final String HTTP_X_DISCOVERY_ALLOW_REDIRECT = "X-Discovery-AllowRedirect";
 
     private static final String VALUE_DELIMITER = ",";
     private static final String COMMA_STRING = VALUE_DELIMITER;
@@ -162,7 +162,8 @@ public class DiscoveryClient implements LookupService {
     private boolean isRegisteredWithDiscovery = false;
     private String discoveryServerAMIId;
     private JerseyClient discoveryJerseyClient;
-    private AtomicReference<String> lastRedirect = new AtomicReference<>();
+    private AtomicReference<String> lastQueryRedirect = new AtomicReference<>();
+    private AtomicReference<String> lastRegisterRedirect = new AtomicReference<>();
     private ApacheHttpClient4 discoveryApacheClient;
     protected static EurekaClientConfig clientConfig;
     private final AtomicReference<String> remoteRegionsToFetch;
@@ -1056,23 +1057,36 @@ public class DiscoveryClient implements LookupService {
      *             on any error.
      */
     private ClientResponse makeRemoteCall(Action action) throws Throwable {
-        if (lastRedirect.get() != null && isQueryAction(action)) {
-            String serviceUrl = lastRedirect.get();
+        ClientResponse response;
+        if(isQueryAction(action)) {
+            response = makeRemoteCallToRedirectedServer(lastQueryRedirect, action);
+        } else {
+            response = makeRemoteCallToRedirectedServer(lastRegisterRedirect, action);
+        }
+        if (response == null) {
+            response = makeRemoteCall(action, 0);
+        }
+        return response;
+    }
+
+    private ClientResponse makeRemoteCallToRedirectedServer(AtomicReference<String> lastRedirect, Action action) {
+        String lastRedirectUrl = lastRedirect.get();
+        if (lastRedirectUrl != null) {
             try {
-                ClientResponse clientResponse = makeRemoteCall(action, serviceUrl);
+                ClientResponse clientResponse = makeRemoteCall(action, lastRedirectUrl);
                 int status = clientResponse.getStatus();
                 if (status >= 200 && status < 300) {
                     return clientResponse;
                 }
                 SERVER_RETRY_COUNTER.increment();
-                lastRedirect.compareAndSet(serviceUrl, null);
+                lastRedirect.compareAndSet(lastRedirectUrl, null);
             } catch (Throwable ignored) {
                 logger.warn("Remote call to last redirect address failed; retrying from configured service URL list");
                 SERVER_RETRY_COUNTER.increment();
-                lastRedirect.compareAndSet(serviceUrl, null);
+                lastRedirect.compareAndSet(lastRedirectUrl, null);
             }
         }
-        return makeRemoteCall(action, 0);
+        return null;
     }
 
     private static boolean isQueryAction(Action action) {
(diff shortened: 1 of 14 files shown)


qiangdavidliu:
This should also fix issue #455

eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
```diff
@@ -1072,11 +1088,60 @@ private ClientResponse makeRemoteCall(Action action) throws Throwable {
...
+    }
+
+    private ClientResponse makeRemoteCallWithFollowRedirect(Action action, String serviceUrl) throws Throwable {
+        URI targetUrl = new URI(serviceUrl);
+        for(int followRedirectCount = 0; followRedirectCount < 10; followRedirectCount++) {
+            ClientResponse clientResponse = makeRemoteCall(action, targetUrl.toString());
+            if(clientResponse.getStatus() < 300) {
+                if(followRedirectCount > 0) {
+                    Matcher pathMatcher = REDIRECT_PATH_REGEX.matcher(targetUrl.getPath());
+                    if(pathMatcher.matches()) {
```
NiteshKant: Are there redirects from the server that we need to ignore?

eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
```diff
@@ -1072,11 +1088,60 @@ private ClientResponse makeRemoteCall(Action action) throws Throwable {
...
+                throw t;
+            }
+        }
+    }
+
+    private ClientResponse makeRemoteCallWithFollowRedirect(Action action, String serviceUrl) throws Throwable {
+        URI targetUrl = new URI(serviceUrl);
+        for(int followRedirectCount = 0; followRedirectCount < 10; followRedirectCount++) {
+            ClientResponse clientResponse = makeRemoteCall(action, targetUrl.toString());
+            if(clientResponse.getStatus() < 300) {
```
NiteshKant: Isn't the below code checking for redirects? Shouldn't this check be checking for code b/w 300-400 ?

tbak: We need to check for code 302. Status codes are analyzed by isOk method
which accepts <200,300) + 302 always, plus 403/404 for certain actions.
I have changed this condition to != 302 in another commit which I will push
soon.

eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
tbak: There shouldn't be, assuming server is in charge of making the decision
about that. I will add a header that will notify server side if a
particular client request can be redirected (by default no). Even if client
states yes, some requests can be handled directly by server (for example
full/delta fetches that we do not support on bootstrap cluster,
registrations, etc).

eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
```diff
@@ -110,16 +111,15 @@
     private static final DynamicPropertyFactory configInstance = DynamicPropertyFactory.getInstance();
 
     // Constants
-    private static final String DNS_PROVIDER_URL = "dns:";
-    private static final String DNS_NAMING_FACTORY = "com.sun.jndi.dns.DnsContextFactory";
-    private static final String JAVA_NAMING_FACTORY_INITIAL = "java.naming.factory.initial";
-    private static final String JAVA_NAMING_PROVIDER_URL = "java.naming.provider.url";
-    private static final String DNS_RECORD_TYPE = "TXT";
+    public static final int MAX_FOLLOWED_REDIRECTS = 10;
+    public static final String HTTP_X_NETFLIX_DISCOVERY_ALLOW_REDIRECT = "X-Netflix-Discovery-AllowRedirect";
```
qiangdavidliu: Remove the reference to Netflix and use a more generic header name?

eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
```diff
@@ -1056,9 +1056,29 @@ private void updateDelta(Applications delta) {
      *             on any error.
      */
     private ClientResponse makeRemoteCall(Action action) throws Throwable {
+        if (lastRedirect.get() != null && isQueryAction(action)) {
+            String serviceUrl = lastRedirect.get();
```
qiangdavidliu: Minor race here, as lastRedirect.get() may change in between the two calls. Can cache it locally first by String serviceUrl = lastRedirect.get() before the if statement.

eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
```diff
@@ -1072,11 +1092,70 @@ private ClientResponse makeRemoteCall(Action action) throws Throwable {
      * @throws Throwable
      *             on any error.
      */
-    private ClientResponse makeRemoteCall(Action action, int serviceUrlIndex)
-            throws Throwable {
+    private ClientResponse makeRemoteCall(Action action, int serviceUrlIndex) throws Throwable {
+        String serviceUrl;
+        try {
+            serviceUrl = eurekaServiceUrls.get().get(serviceUrlIndex);
```
qiangdavidliu: Can eurekaServiceUrls be null here?

eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
```diff
@@ -1072,11 +1092,70 @@ private ClientResponse makeRemoteCall(Action action) throws Throwable {
...
+            return makeRemoteCallWithFollowRedirect(action, serviceUrl);
+        } catch (Throwable t) {
+            if (eurekaServiceUrls.get().size() > ++serviceUrlIndex) {
```
qiangdavidliu: is eurekaServiceUrls.get() nullable?

eureka-client/src/main/java/com/netflix/discovery/shared/EurekaJerseyClient.java
```diff
@@ -218,6 +219,8 @@ public CustomApacheHttpClientConfig(String clientName, int maxConnectionsPerHost
             cm.setDefaultMaxPerRoute(maxConnectionsPerHost);
             cm.setMaxTotal(maxTotalConnections);
             getProperties().put(ApacheHttpClient4Config.PROPERTY_CONNECTION_MANAGER, cm);
+            getProperties().put(PROPERTY_FOLLOW_REDIRECTS, Boolean.FALSE);
```
qiangdavidliu: Can we add some comments on why the jerseyClient's redirects are explicitly turned off? Otherwise it may be confusing for future maintainers.

eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
tbak: I have dropped Netflix part from it.

eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
tbak: Good catch. I have fixed that.

eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
tbak: Validation is done during initialization and there must be at least one service Url.

eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
tbak: See note above.There should be always at least one entry.

eureka-client/src/main/java/com/netflix/discovery/shared/EurekaJerseyClient.java
tbak: Done

PR #490, opened 2015-03-31
2.x bridging fixes

qiangdavidliu:


eureka2-ext/eureka2-eureka1-rest-api/src/main/java/com/netflix/eureka2/eureka1/rest/query/Eureka2FullFetchWithDeltaView.java
```diff
@@ -139,19 +140,32 @@ public RegistryFetch apply(List<ChangeNotification<InstanceInfo>> updates) {
 
             for (ChangeNotification<InstanceInfo> update : updates) {
                 com.netflix.appinfo.InstanceInfo v1Update = toEureka1xInstanceInfo(update.getData());
+                if (v1Update== null) {
```
tbak: Is it possible that v1Update is null? It should never happen in properly working app.

eureka2-server/build.gradle
```diff
@@ -23,6 +23,7 @@ dependencies {
     compile "org.slf4j:slf4j-simple:${slf4j_version}"
     compile "commons-cli:commons-cli:${commons_cli_version}"
 
+    compile "org.codehaus.jettison:jettison:${jettison_version}"
```
tbak: why this dependency?

eureka2-ext/eureka2-eureka1-rest-api/src/main/java/com/netflix/eureka2/eureka1/rest/query/Eureka2FullFetchWithDeltaView.java
qiangdavidliu: see change to toEureka1xInstanceInfo() which catches exceptions during conversion and returns null.

eureka2-server/build.gradle
qiangdavidliu: karyon-admin brings in jettison 1.1, which does not work with serializing v1.Applications. (http://jettison.codehaus.org/)

PR #494, opened 2015-04-08
Add eureka2-eureka1-support module

tbak:
Add eureka2-eureka1-support module to help bridge existing
Eureka1 clients to Eureka2 without adopting Eureka2 client API yet.

### DIFF
--- a/eureka2-ext/eureka2-eureka1-support/src/main/java/com/netflix/eureka2/eureka1/utils/ServerListReader.java
+++ b/eureka2-ext/eureka2-eureka1-support/src/main/java/com/netflix/eureka2/eureka1/utils/ServerListReader.java
@@ -1,0 +1,86 @@
+package com.netflix.eureka2.eureka1.utils;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
+
+import com.netflix.eureka2.client.EurekaInterestClient;
+import com.netflix.eureka2.client.Eurekas;
+import com.netflix.eureka2.client.functions.InterestFunctions;
+import com.netflix.eureka2.client.resolver.ServerResolver;
+import com.netflix.eureka2.interests.Interest;
+import com.netflix.eureka2.interests.Interests;
+import com.netflix.eureka2.registry.instance.InstanceInfo;
+import org.slf4j.Logger;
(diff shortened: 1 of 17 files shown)


eureka2-ext/eureka2-eureka1-support/src/main/java/com/netflix/eureka2/eureka1/utils/ServerListReader.java
```diff
@@ -0,0 +1,86 @@
...
+                .subscribe();
+    }
+
+    public List<com.netflix.appinfo.InstanceInfo> getLatestServerList() {
+        return latestServerList.get();
+    }
+
+    public List<com.netflix.appinfo.InstanceInfo> getLatestServerListOrWait() {
+        try {
+            firstBatchLatch.await(FIRST_RESOLVE_TIMEOUT_SEC, TimeUnit.SECONDS);
```
qiangdavidliu: What happens if the latch does not count down after the timeout (i..e. the first batch is still empty)? Since legacy code does not do well with an empty list, is a null return or an exception better here?

eureka2-ext/eureka2-eureka1-support/src/main/java/com/netflix/eureka2/eureka1/utils/ServerListReader.java
```diff
@@ -0,0 +1,86 @@
...
+/**
+ * Helper class for integrating existing Eureka1 clients to Eureka2.
+ *
+ * @author Tomasz Bak
+ */
+public class ServerListReader {
+
+    private static final Logger logger = LoggerFactory.getLogger(ServerListReader.class);
+
+    public static final int FIRST_RESOLVE_TIMEOUT_SEC = 30;
```
qiangdavidliu: Can we have this as a second constructor option?

eureka2-ext/eureka2-eureka1-support/src/main/java/com/netflix/eureka2/eureka1/utils/ServerListReader.java
tbak: Some change in the legacy code will be required. We will see how much, and what is best way of supporting it. This is just first attempt.

eureka2-ext/eureka2-eureka1-support/src/main/java/com/netflix/eureka2/eureka1/utils/ServerListReader.java
tbak: Yes, I will add it.

PR #500, opened 2015-04-29
2.x Implement feature #447 (Write Server Initial Sync).

tbak:


eureka2-core/src/main/java/com/netflix/eureka2/health/AbstractHealthStatusProvider.java
```diff
@@ -34,6 +34,4 @@ public boolean moveHealthTo(Status newStatus) {
         healthSubject.onNext(new HealthStatusUpdate<SUBSYSTEM>(newStatus, descriptor));
         return true;
     }
-
-    protected abstract Status toEurekaStatus(Status status);
```
qiangdavidliu: With this gone, should this class still be abstract?

eureka2-server/src/main/java/com/netflix/eureka2/server/resolver/EurekaEndpoint.java
```diff
@@ -0,0 +1,101 @@
...
+            default: // == Replication
+                return getReplicationPort();
+        }
+    }
+
+    public static EurekaEndpoint writeServerEndpointFrom(String domainName, int registrationPort, int interestPort, int replicationPort) {
+        return new EurekaEndpoint(domainName, registrationPort, interestPort, replicationPort);
+    }
+
+    public static EurekaEndpoint writeServerEndpointFrom(String hostnameAndPorts) {
```
qiangdavidliu: be nice to have javadoc here to specify the expected string format for `hostnameAndPorts`

eureka2-server/src/main/java/com/netflix/eureka2/server/resolver/DnsEurekaEndpointResolver.java
```diff
@@ -0,0 +1,69 @@
...
+                            return new ChangeNotification<EurekaEndpoint>(Kind.Add, endpoint);
+                        }
+                        return new ChangeNotification<EurekaEndpoint>(Kind.Delete, endpoint);
+                    }
+                });
+    }
+
+    protected abstract EurekaEndpoint createEndpoint(String hostName);
+
+    public static DnsEurekaEndpointResolver writeServerDnsResolver(final String domainName,
```
qiangdavidliu: These statics seems redundant as we have a factory class `EurekaEndpointResolvers` that provides the same functionality.

eureka2-server/src/main/java/com/netflix/eureka2/server/transport/tcp/discovery/TcpDiscoveryHandler.java
```diff
@@ -43,17 +52,74 @@
...
+                            }
+
+                            @Override
+                            public void onError(Throwable e) {
+                                // This should never happen
+                                logger.error("Healthcheck stream terminated with an error and went out of retry loop", e);
+                            }
+
+                            @Override
+                            public void onNext(HealthStatusUpdate<EurekaHealthStatusAggregator> update) {
```
qiangdavidliu: Seems like we should have some more logic here that ties in the realtime server health status to whether we should server interest data. Whenever the server status changes to not be UP (which may happen outside of bootstrap time) we should also stop serving interest data.

eureka2-testkit/src/main/java/com/netflix/eureka2/testkit/embedded/cluster/EmbeddedWriteCluster.java
```diff
@@ -134,7 +136,18 @@ public Integer call(WriteServerAddress writeServerAddress) {
...
-                        Server serverAddress = new Server(data.getHostName(), data.getReplicationPort());
+                        int port;
+                        switch (serviceType) {
+                            case Registration:
+                                port = data.getRegistrationPort();
+                                break;
+                            case Interest:
+                                port = data.getDiscoveryPort();
+                                break;
+                            default: // == Replication
```
qiangdavidliu: Can we change our code pattern to always explicitly check for all cases and defaults to some sort of breaking error? Treating default as a specific valid case can potentially lead to missed updates when an enum is added for example, and thus lead to non-deterministic behaviour.

eureka2-write-server/src/main/java/com/netflix/eureka2/server/PeerAddressesProvider.java
```diff
@@ -0,0 +1,71 @@
...
+
+    public PeerAddressesProvider(Observable<ChangeNotification<Server>> addressStream) {
+        this.config = null;
+        this.serviceType = null;
+        this.addressStream = addressStream;
+    }
+
+    @PostConstruct
+    public void createResolver() {
+        if (config != null) {  // always postConstruct resolve if config exist
```
qiangdavidliu: This whole block doesn't seem to do any heavy lifting, why not just move this to the constructor?

eureka2-core/src/main/java/com/netflix/eureka2/health/AbstractHealthStatusProvider.java
tbak: Syntactically not, but conceptually yes, so I would keep it that way. We may also reintroduce some abstract methods in the future.

eureka2-write-server/src/main/java/com/netflix/eureka2/server/service/bootstrap/PeerRegistryBootstrapService.java
```diff
@@ -0,0 +1,118 @@
...
+import rx.Observable;
+import rx.Scheduler;
+import rx.functions.Func1;
+import rx.schedulers.Schedulers;
+
+/**
+ * @author Tomasz Bak
+ */
+@Singleton
+public class PeerRegistryBootstrapService implements RegistryBootstrapService {
```
qiangdavidliu: Nitpick, peer in the name here assumes bootstrapping from peer write nodes, whereas we can bootstrap from any arbitrary eureka source based on configuration.

eureka2-write-server/src/main/java/com/netflix/eureka2/server/service/bootstrap/PeerRegistryBootstrapService.java
```diff
@@ -0,0 +1,118 @@
...
+                            return Observable.empty();
+                        }
+                        logger.error("Cannot bootstrap registry from peers");
+                        return Observable.error(new Exception("Cannot bootstrap registry from peers"));
+                    }
+                });
+    }
+
+    private Observable<Boolean> loadRegistryFrom(final Server server, final SourcedEurekaRegistry<InstanceInfo> registry, final Source source) {
+        return createLightEurekaInterestClient(server).forInterest(Interests.forFullRegistry())
```
qiangdavidliu: This here seems to be doing almost exactly what the FullFetchInterestClient does for the read servers. Can we not reuse the FullFetchInterestClient and pass it the write server's registry?

eureka2-write-server/src/main/java/com/netflix/eureka2/server/service/bootstrap/RegistryBootstrapCoordinator.java
```diff
@@ -0,0 +1,99 @@
...
+    private static final Logger logger = LoggerFactory.getLogger(RegistryBootstrapCoordinator.class);
+
+    private static final SubsystemDescriptor<RegistryBootstrapCoordinator> DESCRIPTOR = new SubsystemDescriptor<>(
+            RegistryBootstrapCoordinator.class,
+            "Write Server registry bootstrap",
+            "Bootstrap registry loader from external source"
+    );
+
+    /**
+     * TODO Make bootstrap timeout configurable
```
qiangdavidliu: Let's do this and add it to write server config:)

eureka2-write-server/src/main/java/com/netflix/eureka2/server/service/bootstrap/PeerRegistryBootstrapService.java
```diff
@@ -0,0 +1,118 @@
...
+                    public Observable<EurekaEndpoint> call(LinkedHashSet<EurekaEndpoint> eurekaEndpoints) {
+                        return Observable.from(eurekaEndpoints);
+                    }
+                }).flatMap(new Func1<EurekaEndpoint, Observable<Boolean>>() {
+                    @Override
+                    public Observable<Boolean> call(EurekaEndpoint endpoint) {
+                        return loadRegistryFrom(new Server(endpoint.getHostname(), endpoint.getInterestPort()), registry, source);
+                    }
+                })
+                .filter(new Func1<Boolean, Boolean>() { // true - download succeeded, false - failed, so try another server
```
qiangdavidliu: The comment mentioned retrying on another server but I can't find any .retry() in the code?

eureka2-write-server/src/test/java/com/netflix/eureka2/server/service/bootstrap/PeerRegistryBootstrapServiceTest.java
```diff
@@ -0,0 +1,78 @@
...
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.assertThat;
+import static org.mockito.Matchers.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+/**
+ * @author Tomasz Bak
+ */
+public class PeerRegistryBootstrapServiceTest {
```
qiangdavidliu: Would be nice to have additional tests to cover:
- bootstrap fail on first source but succeeds on subsequent source

eureka2-server/src/main/java/com/netflix/eureka2/server/resolver/EurekaEndpoint.java
tbak: Added

eureka2-server/src/main/java/com/netflix/eureka2/server/resolver/DnsEurekaEndpointResolver.java
tbak: Statics are removed from DnsEurekaEndpointResolver, which is package private now.

eureka2-server/src/main/java/com/netflix/eureka2/server/transport/tcp/discovery/TcpDiscoveryHandler.java
tbak: I would keep logic simple for now. Also even if status is not UP anymore, it might be very risky to start discarding connections, unless we have a notion of total/partial failures.

eureka2-testkit/src/main/java/com/netflix/eureka2/testkit/embedded/cluster/EmbeddedWriteCluster.java
tbak: Done

eureka2-write-server/src/main/java/com/netflix/eureka2/server/PeerAddressesProvider.java
tbak: I agree. I moved it to constructor.

eureka2-write-server/src/main/java/com/netflix/eureka2/server/service/bootstrap/PeerRegistryBootstrapService.java
tbak: As discussed, we cannot use here directly FullFetchInterestClient, but we will refactor this code base, so we have more flexibility.

eureka2-write-server/src/main/java/com/netflix/eureka2/server/service/bootstrap/RegistryBootstrapCoordinator.java
tbak: Done

eureka2-write-server/src/main/java/com/netflix/eureka2/server/service/bootstrap/PeerRegistryBootstrapService.java
tbak: Bootstrap status can be true or false. We have here a filter that should terminate as soon as bootstrap == true is received. This implementation is however flawed. I have reimplemented it.

eureka2-write-server/src/test/java/com/netflix/eureka2/server/service/bootstrap/PeerRegistryBootstrapServiceTest.java
tbak: I have added more tests to cover different failure scenarios.

eureka2-write-server/src/main/java/com/netflix/eureka2/server/service/bootstrap/PeerRegistryBootstrapService.java
tbak: Renamed to BackupClusterBootstrapService.

PR #505, opened 2015-05-02
Rearrange sample programs so they can be run correctly from Gradle. 

qu1j0t3:
Also change Eureka server port to 8080 to match Tomcat default.


tbak:
Thank you for your PR. To align with other Netflix OSS projects (including eureka2), could you put both client and server into single gradle module ```eureka-examples```? It will probably require some extra gradle scripting as application plugin does not support multiple apps by default, but it will be more concise.

qu1j0t3:
Not sure I agree. This PR intentionally reflects that client and service are distinct programs, and the Gradle script is consequently minimal, fits the expected pattern. Are you **certain** you want to go that way? (I already tried to do this within the `eureka-server` module in early attempts to get the demo to run properly, and as you intuited, the Gradle was obviously much less natural—and actually never worked.)

tbak:
Give me some time to think about it, and try it out. I would like to
minimize number of gradle submodules.

qiangdavidliu:
@qu1j0t3 we are working on some cleanup of eureka1 to ease migration for eureka2, and part of that will include some fixing of these examples. The changes should be done pretty soon, and hopefully with that the examples should be much more clear and runnable.

qu1j0t3:
@qiangdavidliu Thanks. I couldn't come up with a simpler result than this proposal, so I hope you'll consider its structure.

qiangdavidliu:
@qu1j0t3 please take a look at #515 which also cleans up the examples. Specifically the new examples module readme with updated instructions https://github.com/qiangdavidliu/eureka/blob/master/eureka-examples/README.md and the updated config files.

qu1j0t3:
OK. Abandoning this PR. :)

qiangdavidliu:
@qu1j0t3 the pull has been merged. Let us know if you are still having trouble with the new examples.

PR #506, opened 2015-05-04
2.x Removing dependency on Ocelli from the client (for now)

qiangdavidliu:
Removing dependency on Ocelli (for now, until we are ready to integrate with it and use it's more advanced features). The renamed RoundRobinServerResolver functionally does the same thing as before with an internal round robin impl.


eureka2-client/src/main/java/com/netflix/eureka2/client/resolver/RoundRobinServerResolver.java
```diff
@@ -0,0 +1,143 @@
...
+
+    private final Observable<ChangeNotification<Server>> serverSource;
+    private final AtomicReference<List<Server>> serverCacheRef;
+    private final AtomicInteger positionRef;
+
+    protected RoundRobinServerResolver(Server... servers) {
+        this(sourceFromList(servers), 10, TimeUnit.SECONDS);
+    }
+
+    protected RoundRobinServerResolver(Observable<ChangeNotification<Server>> serverSource) {
```
tbak: Warmup time was a workaround needed prior to introducing buffer sentinels. As we have those now, I think we do not need it anymore.

eureka2-client/src/main/java/com/netflix/eureka2/client/resolver/RoundRobinServerResolver.java
```diff
@@ -0,0 +1,143 @@
...
+
+    /**
+     * Connect and warm up the serverCache to the server source. onComplete the result Observable once the
+     * serverCache has warmed up.
+     */
+    private Observable<List<Server>> refreshServerCache() {
+        return serverSource
+                .compose(ChangeNotifications.<Server>buffers())
+                .compose(ChangeNotifications.<Server>snapshots())
+                .materialize()
```
tbak: Recent refactorings in ChangeNotifications allow for empty server lists. I think in the very beginning we should not emit empty list (filter empty lists out), but we should allow them afterwards, which means that all previously advertised server gone away.
The bottom line is that ChangeNotifications transformers must offer enough flexibility to do that.

eureka2-client/src/main/java/com/netflix/eureka2/client/resolver/RoundRobinServerResolver.java
```diff
@@ -0,0 +1,143 @@
...
+                .concatMap(new Func1<Notification<LinkedHashSet<Server>>, Observable<? extends List<Server>>>() {
+                    @Override
+                    public Observable<? extends List<Server>> call(Notification<LinkedHashSet<Server>> rxNotification) {
+                        switch (rxNotification.getKind()) {
+                            case OnNext:
+                                LinkedHashSet<Server> servers = rxNotification.getValue();
+                                if (servers.isEmpty()) {  // if onNext is empty, do nothing and wait
+                                    return Observable.never();
```
tbak: This observable never completes, which will always result in a timeout below. We should use filter to get rid of empty server lists or just emit Observable.empty(). ```take(1)``` below will wait for first non-empty list.

eureka2-client/src/main/java/com/netflix/eureka2/client/resolver/RoundRobinServerResolver.java
```diff
@@ -0,0 +1,143 @@
...
+    @Override
+    public Observable<Server> resolve() {
+        return refreshServerCache().concatMap(new Func1<List<Server>, Observable<? extends Server>>() {
+            @Override
+            public Observable<? extends Server> call(List<Server> servers) {
+                if (servers.isEmpty()) {
+                    return Observable.error(SERVER_CACHE_EMPTY_EXCEPTION);
+                }
+
+                int currentPos = positionRef.getAndIncrement();
```
tbak: I would do this module servers.size() to avoid Integer overflow. I know it is very unlikely, but still it is better to be strict if there is no cost to it.

eureka2-client/src/main/java/com/netflix/eureka2/client/resolver/RoundRobinServerResolver.java
```diff
@@ -0,0 +1,143 @@
...
+                        }
+                        return Observable.just(serverCacheRef.get());
+                    }
+                })
+                .take(1);
+    }
+
+    private static Observable<ChangeNotification<Server>> sourceFromList(Server... servers) {
+        return Observable.from(servers)
+                .map(new Func1<Server, ChangeNotification<Server>>() {
```
tbak: Each subscription will create a new list/change notification objects. As ChangeNotification is not mutable we can create it eagerly or add share operator.
I have seen mapping of T to ChangeNotification<T> in many places. Can we have a generic implementation in ChangeNotifications?

eureka2-client/src/main/java/com/netflix/eureka2/client/resolver/RoundRobinServerResolver.java
qiangdavidliu: Doing the modulo on the atomicInt value and then setting it means the op will no longer be atomic, in this case I guess it doesn't matter that much as concurrent usage is unlikely, but I'd like to avoid it still. How about changing it to AtomicLong instead.

eureka2-client/src/main/java/com/netflix/eureka2/client/resolver/RoundRobinServerResolver.java
qiangdavidliu: We still need the warmup time as a cap on the time to wait for the initial case when there are no data from the serverSource.

eureka2-client/src/main/java/com/netflix/eureka2/client/resolver/RoundRobinServerResolver.java
qiangdavidliu: Actually even if the atomicInt overflows we are still good, so no need to change this here.

eureka2-client/src/main/java/com/netflix/eureka2/client/resolver/RoundRobinServerResolver.java
qiangdavidliu: Done

PR #507, opened 2015-05-04
2.x Safer implementation of ack handling + pending ack queue size monitor.

tbak:
Bug fix in EurekaHealthStatusModule module.

### DIFF
--- a/eureka2-core/src/main/java/com/netflix/eureka2/transport/base/BaseMessageConnection.java
+++ b/eureka2-core/src/main/java/com/netflix/eureka2/transport/base/BaseMessageConnection.java
@@ -38,8 +38,8 @@ import rx.functions.Action0;
 import rx.functions.Action1;
 import rx.functions.Func1;
 import rx.schedulers.Schedulers;
-import rx.subjects.PublishSubject;
-import rx.subjects.ReplaySubject;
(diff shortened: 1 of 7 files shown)


eureka2-core/src/main/java/com/netflix/eureka2/transport/base/BaseMessageConnection.java
```diff
@@ -267,29 +295,73 @@ public void call() {
...
-                        pendingAck.add(ack);
+                        pendingAckQueue.add(ack);
+                        metrics.incrementPendingAckCounter();
                         metrics.incrementOutgoingMessageCounter(message.getClass(), 1);
+
+                        // Connection might be closed when we serve this request, and since
+                        // pendingAckQueue and closed variables are not changed together atomically, we check
+                        // close status after adding item to pendingAckQueue, and optionally drain it.
+                        if (closed.get()) {
+                            drainPendingAckQueue();
```
qiangdavidliu: Is this threadsafe? Seems like this can be called by multiple observable chains concurrently here.

tbak: Yes, it is. I added a corresponding comment for drainPendingAckQueue()
method.

PR #509, opened 2015-05-06
2.x Improvements in transport + performance/verification tool.

tbak:


eureka2-client/src/main/java/com/netflix/eureka2/client/registration/EurekaRegistrationClientImpl.java
```diff
@@ -84,6 +88,6 @@ public void call() {
     @Override
     public void shutdown() {
         logger.info("Shutting down RegistrationClient");
-        // nothing to shutdown
+        shutdownSubject.onCompleted();
```
qiangdavidliu: Let's guard this with an atomic state boolean to observe the rx contract and not call onCompleted more than once.

eureka2-client/src/main/java/com/netflix/eureka2/client/transport/ResolverBasedTransportClient.java
```diff
@@ -76,7 +79,7 @@ public MessageConnection call(
                                             ObservableConnection<Object, Object> conn) {
                                         return new SelfClosingConnection(
                                                 new HeartBeatConnection(
-                                                        new BaseMessageConnection("client", conn, metrics),
+                                                        new BaseMessageConnection(clientId, conn, metrics),
```
qiangdavidliu: If this is purely for testing purposes, it might be useful to define a counting id in the client, and append the counting id to the end of this to differentiate between channel generations.

eureka2-client/src/test/java/com/netflix/eureka2/client/registration/EurekaRegistrationClientImplTest.java
```diff
@@ -313,6 +313,31 @@ public RegistrationChannel answer(InvocationOnMock invocation) throws Throwable
     }
 
     @Test
+    public void testUnregisterOnClientShutdown() throws Exception {
```
qiangdavidliu: Good catch!

eureka2-core/src/main/java/com/netflix/eureka2/transport/base/BaseMessageConnection.java
```diff
@@ -251,6 +261,22 @@ public void onNext(Void aVoid) {
         }
     }
 
+    private <E extends Exception> E createException(Class<E> type, String message) {
+        String enrichedMessage = "{connection=" + name + "} " + message;
+        E exception;
+        try {
+            Constructor<E> constructor = type.getConstructor(String.class);
```
qiangdavidliu: Can we change this to not use reflection and avoid the performance penalty?

eureka2-client/src/main/java/com/netflix/eureka2/client/transport/ResolverBasedTransportClient.java
tbak: It is useful to have more descriptive names, like registeringClient, lightInterestClient, etc. Ids will be still cryptic.

eureka2-core/src/main/java/com/netflix/eureka2/transport/base/BaseMessageConnection.java
tbak: Yes, I was planning to do so, and missed that. This also pollutes stack trace.

eureka2-client/src/main/java/com/netflix/eureka2/client/transport/ResolverBasedTransportClient.java
qiangdavidliu: By id I meant as an append to the clientName, since we can have multiple generations of MessageConnection per client.

eureka2-core/src/main/java/com/netflix/eureka2/utils/ExceptionUtils.java
```diff
@@ -0,0 +1,35 @@
...
+     */
+    public static <E extends Exception> E detachedExceptionOf(Class<E> type, String message) {
+        StackTraceElement[] stackTrace = Thread.currentThread().getStackTrace();
+        if (stackTrace.length > 2) {
+            stackTrace = new StackTraceElement[]{stackTrace[2]};
+        }
+        E exception;
+        try {
+            Constructor<E> constructor = type.getConstructor(String.class);
+            exception = constructor.newInstance(message);
```
qiangdavidliu: Again change from using reflection?

eureka2-read-server/src/main/java/com/netflix/eureka2/server/FullFetchInterestClientProvider.java
```diff
@@ -22,12 +22,15 @@
...
 import com.netflix.eureka2.server.interest.FullFetchBatchingRegistry;
 import com.netflix.eureka2.server.interest.FullFetchInterestClient;
+import com.netflix.eureka2.transport.TransportClient;
 
 /**
  * @author Tomasz Bak
  */
 public class FullFetchInterestClientProvider implements Provider<FullFetchInterestClient> {
 
+    private static final String EUREKA_READ_CLIENT_ID = "eurekaReadClient";
```
qiangdavidliu: How about change all the default client names to something like client(InterestClient, FullFetchClient).class.getSimpleName()? That way they can be nice and programmatic.

eureka2-server/src/main/java/com/netflix/eureka2/server/channel/InterestChannelImpl.java
```diff
@@ -102,11 +102,14 @@ public void call(Object message) {
             initializeNotificationMultiplexer();
         }
 
+        // We acknowledge before subscription, as subscribing may generate notification stream
```
qiangdavidliu: Good catch

eureka2-client/src/main/java/com/netflix/eureka2/client/registration/EurekaRegistrationClientImpl.java
tbak: Async subject provides guard here.

eureka2-client/src/main/java/com/netflix/eureka2/client/transport/ResolverBasedTransportClient.java
tbak: Good idea. I have added it.

eureka2-core/src/main/java/com/netflix/eureka2/transport/base/BaseMessageConnection.java
tbak: Done

eureka2-core/src/main/java/com/netflix/eureka2/utils/ExceptionUtils.java
tbak: I will replace this with stack trimming function.

eureka2-read-server/src/main/java/com/netflix/eureka2/server/FullFetchInterestClientProvider.java
tbak: We might use the same class in different contexts. In this particular case it would work, but might not work in the other. I propose it keep it as an arbitrary name. In Eureka client it is even possible to set it explicitly via transport config.

PR #511, opened 2015-05-11
Fix typos in per-host config keys.

qu1j0t3:


tbak:
Thank you for spotting it. We cannot merge this however, as this change is not backwards compatible (all user would have to update their configuration files). It is safer to leave it as it is, even if the property name is a bit misleading.

toby5box:
Their configuration files are either setting the right parameter, in which case they do not need to change, or they are setting the wrong parameter, in which case they are wrong. However, it seems very unlikely that any client is using the perHost parameters or they would have already noticed the bug.

So this seems safely mergeable.

tbak:
You are right. I have missed that we have also ```getPeerNodeTotalConnections```, and the property names overlap:

```java
 @Override
    public int getPeerNodeTotalConnections() {
        return configInstance.getIntProperty(
                namespace + "peerNodeTotalConnections", 1000).get();
    }

    @Override
    public int getPeerNodeTotalConnectionsPerHost() {
        return configInstance.getIntProperty(
                namespace + "peerNodeTotalConnectionsPerHost", 500).get();
    }
```

 I definitely makes sense to merge this in. Thanks.

qu1j0t3:
No worries.

PR #515, opened 2015-05-14
Adding EurekaModule for DI support + Example cleanup

qiangdavidliu:
- EurekaModule for DI support
- Added EurekaClient interface for DiscoveryClient
- cleaned up examples and moved examples to eureka-examples


eureka-client/src/main/java/com/netflix/appinfo/providers/MyDataCenterInstanceConfigProvider.java
```diff
@@ -6,6 +6,9 @@
 import com.netflix.discovery.DiscoveryManager;
 import com.netflix.discovery.EurekaNamespace;
 
+import javax.inject.Singleton;
+
+@Singleton
```
elandau: @Singleton is unnecessary here.   Making the Provider @Singleton does not make MyDataCenterInstanceConfig a singleton.  You need to do that when you specify the binding.  This should also be a provider for EurekaInstanceConfig and not MyDataCenterInstanceConfig.

bind(EurekaInstanceConfig.class).toProvider(MyDataCenterInstanceConfigProvider.class).in(Scopes.SINGLETON)

eureka-client/src/main/java/com/netflix/discovery/DefaultEurekaClientConfig.java
```diff
@@ -55,6 +56,7 @@
  * @author Karthik Ranganathan
  *
  */
+@FineGrainedLazySingleton
 @ProvidedBy(DefaultEurekaClientConfigProvider.class)
```
elandau: Add @Singleton

eureka-client/src/main/java/com/netflix/discovery/providers/DefaultEurekaClientConfigProvider.java
```diff
@@ -6,10 +6,13 @@
 import com.netflix.discovery.EurekaClientConfig;
 import com.netflix.discovery.EurekaNamespace;
 
+import javax.inject.Singleton;
+
 /**
  * This provider is necessary because the namespace is optional.
  * @author elandau
  */
+@Singleton
```
elandau: @Singleton not necessary

eureka-client/src/main/java/com/netflix/appinfo/InstanceInfo.java
```diff
@@ -1153,4 +1153,30 @@ private static String resolveDeploymentContextBasedVipAddresses(
...
+     * Get the zone that a particular instance is in.
+     * Note that for AWS deployments, myInfo should contain AWS dataCenterInfo which should contain
+     * the AWS zone of the instance, and availZones is ignored.
+     *
+     * @param availZones the list of available zones for non-AWS deployments
+     * @param myInfo
+     *            - The InstanceInfo object of the instance.
+     * @return - The zone in which the particular instance belongs to.
+     */
+    public static String getZone(String[] availZones, InstanceInfo myInfo) {
```
tbak: Why this is not instance method of InstanceInfo, but static method?

eureka-client/src/main/java/com/netflix/discovery/converters/EurekaJacksonCodec.java
```diff
@@ -230,7 +230,7 @@ public Name getName() {
                 };
             }
 
-            Map<String, String> metaData = new HashMap<>();
+            Map<String, String> metaData = new HashMap<String, String>();
```
tbak: We are at Java7 language level, so adding type parameters here on the left side is redundant.

eureka-client/src/test/java/com/netflix/discovery/guice/EurekaModuleTest.java
```diff
@@ -0,0 +1,86 @@
...
+
+    private Injector injector;
+    private LifecycleManager lifecycleManager;
+
+    @Before
+    public void setUp() throws Exception {
+        ConfigurationManager.getConfigInstance().setProperty("eureka.region", "default");
+        ConfigurationManager.getConfigInstance().setProperty("eureka.shouldFetchRegistry", "false");
+        ConfigurationManager.getConfigInstance().setProperty("eureka.registration.enabled", "false");
+        ConfigurationManager.getConfigInstance().setProperty("eureka.serviceUrl.default", "http://localhost:8080/eureka/v2");
```
tbak: Can we use EurekaClientConfig mock here, and avoid using archaius singleton?

eureka-core/src/main/java/com/netflix/eureka/PeerAwareInstanceRegistry.java
```diff
@@ -209,9 +208,14 @@ public void run() {
      */
     private void updatePeerEurekaNodes() {
         InstanceInfo myInfo = ApplicationInfoManager.getInstance().getInfo();
-        List<String> replicaUrls = DiscoveryManager.getInstance()
-                .getDiscoveryClient()
-                .getDiscoveryServiceUrls(DiscoveryClient.getZone(myInfo));
+
+        EurekaClientConfig clientConfig = DiscoveryManager.getInstance().getEurekaClientConfig();
```
tbak: PeerAwareInstanceRegistry is still a singleton. It should be easy to convert it to guice singleton, and inject all required dependencies.

eureka-client/src/main/java/com/netflix/discovery/converters/EurekaJacksonCodec.java
qiangdavidliu: Actually most of Eureka (1.0) is at Java6, and since it has some pretty old users it would be nice to keep it that way.

eureka-client/src/test/java/com/netflix/discovery/guice/EurekaModuleTest.java
qiangdavidliu: We'd have to mock a few different configs. This also allows us to test the actual config bindings. The test runs pretty fast even with the archaius singleton.

eureka-core/src/main/java/com/netflix/eureka/PeerAwareInstanceRegistry.java
qiangdavidliu: Agreed. But let's do this in a separate pull and not conflate the current changes.

eureka-client/src/main/java/com/netflix/appinfo/InstanceInfo.java
qiangdavidliu: That's true, it could be. It was mostly because this was just a transport of the method from the client.

PR #522, opened 2015-05-19
Example update

qiangdavidliu:


datvikash:
@qiangdavidliu thanks! this really helped clear up the distinction between client initialization and service registration :rocket:

PR #524, opened 2015-05-22
Add event control for initial instance info heartbeat and instanceinfo update (UP status typically)

aspyker:
This is the patch that allows an event oriented approach to starting InstanceInfo scheduled tasks (general HeartBeat and InstanceInfo (HealthCheck/Status/etc)).  It should be backwardsly compatible with no change in behavior (by default the tasks are started the same delays as before from DiscoveryClient initialization).  If however, the app sets the useExplicitEventsForInitialDiscoveryCalls, then the client is expected to call the methods startInstanceInfoHeartBeatNow and startInstanceInfoRefreshNow, likely calling them far sooner than the default initial delay.

### DIFF
--- a/eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
+++ b/eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
@@ -324,7 +324,14 @@ public class DiscoveryClient implements EurekaClient {
         if (clientConfig.shouldFetchRegistry() && !fetchRegistry(false)) {
             fetchRegistryFromBackup();
         }
-        initScheduledTasks();
+
+        initRegistryFetchPoller();
+        boolean eventsForInstanceInfoUpdaters = clientConfig.getUsesExplicitEventsForInitialDiscoveryCalls();
+        if (!eventsForInstanceInfoUpdaters) {
+            initInstanceHeartBeatUpdater(eventsForInstanceInfoUpdaters);
+            initInstanceInfoRefreshUpdater(eventsForInstanceInfoUpdaters);
+        }
+
         try {
             Monitors.registerObject(this);
         } catch (Throwable e) {
@@ -594,17 +601,11 @@ public class DiscoveryClient implements EurekaClient {
     /**
      * Checks to see if the eureka client registration is enabled.
      *
-     * @param myInfo
-     *            - The instance info object
      * @return - true, if the instance should be registered with eureka, false
      *         otherwise
      */
-    private boolean shouldRegister(InstanceInfo myInfo) {
-        if (!clientConfig.shouldRegisterWithEureka()) {
-            return false;
-        }
-
-        return true;
+    private boolean shouldRegister() {
+        return clientConfig.shouldRegisterWithEureka();
     }
 
     /**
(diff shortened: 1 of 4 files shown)


NiteshKant:
This is interesting! 
Would there be a reason why someone will need to register and not send heartbeats or not update instance info periodically?
IOW, should the change more be for controlling registrations (and possibly registry fetch) explicitly instead of getting registered in the constructor but deferring heartbeats & instance info refresh?

spencergibb:
I would be interested in this.  There are cases where all of the information to register a service isn't available when the object is constructed.  We go through some gyrations in [spring-cloud-netflix](https://github.com/spring-cloud/spring-cloud-netflix) to approximate this.

aspyker:
@NiteshKant and @spencergibb A change to externalize the scheduling of any of the events (registry refresh, heartbeat, and instance info dirty check/update) isn't the intent of this patch.  Also, it isn't the intent to give any more configurability than already existed than the initial timing of heartbeat and instance info direct check/update.  This change was mainly to allow for us to register the instance in discovery as fast as possible on an external event (vs. some timing that doesn't align with the app container startup).

That said, @spencergibb could use this fix to be slower and when all data was available (as using the property will cause the initial registration to wait until the startInstanceInfoHeartBeatNow call).

qiangdavidliu:
@aspyker thanks for the pull, this would definitely help with some of the slow eureka startup issues we have. One point of interest is whether we should separate registration/update and heartbeat scheduler initializations from each other, as there are some inherent assumptions in eureka of a client's behaviour on the registration path concerning registrations/updates/heartbeats. From the changes it looks like the use cases would be either to initialize both tasks with the existing delays, or to initialize both tasks with 0 initial delay. If this is the case, we should be able to combine the inits for both heartbeat and update, so at least that will guarantee a super use won't init one and forget to init the other.

NiteshKant:
>>  A change to externalize the scheduling of any of the events (registry refresh, heartbeat, and instance info dirty check/update) isn't the intent of this patch.

My point here is, if we do externalize registration out of the constructor, we do not need these additional knobs and that also provides a intuitive API. This is the model eureka 2.x follows i.e. registration is not baked into the constructor and is driven by the users. 

>> One point of interest is whether we should separate registration/update and heartbeat scheduler initializations from each other

IMO, we shouldn't separate them (what I was saying in my earlier comment) as heartbeat & instance info updates are required for registration and without them, registrations do not work as expected.

eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
```diff
@@ -1792,6 +1878,24 @@ public HealthCheckHandler getHealthCheckHandler() {
         return healthCheckHandler;
     }
 
+    @Override
+    public void startInstanceInfoHeartBeatNow() {
+        boolean eventsForInstanceInfoUpdaters = clientConfig.getUseExplicitEventsForInitialDiscoveryCalls();
```
tbak: We should protect this method with a guard to prevent executing it twice.

aspyker:
Going to close due to #530 being a better solution

PR #527, opened 2015-05-28
Cross account describe ASGs

amanya:
The motivation of this patch is to make Eureka capable of knowing the state of auto scaling groups even if they are from microservices running in accounts other than the one where Eureka is running.

In that case, it will do a a cross-account assume role to get permission to describe the ASGs.

[Here](http://docs.aws.amazon.com/IAM/latest/UserGuide/roles-walkthrough-crossacct.html ) is a walkthrough from Amazon that explains how to setup the cross-account IAM role delegation: 

The role must be called `ListAutoScalingGroups` and the policy must be like this:
```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Action": [
                "autoscaling:DescribeAutoScalingGroups"
            ],
            "Effect": "Allow",
            "Resource": "*"
        }
    ]
} 
```

### DIFF
--- a/eureka-core/src/main/java/com/netflix/eureka/util/AwsAsgUtil.java
+++ b/eureka-core/src/main/java/com/netflix/eureka/util/AwsAsgUtil.java
@@ -25,8 +25,20 @@ import java.util.TimerTask;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 
+import com.amazonaws.auth.AWSCredentials;
+import com.amazonaws.services.securitytoken.AWSSecurityTokenService;
+import com.amazonaws.services.securitytoken.AWSSecurityTokenServiceClient;
+import com.amazonaws.services.securitytoken.model.AssumeRoleResult;
+import com.google.common.base.Strings;
+import com.netflix.appinfo.AmazonInfo;
+import com.netflix.appinfo.AmazonInfo.MetaDataKey;
+import com.netflix.appinfo.ApplicationInfoManager;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import com.amazonaws.ClientConfiguration;
 import com.amazonaws.auth.BasicAWSCredentials;
+import com.amazonaws.auth.BasicSessionCredentials;
 import com.amazonaws.auth.InstanceProfileCredentialsProvider;
 import com.amazonaws.services.autoscaling.AmazonAutoScaling;
 import com.amazonaws.services.autoscaling.AmazonAutoScalingClient;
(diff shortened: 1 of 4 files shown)


eureka-core/src/main/java/com/netflix/eureka/util/AwsAsgUtil.java
```diff
@@ -142,7 +152,13 @@ public void setStatus(String asgName, boolean enabled) {
      * @return - true if the ASG is disabled, false otherwise
      */
     private boolean isAddToLoadBalancerSuspended(String asgName) {
-        AutoScalingGroup asg = retrieveAutoScalingGroup(asgName);
+        String asgAccount = getASGAccount(asgName);
+        AutoScalingGroup asg;
+        if(asgAccount.equals(accountId)) {
```
qiangdavidliu: The accountId metadata field was added to eureka-client relatively recently, so for backwards compatibility, it is possible in existing deployments to have older clients that do not register this field. This will cause an NPE at this point in code. Can you please add a check for null asgAccount, and in the null case treat this as same account as the server?

eureka-core/src/main/java/com/netflix/eureka/util/AwsAsgUtil.java
```diff
@@ -313,6 +385,29 @@ public void run() {
...
+     * @return the account id
+     */
+    private String getASGAccount(String asgName) {
+        Applications apps = PeerAwareInstanceRegistry.getInstance().getApplications(false);
+
+        for (Application app : apps.getRegisteredApplications()) {
+            for (InstanceInfo instanceInfo : app.getInstances()) {
+                String thisAsgName = instanceInfo.getASGName();
+                if (thisAsgName != null && thisAsgName.equals(asgName)) {
+                    return ((AmazonInfo) instanceInfo.getDataCenterInfo()).get(MetaDataKey.accountId);
```
qiangdavidliu: The accountId metadata field was added to eureka-client relatively recently, so for backwards compatibility, it is possible in existing deployments to have older clients that do not register this field. This will return null here and cause an NPE upstream. Can you please add a null check here and let it fall through instead?

eureka-core/src/main/java/com/netflix/eureka/util/AwsAsgUtil.java
```diff
@@ -195,6 +211,62 @@ private AutoScalingGroup retrieveAutoScalingGroup(String asgName) {
...
 
+    private Credentials initializeStsSession(String asgAccount) {
+        AWSSecurityTokenService sts = new AWSSecurityTokenServiceClient(new InstanceProfileCredentialsProvider());
+        String region = DiscoveryManager.getInstance().getEurekaClientConfig()
+                .getRegion();
+        if (!region.equals("us-east-1")) {
+            sts.setEndpoint("sts." + region + ".amazonaws.com");
+        }
+
+        String roleArn = "arn:aws:iam::" + asgAccount + ":role/ListAutoScalingGroups";
```
qiangdavidliu: Can you make the role name (ListAutoScalingGroups) a configurable parameter? You can add it as a config param to EurekaServerConfig/DefaultEurekaServerConfig with this as the default.

qiangdavidliu:
@amanya thank you for the pull request, this would be a very useful feature for Eureka to have. I added a couple of minor comments that I hope you can take a quick look at. 

Additionally, have you been able to verify this code change? We have mockito as a dependency and it should be able to write a quick unit test for this mocking out the AWS SDK with mockito.

amanya:
@qiangdavidliu thanks for your comments :)

I think I've managed to solve the first two issues, but for making the tests, I'll need to use a tool like Powermockito due to the static dependencies of the AwsAsgUtil class, right?

qiangdavidliu:
Thanks @amanya . We can merge this in and I'll take a look at what kind of testing we can do for this.

PR #529, opened 2015-05-31
1.x batched replication refactoring

tbak:


eureka-core/src/main/java/com/netflix/eureka/AbstractInstanceRegistry.java
```diff
@@ -0,0 +1,1282 @@
+/*
```
qiangdavidliu: How much change is this from the original InstanceRegistry?

eureka-core/src/main/java/com/netflix/eureka/PeerAwareInstanceRegistryImpl.java
```diff
@@ -0,0 +1,749 @@
+/*
```
qiangdavidliu: Was there any logic changes other than the interface refactoring?

eureka-core/src/main/java/com/netflix/eureka/AbstractInstanceRegistry.java
tbak: Almost no change, except extraction of interface. Unfortunately git treats this as a new file.

eureka-core/src/main/java/com/netflix/eureka/PeerAwareInstanceRegistryImpl.java
tbak: All logic related to peer nodes management is taken out to PeerNodes class. No other changes.

eureka-client/src/main/java/com/netflix/discovery/shared/EurekaHttpClient.java
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * @author Tomasz Bak
+ */
+public interface EurekaHttpClient {
+
+    HttpResponse<Void> register(InstanceInfo info);
+
+    HttpResponse<Void> cancel(String appName, String id);
+
+    HttpResponse<InstanceInfo> sendHeartBeat(String appName, String id, InstanceInfo info, InstanceStatus overriddenStatus);
```
qiangdavidliu: the InstanceInfo should contain the appName and id, so no need to pass these as params.

eureka-client/src/main/java/com/netflix/discovery/shared/JerseyEurekaHttpClient.java
```diff
@@ -0,0 +1,153 @@
...
+            WebResource webResource = getJerseyApacheClient().resource(serviceUrl)
+                    .path(urlPath)
+                    .queryParam("status", info.getStatus().toString())
+                    .queryParam("lastDirtyTimestamp", info.getLastDirtyTimestamp().toString());
+            if (overriddenStatus != null) {
+                webResource = webResource.queryParam("overriddenstatus", overriddenStatus.name());
+            }
+            addExtraHeaders(webResource);
+            response = webResource.accept(MediaType.APPLICATION_JSON_TYPE).put(ClientResponse.class);
+            InstanceInfo infoFromPeer = null;
```
qiangdavidliu: Can we move these peer specific code to a server implementation of this class? If we can break these out, it would be nice to refactor DiscoveryClient with this class.

eureka-client/src/main/java/com/netflix/discovery/shared/EurekaHttpClient.java
tbak: This API maps one to one to existing client API, but you are right, we can
do simplification here.

eureka-client/src/main/java/com/netflix/discovery/shared/JerseyEurekaHttpClient.java
tbak: Good point. We need to override this implementation in replicating client.

tbak:
Due to merge issues I need to close this PR and make a new one.

PR #530, opened 2015-06-01
Adding the ability to execute on-demand updates to the eureka server

qiangdavidliu:
Adding the ability to execute on-demand updates to the eureka server. Current set up is triggered by changes to ApplicationInfoManager.setInstanceStatus(). This capability can be turned on/off via eureka.shouldOnDemandUpdateStatusChange.


eureka-client/src/main/java/com/netflix/discovery/InstanceInfoReplicator.java
```diff
@@ -0,0 +1,114 @@
...
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
+
+/**
+ * A task for updating and replicating the local instanceinfo to the remote server. Properties of this task are:
+ * - configured with a single update thread to guarantee sequential update to the remote server
```
tbak: As this thread executes remote call, it is vulnerable to halts on socket IO operation.

qiangdavidliu: Good point. It should not be fatal in this case, as we still have the (guarded) heartbeat thread that will enforce re-registration for dirty instanceInfo (server side we check the lastDirtyTimestamp) if this thread is hanging on the network. This just means updates will be periodic as opposed to on-demand during the hangs until the hangs end.

eureka-client/src/test/java/com/netflix/discovery/DiscoveryClientRegisterUpdateTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+        applicationInfoManager = ApplicationInfoManager.getInstance();
+        applicationInfoManager.initComponent(new MyDataCenterInstanceConfig());
+        EurekaClient client = new DiscoveryClient(applicationInfoManager.getInfo(), new DefaultEurekaClientConfig());
+    }
+
+    @Test
+    public void registerUpdateLifecycleTest() throws Exception {
+        applicationInfoManager.setInstanceStatus(InstanceInfo.InstanceStatus.UP);
+        Thread.sleep(1000);  // give some execution time
```
tbak: Can these all sleeps be replaced with active pooling? This would speed up the test a lot. I have applied this strategy in replication tests, and they execute pretty fast.

eureka-examples/build.gradle
```diff
@@ -3,6 +3,8 @@ apply plugin: 'application'
 dependencies {
     compile project(':eureka-client')
     compile files('conf')
+
+    compile 'org.slf4j:slf4j-simple:1.7.7'
```
tbak: Lets keep versions in root build.gradle for easy upgrade.

eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
```diff
@@ -723,6 +777,7 @@ void register() {
     @PreDestroy
     @Override
     public void shutdown() {
+        ApplicationInfoManager.getInstance().unregisterStatusChangeListener(statusChangeListener.getId());
```
tbak: Is it possible to pass ApplicationInfoManager instance in constructor via injection to eliminate this singleton access?

eureka-client/src/test/java/com/netflix/discovery/DiscoveryClientRegisterUpdateTest.java
qiangdavidliu: The 1s sleeps here are explicit to allow the operations to go through without getting rateLimited (contrast with the @ Test below). May be able to shorten the longer sleeps though by making the replication interval even tighter than 10s.

eureka-examples/build.gradle
qiangdavidliu: This is just for the example project so that we can actually see logs (cient dependency is on slf4j-api);

eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
qiangdavidliu: Yes it is:). But cleaning up the statics (there are many others) should be a separate task so as not to conflate the pull requests.

eureka-client/src/main/java/com/netflix/appinfo/ApplicationInfoManager.java
```diff
@@ -103,13 +109,26 @@ public void registerAppMetadata(Map<String, String> appMetadata) {
...
-     *            Status of the instance
+     * @param status Status of the instance
      */
-    public void setInstanceStatus(InstanceStatus status) {
-        instanceInfo.setStatus(status);
+    public synchronized void setInstanceStatus(InstanceStatus status) {
+        InstanceStatus prev = instanceInfo.setStatus(status);
+        if (prev != null) {
+            for (StatusChangeListener listener : listeners.values()) {
+                listener.notify(new StatusChangeEvent(prev, status));
```
NiteshKant: Handle errors on listener invocation?

eureka-client/src/main/java/com/netflix/discovery/InstanceInfoReplicator.java
```diff
@@ -0,0 +1,114 @@
...
+
+            Long dirtyTimestamp = instanceInfo.isDirtyWithTime();
+            if (dirtyTimestamp != null) {
+                discoveryClient.register();
+                instanceInfo.unsetIsDirty(dirtyTimestamp);
+            }
+        } catch (Throwable t) {
+            logger.warn("There was a problem with the instance info replicator", t);
+        } finally {
+            Future next = scheduler.schedule(this, replicationIntervalSeconds, TimeUnit.SECONDS);
```
NiteshKant: Are you trying to achieve scheduling with fixed delay with this schedule-at-the-end-of-a-run construct?

eureka-client/src/main/java/com/netflix/discovery/InstanceInfoReplicator.java
```diff
@@ -0,0 +1,114 @@
...
+    public boolean onDemandUpdate() {
+        if (rateLimiter.acquire(burstSize, allowedRatePerMinute)) {
+            scheduler.submit(new Runnable() {
+                @Override
+                public void run() {
+                    logger.debug("Executing on-demand update of local InstanceInfo");
+
+                    // cancel the latest scheduled update, it will be rescheduled at the end of run()
+                    Future latestPeriodic = scheduledPeriodicRef.get();
+                    if (latestPeriodic != null && !latestPeriodic.isDone()) {
```
NiteshKant: May be having the last onDemand run time stored and shared between onDemand & scheduled tasks, can remove this schedule-cancel-schedule song & dance :)
In that case, the scheduled task can just complete, if the onDemand update has run within X amount of time.

aspyker:
On shutdown of tomcat, I am consistently getting:
SEVERE: The web application [] appears to have started a thread named [DiscoveryClient-InstanceInfoReplicator-0] but has failed to stop it. This is very likely to create a memory leak.

aspyker:
I tested this code and it does improve the startup time.  I am seeing ~30 seconds (instead of the previous 58).

qiangdavidliu:
@aspyker changed the close() on the replicator to call executors.shutdownNow() instead of shutdown().

Issue #533, opened 2015-06-05
BelowRenewThreshold: isBelowRenewThreshold() & isLeaseExpirationEnabled() do not agree


brenuart:
Eureka activate its self protection mode when last minute's renewal rate is below the threshold.
The following two methods of `com.netflix.eureka.PeerAwareInstanceRegistry` do not agree on the condition:

     public int isBelowRenewThreshold() {
        if ((getNumOfRenewsInLastMin() < numberOfRenewsPerMinThreshold)
    ...

For that method, self protection should be activated when `renewsRate < threshold`. However, for the following method, the mode is activated if `renewsRate <= threshold`:

    public boolean isLeaseExpirationEnabled() {
        boolean leaseExpirationEnabled = (numberOfRenewsPerMinThreshold > 0)
                && (getNumOfRenewsInLastMin() > numberOfRenewsPerMinThreshold);
    ...

So we get unexpected behaviour when `renewsRate == threshold`.

As far as I can tell, `isbelowRenewThreshold()` is only used by Servo to report metrics, while `isLeaseExpirationEnabled()` is used by the eviction thread to perform the actual eviction of expired leases.

qiangdavidliu:
fixed in #537

Issue #531, opened 2015-06-03
Lease renewal: inconsistency between Lease.renew() and Lease.isExpired()


brenuart:
There seems to be an inconsistency between the two following methods (from `com.netflix.eureka.lease.Lease`:

    public void renew() {
        lastUpdateTimestamp = System.currentTimeMillis() + duration;
    }

    public boolean isExpired() {
        return (evictionTimestamp > 0 || System.currentTimeMillis() > (lastUpdateTimestamp + duration));
    }

When the lease is renewed, the `lastUpdatedTimestamp` is set to `now()` + the expected client lease renewal period. It therefore represents _when the lease is about to expire_ and not when it was last renewed.

To check if the lease is expired, the `isExpired()` method adds again the duration to the `lastUpdateTimestamp` - hence the lease will expire only after two times the duration.

Finally, javadocs on the `getLastRenewalTimestamp()` clearly defines the actual meaning of the `lastUpdateTimestamp`:

    /**
     * Gets the milliseconds since epoch when the lease was last renewed.
     *
     * @return the milliseconds since epoch when the lease was last renewed.
     */
    public long getLastRenewalTimestamp() {
        return lastUpdateTimestamp;
    }

Conclusion: the `renew()` method seems to be broken: *it shouldn't add the duration to the current time.*

qiangdavidliu:
@brenuart thanks for point this out, this indeed looks to be the case. We'll take a look at the impact of this and get back to you on whether we fix or update the comments (as this code is very old and been working in our production for a while, this unintended behaviour may be something that clients now expect, and is therefore potentially harmful to change). Thanks again.

brenuart:
The self protection mode of our Eureka got activated for some reason so lease expiration was temporarily suspended. We then had a look at the registry lease data searching for instances that expired... and discovered that some had their `lastRenewalTimestamp` in the future... This is how we discovered the issue.

qiangdavidliu:
Comments updated in #537

PR #537, opened 2015-06-08
Various fixes (specifics in commit log)

qiangdavidliu:
Also fixes for issues #531 and #533

### DIFF
--- a/eureka-core/src/main/java/com/netflix/eureka/PeerAwareInstanceRegistry.java
+++ b/eureka-core/src/main/java/com/netflix/eureka/PeerAwareInstanceRegistry.java
@@ -674,14 +674,14 @@ public class PeerAwareInstanceRegistry extends InstanceRegistry {
     }
 
     /**
-     * Checks if the number of renewals is lesser than threshold.
+     * Checks if the number of renewals is lesser than threshold. This method is for servo metrics only.
      *
      * @return 0 if the renewals are greater than threshold, 1 otherwise.
      */
     @com.netflix.servo.annotations.Monitor(name = "isBelowRenewThreshold", description = "0 = false, 1 = true",
             type = com.netflix.servo.annotations.DataSourceType.GAUGE)
     public int isBelowRenewThresold() {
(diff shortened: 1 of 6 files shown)


eureka-core/src/main/java/com/netflix/eureka/PeerAwareInstanceRegistry.java
```diff
@@ -674,14 +674,14 @@ public long getNumOfReplicationsInLastMin() {
...
-     * Checks if the number of renewals is lesser than threshold.
+     * Checks if the number of renewals is lesser than threshold. This method is for servo metrics only.
      *
      * @return 0 if the renewals are greater than threshold, 1 otherwise.
      */
     @com.netflix.servo.annotations.Monitor(name = "isBelowRenewThreshold", description = "0 = false, 1 = true",
             type = com.netflix.servo.annotations.DataSourceType.GAUGE)
     public int isBelowRenewThresold() {
-        if ((getNumOfRenewsInLastMin() < numberOfRenewsPerMinThreshold)
+        if ((getNumOfRenewsInLastMin() <= numberOfRenewsPerMinThreshold)
```
tbak: Why was it important to change this condition?

tbak:
Looks good.

eureka-core/src/main/java/com/netflix/eureka/PeerAwareInstanceRegistry.java
qiangdavidliu: see issue #533. This is to align servo reporting logic with actual selfPreservation logic.

brenuart: Would be even better if `isLeaseExpirationEnabled()` delegates to `isBelowRenewThreshold()` instead of copy/paste the code...

qiangdavidliu: @brenuart there are subtle differences in the two, so it would not be a trivial replacement.

PR #540, opened 2015-06-08
2.x Integrate server RegistrationChannel with new registry API.

tbak:


eureka2-write-server/src/main/java/com/netflix/eureka2/server/channel/RegistrationChannelImpl.java
```diff
@@ -99,10 +100,10 @@ public void onNext(Void aVoid) {
             }
 
             private void evictIfPresent() {
-                InstanceInfo toEvict = instanceInfoRef.get();
+                InstanceInfo toEvict = registrationSubject.getValue();
```
qiangdavidliu: Can we avoid using this if possible? This is still @Experimental in rxJava.

eureka2-write-server/src/main/java/com/netflix/eureka2/server/channel/RegistrationChannelImpl.java
```diff
@@ -124,55 +125,81 @@ public Source getSource() {
...
+        STATE currentState = getState();
+        if (STATE.Closed == currentState) {
+            // Since channel is already closed and hence the transport, we don't need to send an error on transport.
+            return Observable.error(CHANNEL_CLOSED_EXCEPTION);
+        } else {
+            Exception exception = new IllegalStateException("Unknown state error when registering, current state: " + currentState);
+            return sendErrorOnTransport(exception).doOnTerminate(new Action0() {
+                @Override
+                public void call() {
+                    close();
```
qiangdavidliu: close(exception)?

eureka2-write-server/src/test/java/com/netflix/eureka2/server/channel/RegistrationChannelImplTest.java
```diff
@@ -0,0 +1,157 @@
...
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+/**
+ * server side registration channel test
+ *
+ * @author David Liu
+ */
+public class RegistrationChannelImplTest {
```
qiangdavidliu: This new test doesn't seem to cover all the cases that the old test covers?

eureka2-write-server/src/main/java/com/netflix/eureka2/server/transport/tcp/registration/TcpRegistrationHandler.java
```diff
@@ -32,25 +33,23 @@
...
     private final WriteServerConfig config;
-    private final SourcedEurekaRegistry<InstanceInfo> registry;
+    private final EurekaRegistrationProcessor<InstanceInfo> registrationProcessor;
     private final EvictionQueue evictionQueue;
     private final WriteServerMetricFactory metricFactory;
 
     @Inject
     public TcpRegistrationHandler(WriteServerConfig config,
-                                  SourcedEurekaRegistry registry,
+                                  @Named("registration") EurekaRegistrationProcessor registrationProcessor,
```
qiangdavidliu: Can we use the fields in the com.netflix.eureka2.Names class here instead of strings? (Also the same in the Module binding).

eureka2-write-server/src/main/java/com/netflix/eureka2/server/registry/EvictionQuotaProvider.java
```diff
@@ -1,10 +1,10 @@
...
 
 import rx.Observable;
 
 /**
  * Back pressure aware observable that emits eviction quotas, when requested
- * by {@link com.netflix.eureka2.registry.PreservableRegistryProcessor}.
+ * by {@link PreservableRegistryProcessor}.
  *
  * @author Tomasz Bak
  */
```
qiangdavidliu: Can we use a different name than Provider here? It conflates with Guice Provider<>.

eureka2-write-server/src/main/java/com/netflix/eureka2/server/registry/EvictionQuotaProviderImpl.java
```diff
@@ -54,9 +56,10 @@
...
 
-    public EvictionQuotaProviderImpl(final SourcedEurekaRegistry registry, int allowedPercentageDrop) {
+    @Inject
+    public EvictionQuotaProviderImpl(final SourcedEurekaRegistry registry, WriteServerConfig config) {
         this.registry = registry;
-        this.allowedPercentageDrop = allowedPercentageDrop;
+        this.allowedPercentageDrop = config.getEvictionAllowedPercentageDrop();
 
         /**
          * Registry size change should trigger state evaluation, as it may generate eviction permits.
```
qiangdavidliu: on line 78 distinctUntilChanged instead of distinct?

eureka2-write-server/src/main/java/com/netflix/eureka2/server/channel/RegistrationChannelImpl.java
```diff
@@ -187,54 +214,32 @@ public void call(Throwable throwable) {
...
-                                instanceInfoRef.set(null);
-                                sendAckOnTransport().doOnTerminate(new Action0() {
-                                    @Override
-                                    public void call() {
-                                        close();
-                                    }
-                                }).subscribe();
-                            }
-                        });
+                InstanceInfo instanceInfo = registrationSubject.getValue();
```
qiangdavidliu: @ Experimental method, can we avoid?

eureka2-write-server/src/main/java/com/netflix/eureka2/server/channel/RegistrationChannelImpl.java
```diff
@@ -124,55 +125,81 @@ public Source getSource() {
...
-                Exception exception = new IllegalStateException("Unknown state error when registering, current state: " + currentState);
-                return sendErrorOnTransport(exception).doOnTerminate(new Action0() {
-                    @Override
-                    public void call() {
-                        close();
-                    }
-                });
-            }
+        if (moveToState(STATE.Idle, STATE.Registered)) {
+            return firstRegistration(instanceInfo);
```
qiangdavidliu: The returned result here is the chained result of registrationProcessor.register, whereas for update and unregister we have separate observables. What instead of setting up the initial registration processor in the first register, we do the set up independently and then can just use the same code as in registrationUpdate() for both the initial register and subsequent updates?

eureka2-write-server/src/main/java/com/netflix/eureka2/server/channel/RegistrationChannelImpl.java
tbak: I have missed the `@Experimental` annotation. Reading last value from subject let eliminate InstanceInfo field in the channel. I will have to put it back.

eureka2-write-server/src/main/java/com/netflix/eureka2/server/channel/RegistrationChannelImpl.java
tbak: Changed.

eureka2-write-server/src/test/java/com/netflix/eureka2/server/channel/RegistrationChannelImplTest.java
tbak: ```RegistrationChannelImpl``` implementation is much simpler now, and we do not need this amount of tests.

eureka2-write-server/src/main/java/com/netflix/eureka2/server/transport/tcp/registration/TcpRegistrationHandler.java
tbak: Fixed here, and in a few other places.

eureka2-write-server/src/main/java/com/netflix/eureka2/server/registry/EvictionQuotaProvider.java
tbak: Renamed to ```EvictionQuotaKeeper```

eureka2-write-server/src/main/java/com/netflix/eureka2/server/registry/EvictionQuotaProviderImpl.java
tbak: Absolutely. Thanks for spotting this.

eureka2-write-server/src/main/java/com/netflix/eureka2/server/channel/RegistrationChannelImpl.java
tbak: Done

eureka2-write-server/src/main/java/com/netflix/eureka2/server/channel/RegistrationChannelImpl.java
tbak: We could do eager registration with initially empty subject, if id was not needed. With current API we can do this only on the first register, hence we have different code path. The return observable for update is a artificial with current API, as we do not provide per update state from the registry. I also do not think we should do that, as it would add a lot of extra cost with no benefit.

Issue #536, opened 2015-06-08
Improvement: optional read only ResponseCache


brenuart:
It would be nice if the use of the read only response cache could be made optional. Although it seems be useful in some scenarios (see https://github.com/Netflix/eureka/issues/147), it also imposes an additional delay before changes are reflected in the Eureka REST API. This extra delay is very annoying in some deployments.

Is this something that could be foreseen?

qiangdavidliu:
@brenuart this is a feature that we needed internally to help eureka scale to our production usage. It would be useful to have this controllable via a server configuration however. Would you like to make a pull request for this?

brenuart:
I will submit a PR for this.

To be honest, I don't understand why this extra RO cache helps to make performance better. There is already a first cache... why isn't it sufficient? 
I understand the RO cache has the same effect of delaying the invalidations - just like if you applied them in batch of say 100. Is that the objective?

qiangdavidliu:
Hi @brenuart thanks for the pull request. The readonly cache exist as it's a non-expiry based cache compared to the readWriteCache. What we had observed previously was that the increased numbers of 503s (caused due to cache misses on gets as items are expiring) was causing more issues for us due to the scale of usage.

qiangdavidliu:
#544

PR #541, opened 2015-06-09
Disable the read only cache if 'responseCacheUpdateIntervalMs' is set to 0

brenuart:
'responseCacheUpdateIntervalMs' controls how often the read only cache is updated with changes collected in the read-write cache. The lesser the value, the faster the cache is updated. So a value of 0 can be interpreted as an immediate update. 
This seems a good and easy way to disable the read only cache.

See https://github.com/Netflix/eureka/issues/536

### DIFF
--- a/eureka-core/src/main/java/com/netflix/eureka/EurekaServerConfig.java
+++ b/eureka-core/src/main/java/com/netflix/eureka/EurekaServerConfig.java
@@ -259,7 +259,7 @@ public interface EurekaServerConfig {
 
     /**
      * Gets the time interval with which the payload cache of the client should
-     * be updated.
+     * be updated. Set to 0 to update the cache immediately.
      *
      * @return time in milliseconds.
      */
--- a/eureka-core/src/main/java/com/netflix/eureka/resources/ResponseCache.java
+++ b/eureka-core/src/main/java/com/netflix/eureka/resources/ResponseCache.java
@@ -154,12 +154,23 @@ public class ResponseCache {
 
     private static final ResponseCache s_instance = new ResponseCache();
 
+    /**
+     * Indicates if the read only cache is enabled or not.
+     */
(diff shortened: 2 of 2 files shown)


qiangdavidliu:
Thanks for this @brenuart . Rather than conflating the same property for two different use cases (setting the refresh rate as well as enabling/disabling the readonly cache), what about defining a new configuration for disabling/enabling the cache? There is a top level get method (get(final Key key)) that then calls the internal get methods with ignoreReadonly as a param, but right now it always calls with false. This seems a nice hook point for bypassing the readonly cache.

brenuart:
The idea was that if I want to speed-up things, I had to use a lower value for the `responseCacheUpdateIntervalMs`, say I would have set it to `100ms`. If not fast enough, I would have use `50ms`, then `1ms`... So at the end, `0` means _immediate_, and I get the desired behaviour.
This had the added benefit of not introducing a new configuration property in the game while still keeping an easy to understand meaning for the existing one.

Now, if you prefer to add a new configuration property, I can update the PR with that. 
What property name would you suggest?

qiangdavidliu:
Thanks again for the pull @brenuart . The readonly cache and the readWrite caches are actually serving two different use cases that are not quite symmetric. The readonly cache is not a guava cache and does not actually have any expiry or clean up policy, so once it caches data the data stays in the cache without expiry (the update task helps with purging removed apps and leaves tombstones). The issue with using the readWrite cache directly, is that the auto expiry for that causes 503s for gets when a cache miss occur.

For convenience I created pull #544 for this, does this pull satisfy your needs?

brenuart:
This is ok for me. The new config property allows me to disable the read only cache and have changes propagated immediately. Thanks.

qiangdavidliu:
Thanks @brenuart . Closing.

PR #543, opened 2015-06-09
AwsAsgUtil fixes

qiangdavidliu:
- async load data from the asgCache for isASGEnabled queries.
- avoid using expensive getASGAccount(asgName) if possible.


tbak:
Looks good

PR #548, opened 2015-06-15
2.x Adding interest channel eviction mechanism + tests.

qiangdavidliu:


eureka2-client/src/main/java/com/netflix/eureka2/client/channel/InterestChannelInvoker.java
```diff
@@ -28,6 +32,11 @@ public InterestChannelInvoker(InterestChannel delegate, EurekaClientMetricFactor
     public InterestChannelInvoker(InterestChannel delegate, EurekaClientMetricFactory metricFactory, Scheduler scheduler) {
         super(metricFactory.getSerializedTaskInvokerMetrics(InterestChannelInvoker.class), scheduler);
         this.delegate = delegate;
+        if (delegate instanceof Sourced) {
```
tbak: Shouldn't we have always single case here with sourced delegate?

eureka2-client/src/test/java/com/netflix/eureka2/client/interest/AbstractInterestClientTest.java
```diff
@@ -0,0 +1,115 @@
...
+
+    private SourcedEurekaRegistryImpl registry;
+    private RetryableConnection<InterestChannel> retryableConnection;
+    private ReplaySubject<InterestChannel> channelSubject;
+
+    private MyInterestClient client;
+
+    @Before
+    public void setUp() {
+        registry = mock(SourcedEurekaRegistryImpl.class);
```
tbak: Why not mock SourcedEurekaRegistry which is an interface?

eureka2-client/src/test/java/com/netflix/eureka2/client/interest/AbstractInterestClientTest.java
```diff
@@ -0,0 +1,115 @@
...
+
+        channelSubject.onNext(channel2);
+        channel1Lifecycle.onError(new Exception("test error"));
+
+        channel2BatchSubject.onNext(newBufferStart());
+        verify(registry, never()).evictAll(any(Source.SourceMatcher.class));
+
+        channel2BatchSubject.onNext(newBufferEnd());
+        verify(registry, times(1)).evictAll(any(Source.SourceMatcher.class));
+        Thread.sleep(1000);
```
tbak: Why this sleep here?

eureka2-client/src/test/java/com/netflix/eureka2/client/interest/AbstractInterestClientTest.java
```diff
@@ -0,0 +1,115 @@
...
+            return retryableConnection;
+        }
+
+        @Override
+        public Observable<ChangeNotification<InstanceInfo>> forInterest(Interest<InstanceInfo> interest) {
+            return null;
+        }
+
+        @Override
+        public void registryEvictionSubscribe(RetryableConnection<InterestChannel> retryableConnection) {
```
tbak: Why override here if we only call super?

eureka2-server/src/main/java/com/netflix/eureka2/server/channel/InterestNotificationMultiplexer.java
```diff
@@ -112,6 +112,6 @@ public void unregister() {
      * simplifies implementation of the multiplexer.
      */
     public Observable<ChangeNotification<InstanceInfo>> changeNotifications() {
-        return aggregatedStream;
+        return aggregatedStream.share();
```
tbak: There is no need for that as channel contract guarantees single subscription only.

eureka2-client/src/main/java/com/netflix/eureka2/client/channel/InterestChannelInvoker.java
qiangdavidliu: In actual implementation yes, but this is not enforced by the current InterestChannel contract. Going forward I propose we have all the channel interfaces extend the Sourced interface.

eureka2-client/src/test/java/com/netflix/eureka2/client/interest/AbstractInterestClientTest.java
qiangdavidliu: Trigger happy autocomplete. Fixed.

eureka2-client/src/test/java/com/netflix/eureka2/client/interest/AbstractInterestClientTest.java
qiangdavidliu: Legacy code not removed. Gone.

eureka2-client/src/test/java/com/netflix/eureka2/client/interest/AbstractInterestClientTest.java
qiangdavidliu: In a previous iteration this was different. Removed.

eureka2-server/src/main/java/com/netflix/eureka2/server/channel/InterestNotificationMultiplexer.java
qiangdavidliu: As part of this pull InterestChannel have a new API getChangeNotificationStream(). On the server side this is bridged to the Multiplexer's changeNotifications so even though we don't subscribe to this stream more than once, this share() is to guard against that per the new contract.

Issue #545, opened 2015-06-11
Improvement: "CacheRefresh" event


brenuart:
We are looking for a way to be notified when the Eureka client has refresh its local cache. This could help to quickly react on changes in the list of discovered instances, like adapt balancing/routing behaviour.

The `DiscoveryClient` doesn't expose any public/protected method that could be easily overridden to implement such behaviour. However, we were wondering if the `DiscoveryClient` could be enhanced to publish a `CacheRefresh` event on the `eventBus` every time it refreshes its cache?

qiangdavidliu:
@brenuart this would be a useful feature. A simple hook point for this would be to publish this event after every successful execution of the CacheRefreshThread, however this may return some false positives as not every refresh will actually have updates to the local cache. To publish this event only for real changes  may involved some careful hooks to a few different methods (starting point would be the CacheRefreshThread).

Would you be interested in making a pull for this feature?

brenuart:
Some remarks:

(1) First question is what kind of event do we need ?
A `CacheRefreshedEvent` would simply tell the world the cache is refreshed and up-to-date. It doesn't say something has changed. We could leave the change detection to a separate component that would perform the diff and eventually send another event with the change information.

A `CacheUpdateEvent` on the other hand would be sent only if there is a change. But I wonder is if it is useful at all if it doesn't hold information about _what_ has actually changed:
- new instance discovered (which one)?
- instance gone (which one)?
- instance status changed
- etc

It start getting complicated. So my advice would be to go for the simplest `CacheRefreshedEvent` and possibly build a separate listener that would track changes. People interested in the actual delta could use it.


(2) Event or listener?
The second question is whether we should plug into the `eventBus` or go for a simple listener. In the later case, interested parties would register with the `DiscoveryClient` to get the info.
Drawback of the listener pattern: requires component to unregister


(3) Starting hook
According to me, it is probably better to hook into the `fetchRegistry()` method instead of the `CacheRefreshThread`. The `fetchRegistry()` method is where the actual work is done and can/could be invoked from many different places. Currently only at startup and then at regular intervals by the `CacheRefreshThread`. 

Your opinion?

brenuart:
@qiangdavidliu ping?

qiangdavidliu:
@brenuart sorry was busy with other work.

I believe you and I essentially have the same agreement in both of our comments, that is there is two ways to go about this.

[...]

PR #549, opened 2015-06-15
Capture "CacheRefreshed" event

brenuart:
The objective is to capture "cache refreshed" events. A few comments (following discussion in https://github.com/Netflix/eureka/issues/545):

1/ Added protected methods invoked when remote status is changed and cache is refreshed. They  are currently responsible to fire the appropriate event to the EventBus if one is available.
They are intentionally made protected so subclasses of the `DiscoveryClient` can override them if they need more advanced behaviour like:
- implement the Observer pattern and invoke registered listeners/observers
- compute actual changes when cache is refreshed
- etc

2/ I could create a `CacheRefreshedEvent` similar to the existing `StatusChangedEvent` and fire it every time the cache is refreshed. However, I'm not sure about the implications this additional event may have on components currently listening to the only StatusChangedEvent... What is your opinion?

3/ The `FETCH_REGISTRY_TIMER` includes the time taken by the following bloc of code:

     private boolean fetchRegistry(boolean forceFullRegistryFetch) {
         try {
            ....
            // Notify about cache refreshed before updating the instance remote status
            onCacheRefreshed();
            
            // Update remote status based on refreshed data held in the cache
            updateInstanceRemoteStatus();
       } catch(Throwable t) {
       ....
       } finally {
           tracer.stop();

The tracer includes the time it takes to notify listening parties. Should't we stop the timer before?

4/ I could have introduced the `Observer` pattern directly. But then I think it should have been done for both the `StatusChangedEvent` AND the newly introduced `CacheRefreshedEvent`...
They new protected methods are good starting point to build the Observer pattern if needed.

### DIFF
--- a/eureka-client/src/main/java/com/netflix/discovery/CacheRefreshedEvent.java
+++ b/eureka-client/src/main/java/com/netflix/discovery/CacheRefreshedEvent.java
@@ -1,0 +1,11 @@
+package com.netflix.discovery;
+
+/**
+ * This event is sent by {@link EurekaClient) whenever it has refreshed its local 
+ * local cache with information received from the Eureka server.
+ * 
+ * @author brenuart
+ */
+public class CacheRefreshedEvent extends DiscoveryEvent {
+
+}
--- a/eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
+++ b/eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
@@ -2010,11 +2010,7 @@ public class DiscoveryClient implements EurekaClient {
      * @param newStatus the new remote {@link InstanceStatus} 
      */
     protected void onRemoteStatusChanged(InstanceInfo.InstanceStatus oldStatus, InstanceInfo.InstanceStatus newStatus) {
-    	// Publish event if an EventBus is available
-        if (eventBus != null) {
-            StatusChangeEvent event = new StatusChangeEvent(oldStatus, newStatus);
-            eventBus.publish(event);
-        }
+    	fireEvent(new StatusChangeEvent(oldStatus, newStatus));
     }
     
     /**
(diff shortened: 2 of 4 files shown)


brenuart:
Don't see why the test failed - cannot reproduce the failure locally.
Just wondering if it is not a timing issue?

qiangdavidliu:
Thanks @brenuart . I restarted the test and it looks like it was a transient failure.

On your comments:

1/ Agree.

2/ Messages published on the event bus can be consumed with the following recipe:
```java
// Here I am using an anonymous inner class to reduce verbosity of the example, in reality, although this works, you may want to have a named class.
eventBus.registerSubscriber(new Object() {
    // Any method annotated with @Subscribe will be a consumer of the event class (or any subclass thereof), which is the first (and only) argument in the method.
    @Subscribe
    /* Any access modifier & any method name works. You do not have to implement any interface */
    public void consume(MyEvent event) {
        // Consumption of the event code.
    }
});
``` 
So the consumers can listen for specific event types without interfering with the consumers of different event types.

Another example to look at for event bus consumption is [EurekaUpStatusResolver](https://github.com/Netflix/eureka/blob/master/eureka-client/src/main/java/com/netflix/discovery/EurekaUpStatusResolver.java)

3/ You are correct on this. I guess right now since EventBus publish and consumption is async it doesn't make a big difference.

4/ Agree.

brenuart:
Additional modifications after your comments:

(point 2) 
A new `CacheRefreshedEvent` has been created and is automatically published by the `DiscoveryClient` if an `EventBus` is provided.
I also made the `StatusChangeEvent` and `CacheRefreshEvent` inherit from the same parent `DiscoveryEvent`. Having a common parent for all events sent by the same source is quite convenient.

(point 3)
I moved the code not directly related to the refresh of the local cache outside of the scope of the timer.

qiangdavidliu:
Thanks @brenuart

PR #552, opened 2015-06-17
2.x Adding buffer hint based eviction for replication

qiangdavidliu:
Also done:
- renaming occurrences of "discovery" to "interest" in schema
- removed RegisterCopy and UnregisterCopy from replication protocol. Using equiv from interest protocol instead
- replication protocol also include StreamStateUpdate now
- Clean up ReplicationChannel interface
- Clean up misleading usages of "discovery" to be replaced with "interest"
- Clean up naming confusion with ReplicationHandler (now ReplicationSender) and TcpReplicationHandler


eureka2-core/src/main/avro/replication-schema.avdl
```diff
@@ -21,20 +23,31 @@ protocol ReplicationProtocol {
         boolean sendSnapshot;
     }
 
-    record RegisterCopy {
+    @namespace("com.netflix.eureka2.protocol.interest")
+    record AddInstance {
```
tbak: As we now share many types across different protocols, it makes sense to put them all into common "model" schema and leave protocol schemas with protocol specific declarations only (like ReplicationMessages below). This way we can avoid duplicating definitions.

eureka2-write-server/src/main/java/com/netflix/eureka2/server/transport/tcp/replication/TcpReplicationHandler.java
```diff
@@ -40,35 +48,117 @@
...
+                            StreamStateNotification<InstanceInfo> n = (StreamStateNotification<InstanceInfo>) notification;
+                            if (n.getBufferState() == StreamStateNotification.BufferState.BufferEnd) {
+                                return true;
+                            }
+                        }
+                        return false;
+                    }
+                })
+                .take(1)
+                .map(new Func1<ChangeNotification<InstanceInfo>, ReceiverReplicationChannel>() {
```
tbak: Is this extra map step needed? Prev can be accessed directly in the flatMap below.

eureka2-write-server/src/main/java/com/netflix/eureka2/server/transport/tcp/replication/TcpReplicationHandler.java
```diff
@@ -40,35 +48,117 @@
...
                                  WriteServerMetricFactory metricFactory) {
         this.config = config;
-        this.SelfIdentityService = SelfIdentityService;
+        this.selfIdentityService = selfIdentityService;
         this.registry = registry;
-        this.evictionQueue = evictionQueue;
         this.metricFactory = metricFactory;
+        this.channelSubject = BehaviorSubject.create();
+
+        channelSubject
```
tbak: TCP replication handler is shared by all replication connections. We might get three new replication connections first, with random failures later on. How this scan process resolves this?

eureka2-write-server/src/test/java/com/netflix/eureka2/server/transport/tcp/replication/TcpReplicationHandlerTest.java
```diff
@@ -0,0 +1,104 @@
...
+        handler = new TcpReplicationHandler(
+                WriteServerConfig.writeBuilder().build(),
+                selfInfoResolver,
+                registry,
+                WriteServerMetricFactory.writeServerMetrics()
+        );
+    }
+
+    @Test
+    public void testDoHandleEvictOlderChannelData() throws Exception {
```
tbak: We need test here where multiple concurrent connections are established.

eureka2-core/src/main/avro/replication-schema.avdl
qiangdavidliu: Agree. Did some of the that but there are still more we can consolidate here.

eureka2-write-server/src/main/java/com/netflix/eureka2/server/transport/tcp/replication/TcpReplicationHandler.java
qiangdavidliu: It's not. In the interest channel version (from which this was copied) it was needed.

eureka2-write-server/src/main/java/com/netflix/eureka2/server/transport/tcp/replication/TcpReplicationHandler.java
qiangdavidliu: Good point. This is a bug. Fixing.

PR #555, opened 2015-06-17
Refactor so other DI Systems can extend eureka.

spencergibb:
Make EurekaJacksonCodec more inline with Converters with public static classes to extend.
Make EurekaJacksonCodec.INSTANCE mutable.
Add get/set to DiscoveryClientOptionalArgs.

I'd rather not have a static instance in EurekaJacksonCodec, suggestions welcome.

### DIFF
--- a/eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
+++ b/eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
@@ -99,6 +99,7 @@ import org.slf4j.LoggerFactory;
  * </p>
  *
  * @author Karthik Ranganathan, Greg Kim
+ * @author Spencer Gibb
  *
  */
 @FineGrainedLazySingleton
--- a/eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
+++ b/eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
@@ -197,29 +197,14 @@ public class DiscoveryClient implements EurekaClient {
         @Inject(optional = true)
         private Provider<HealthCheckHandler> healthCheckHandlerProvider;
 
-        public EventBus getEventBus() {
-            return eventBus;
-        }
+        public DiscoveryClientOptionalArgs() { }
 
-        public void setEventBus(EventBus eventBus) {
+        public DiscoveryClientOptionalArgs(EventBus eventBus, Provider<HealthCheckCallback> healthCheckCallbackProvider, Provider<HealthCheckHandler> healthCheckHandlerProvider) {
             this.eventBus = eventBus;
-        }
-
-        public Provider<HealthCheckCallback> getHealthCheckCallbackProvider() {
-            return healthCheckCallbackProvider;
-        }
-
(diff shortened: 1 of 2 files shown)


spencergibb:
Not sure of the build failure.  Didn't change any behavior.

eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
```diff
@@ -196,6 +196,30 @@
 
         @Inject(optional = true)
         private Provider<HealthCheckHandler> healthCheckHandlerProvider;
+
+        public EventBus getEventBus() {
```
tbak: I understand that these methods are an alternative to DiscoveryClientOptionalArgs. As we use constructor injection everywhere, why not add yet another constructor that accepts the additional, optional values directly without a wrapper?

eureka-client/src/main/java/com/netflix/discovery/converters/EurekaJacksonCodec.java
```diff
@@ -50,7 +50,7 @@
 
     private static final Logger logger = LoggerFactory.getLogger(EurekaJacksonCodec.class);
 
-    private static final Version VERSION = new Version(1, 1, 0, null);
+    protected static final Version VERSION = new Version(1, 1, 0, null);
```
tbak: Is this refactoring needed to be able to extend serializer to support other data center types?

eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
spencergibb: They aren't an "alternative to DiscoveryClientOptionalArgs" as they are methods on the class.  I'll add the new constructor instead.

eureka-client/src/main/java/com/netflix/discovery/converters/EurekaJacksonCodec.java
spencergibb: probably not.

eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
spencergibb: Hmm, adding a new constructor to `DiscoveryClient` would seem more invasive than needs be.  Let me update and see what you think.

spencergibb:
@tbak I make the `DiscoveryClientOptionalArgs` have a constructor.  It seems to be the simplest thing to do what I'd like to accomplish.

spencergibb:
@tbak ping, I've rebased against master.

qiangdavidliu:
Thanks @spencergibb . Changes looks clean.

spencergibb:
@qiangdavidliu or @tbak any idea of when you might have a release?

qiangdavidliu:
@spencergibb we are looking at doing an rc release today. Since many of the dependencies has been updated, we will make the full release once the rc from today is confirmed to be stable.

spencergibb:
@qiangdavidliu thanks for the update!  Our team really appreciates it.

PR #556, opened 2015-06-17
Fixing a flaky unit test

qiangdavidliu:


qiangdavidliu:
looks like the build system is having hiccups

PR #557, opened 2015-06-18
fixed package name of serializer

kukgini:
StatusInfo's serializer should be 'com.netflix.discovery...' instead of
'com.netflix.eureka...'


qiangdavidliu:
thanks @kukgini

PR #560, opened 2015-06-21
2.x Updating to Governator 1.7 and archaius2.

qiangdavidliu:
- dependency updates
 - server refactoring to be inline with governator 1.7


eureka2-integration/src/test/java/com/netflix/eureka2/integration/client/EurekaClientFailoverTest.java
```diff
@@ -87,29 +87,35 @@ public void run() {
...
 
-        assertThat(registrationSubscription.isUnsubscribed(), is(false));
-        assertThat(interestSubscriber.takeNext(60, TimeUnit.SECONDS), is(modifyChangeNotificationOf(INSTANCE_DOWN)));
+            // Update instance status, and verify that it was handled
+            registrationSubject.onNext(INSTANCE_DOWN);
+
+            assertThat(registrationSubscription.isUnsubscribed(), is(false));
+            assertThat(interestSubscriber.takeNext(60, TimeUnit.SECONDS), is(modifyChangeNotificationOf(INSTANCE_DOWN)));
+        } finally {
+            registrationClient.shutdown();
```
tbak: There is no need to cleanup these resources explicitly. EurekaDeployment tracks all registering/interest clients and disconnects them on shutdown. If that does not happen in some cases, there must be a bug.

eureka2-integration/src/test/java/com/netflix/eureka2/integration/server/startup/BridgeServerStartupAndShutdownIntegrationTest.java
```diff
@@ -30,13 +36,15 @@
...
-                    protected void configure() {
-                        bind(DiscoveryClient.class).toInstance(eureka1ServerResource.createDiscoveryClient("bridgeService"));
-                    }
-                };
+            protected Module getModule() {
+                return Modules.override(super.getModule())
+                        .with(new AbstractModule() {
+                            @Override
+                            protected void configure() {
+                                bind(EurekaInstanceConfig.class).toProvider(MyDataCenterInstanceConfigProvider.class).in(Scopes.SINGLETON);
```
tbak: This should not be needed. ```Eureka1ClientResource``` builds InstanceInfo explicitly.

eureka2-server/src/main/java/com/netflix/eureka2/server/audit/AuditServiceController.java
```diff
@@ -42,13 +43,13 @@
...
     private final AuditService auditService;
-    private final SelfInfoResolver serverIdentity;
+    private final Provider<SelfInfoResolver> serverIdentity;
 
     @Inject
     public AuditServiceController(
             SourcedEurekaRegistry registry,
             AuditService auditService,
-            SelfInfoResolver serverIdentity) {
+            Provider<SelfInfoResolver> serverIdentity) {
```
tbak: Why provider in ```SelfInfoResolver``` directly?

eureka2-server/src/main/java/com/netflix/eureka2/server/health/HealthStatusProviderRegistry.java
```diff
@@ -32,7 +34,8 @@ public void add(HealthStatusProvider<?> provider) {
         return providerSubject;
     }
 
-    @PostInjector
+//    @PostInjector
```
tbak: PostInjector is a special annotation that will run annotated methods after injector is initialized. PostConstruct is non-deterministic, and the initialization may happen in any valid order. ```HealthStatusProviderRegistry``` emits events directly after the activate method is called, and if some components are not ready yet they will miss it.

eureka2-testkit/src/main/java/com/netflix/eureka2/testkit/embedded/server/EmbeddedReadServer.java
```diff
@@ -80,21 +81,21 @@ public void start() {
...
-                new EurekaReadServerModule(config, registrationClient, interestClient),
-                new AbstractModule() {
+        Module embeddedReadServerModule = Modules.override(new EurekaReadServerModule(config, registrationClient, interestClient))
+                .with(new AbstractModule() {
                     @Override
                     protected void configure() {
                         if (networkRouter != null) {
                             bind(NetworkRouter.class).toInstance(networkRouter);
-                            bind(TcpInterestServer.class).to(EmbeddedTcpInterestServer.class);
+                            bind(TcpInterestServer.class).to(EmbeddedTcpInterestServer.class).in(Scopes.SINGLETON);
```
tbak: Why in scope is needed here? We have already singleton annotation on the ```EmbeddedTcpInterestServer``` class.

eureka2-write-server/src/main/java/com/netflix/eureka2/server/service/EurekaWriteServerSelfInfoResolver.java
```diff
@@ -27,9 +28,9 @@
     public EurekaWriteServerSelfInfoResolver(
             final EurekaServerConfig config,
             final EurekaHttpServer httpServer,
-            final TcpRegistrationServer registrationServer,
-            final TcpReplicationServer replicationServer,
-            final TcpInterestServer discoveryServer) {
+            final Provider<TcpRegistrationServer> registrationServer,
```
tbak: Why provider?

eureka2-integration/src/test/java/com/netflix/eureka2/integration/client/EurekaClientFailoverTest.java
qiangdavidliu: There seemed to be some issues with these test when run together (one of them fails, but both passes fine when run independently). This points to resource clean up issues and this try finally fixes this.

eureka2-integration/src/test/java/com/netflix/eureka2/integration/server/startup/BridgeServerStartupAndShutdownIntegrationTest.java
qiangdavidliu: There is an upstream change where we install(EurekaModule) instead of just binding DiscoveryClient, which necessitates this here.

eureka2-server/src/main/java/com/netflix/eureka2/server/audit/AuditServiceController.java
qiangdavidliu: We have a circular injection issues between the resolver, the TcpHanders and the TcpServers which necessitates the providers.

eureka2-server/src/main/java/com/netflix/eureka2/server/health/HealthStatusProviderRegistry.java
qiangdavidliu: With governator 1.7 we can @ Inject LifecycleManager and then add a listener for onStarted(). Changing this to avoid needing @ PostInject all together.

eureka2-testkit/src/main/java/com/netflix/eureka2/testkit/embedded/server/EmbeddedReadServer.java
qiangdavidliu: habit:)

eureka2-write-server/src/main/java/com/netflix/eureka2/server/service/EurekaWriteServerSelfInfoResolver.java
qiangdavidliu: See earlier comment on circular injections

PR #561, opened 2015-06-23
1.x Add VPC support

tbak:


tbak:
This was prototype work. With coming changes, we will have to visit VPC support.

PR #565, opened 2015-06-29
2.x (WIP) s3 override registry for OOS status

qiangdavidliu:
Still WIP, need to integrate with #564 before merging.


eureka2-ext/eureka2-overrides-aws/src/main/java/com/netflix/eureka2/server/service/overrides/OverridesCodec.java
```diff
@@ -0,0 +1,20 @@
...
+import com.netflix.eureka2.codec.json.EurekaJsonCodec;
+
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * @author David Liu
+ */
+public final class OverridesCodec {
```
tbak: I would keep codecs together in com.netflix.eureka2.codec package, just like metrics are aggregated.

eureka2-ext/eureka2-overrides-aws/src/main/java/com/netflix/eureka2/server/service/overrides/S3OutOfServiceOverridesSource.java
```diff
@@ -0,0 +1,141 @@
...
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * An optimized s3 backed overrides source for OUT_OF_SERVICE status overrides only.
+ * In most cases overrides are only needed for this specific case, and given this constraint,
+ * we can actually make use of s3 object names for storage.
+ *
+ * The file format will be of the form: {config.getPrefix()}/overrideId
```
tbak: override id == instance id. For clarity what about renaming Overrides.getId() to Overrides.getInstanceId()?

eureka2-ext/eureka2-overrides-aws/src/main/java/com/netflix/eureka2/server/service/overrides/S3OutOfServiceOverridesSource.java
```diff
@@ -0,0 +1,141 @@
...
+        String name = toS3Name(config.getPrefix(), overrides.getId());
+
+        PutObjectResult result = s3Client.putObject(
+                config.getBucketName(),
+                name,
+                inputStream,
+                metadata
+        );
+
+        logger.debug("S3 put result: {}", result);
```
tbak: If operation fails, will this result in an exception or error code will be send in the result? In the latter case we should have error logging added.

eureka2-ext/eureka2-overrides-aws/src/main/java/com/netflix/eureka2/server/service/overrides/S3OverridesModule.java
```diff
@@ -0,0 +1,18 @@
...
+ * to use the S3 backed OUT_OF_SERVICE overrides source.
+ *
+ * @author David Liu
+ */
+public class S3OverridesModule extends AbstractModule {
+    @Override
+    protected void configure() {
+        requireBinding(OverridesRegistry.class);
+
+        bind(LoadingOverridesRegistry.ExternalOverridesSource.class).to(S3OutOfServiceOverridesSource.class);
```
tbak: We provide two alternative sources. Maybe instead of choosing here one of them we should provide a constructor with ExternalOverridesSource class as argument?

eureka2-ext/eureka2-overrides-aws/src/test/java/com/netflix/eureka2/server/service/overrides/S3OutOfServiceOverridesSourceTest.java
```diff
@@ -0,0 +1,74 @@
...
+import org.junit.Ignore;
+import org.junit.Test;
+
+import java.util.Map;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.*;
+
+/**
+ * Ignored as these tests needs an actual s3 backing to execute. Uncomment ignore and execute with the correct
```
tbak: Given we have more tests like these, which are external systems integration tests, it would be good to define common runtime requirements, that once fulfilled would allow anybody to run those tests. In this particular case, AWS credentials would be sufficient.

eureka2-write-server/src/main/java/com/netflix/eureka2/server/service/overrides/LoadingOverridesRegistry.java
```diff
@@ -5,60 +5,72 @@
...
 import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
 
 /**
- * An in memory version of the overrides module primarily useful for testing purposes with the embedded cluster.
+ * A loading overrides registry that stores overrides in an external source, and periodically loads a global
+ * snapshot to a locally backed hash map. Writes to this registry is passed on to the external overrides source,
```
tbak: Writes do not modify local map?

eureka2-write-server/src/main/java/com/netflix/eureka2/server/service/overrides/OverridesModule.java
```diff
@@ -10,10 +11,10 @@
...
     @Override
     protected void configure() {
+        // override this binding with more specific impls in non-local-testing environments
+        bind(LoadingOverridesRegistry.ExternalOverridesSource.class).to(InMemoryOverridesSource.class).in(Scopes.SINGLETON);
 
         bind(OverridesService.class).to(OverridesServiceImpl.class);
-
-        // override this binding with more specific impls in non-local-testing environments
-        bind(OverridesRegistry.class).to(InMemoryOverridesRegistry.class);
+        bind(OverridesRegistry.class).to(LoadingOverridesRegistry.class);
```
tbak: Since it is a very specific overrides setup, maybe the module should be renamed to InMemoryOverridesModule? Or we could provide alternatively a builder for the module itself.

eureka2-ext/eureka2-overrides-aws/src/main/java/com/netflix/eureka2/server/service/overrides/OverridesCodec.java
qiangdavidliu: Agree. This is currently in an awkward place here but it's also slightly different in that it's not strictly for the eureka transport layer. Some small reorg of the transport and the codecs may be necessary.

eureka2-ext/eureka2-overrides-aws/src/main/java/com/netflix/eureka2/server/service/overrides/S3OutOfServiceOverridesSource.java
qiangdavidliu: It's not strictly the instanceId (in that EC2 instance sense). The id is the id of the InstanceInfo, which is uniquely identifiable in ec2 or in containers etc. Will rename to instanceInfoId.

eureka2-ext/eureka2-overrides-aws/src/main/java/com/netflix/eureka2/server/service/overrides/S3OutOfServiceOverridesSource.java
qiangdavidliu: If the PutObjectResult is returned, then the file is created. Failed putObject() will exception (which is caught at an upper layer).

eureka2-ext/eureka2-overrides-aws/src/main/java/com/netflix/eureka2/server/service/overrides/S3OverridesModule.java
qiangdavidliu: Yes. I'm going to take a look at the other pull but there'll need to be a way to merge the ASG strategy and this one.

eureka2-write-server/src/main/java/com/netflix/eureka2/server/service/overrides/LoadingOverridesRegistry.java
qiangdavidliu: Good catch. Should eagerly update the local map here if the underlying put to the external store is successful.

qiangdavidliu:
Still todo: adding integ tests

eureka2-core/src/main/java/com/netflix/eureka2/transport/EurekaTransports.java
```diff
@@ -89,7 +90,7 @@
     static final String INTEREST_ENVELOPE_TYPE = "com.netflix.eureka2.protocol.interest.InterestMessage";
 
     static final Class<?>[] INTEREST_PROTOCOL_MODEL = {
-            InterestRegistration.class, UnregisterInterestSet.class, Heartbeat.class,
+            InterestRegistration.class, UnregisterInterestSet.class, Heartbeat.class, DeltaDTO.class,
```
tbak: Only top level classes must be registered here. Do we send DeltaDTO as a top level message?

eureka2-ext/eureka2-aws/src/main/java/com/netflix/eureka2/ext/aws/S3OverridesConfiguration.java
```diff
@@ -0,0 +1,14 @@
+package com.netflix.eureka2.ext.aws;
+
+import com.netflix.archaius.annotations.DefaultValue;
+
+/**
+ * @author David Liu
+ */
+public interface S3OverridesConfiguration extends AwsConfiguration {
```
tbak: Does it make sense to have single AwsConfiguration interface for all AWS configuration data?

eureka2-write-server/src/main/java/com/netflix/eureka2/server/EurekaWriteServerModule.java
```diff
@@ -104,5 +102,7 @@ public void configureEureka() {
         bind(SelfRegistrationService.class).to(EurekaWriteServerSelfRegistrationService.class).asEagerSingleton();
 
         bind(AbstractEurekaServer.class).to(EurekaWriteServer.class);
+
+        install(new OverridesModule());
```
tbak: This module is installed directly in WriteServerRunner. It is not installed in the embedded cluster.

eureka2-write-server/src/main/java/com/netflix/eureka2/server/service/overrides/OverridesModule.java
```diff
@@ -11,10 +11,12 @@
...
-        multibinder.addBinding().to(OverridesServiceImpl.class);
 
-        // override this binding with more specific impls in non-local-testing environments
-        bind(OverridesRegistry.class).to(InMemoryOverridesRegistry.class);
+        bind(InstanceStatusOverridesSource.class).to(InMemoryStatusOverridesRegistry.class);
+        bind(InstanceStatusOverridesView.class).to(InMemoryStatusOverridesRegistry.class);
+
+        MapBinder<Integer, OverridesService> mapbinder = MapBinder.newMapBinder(binder(), Integer.class, OverridesService.class);
+        // for ordering
+        mapbinder.addBinding(0).to(InstanceStatusOverridesService.class);
```
tbak: Does it mean there is always in memory override registry?

eureka2-ext/eureka2-aws/src/main/java/com/netflix/eureka2/ext/aws/AwsServiceModule.java
```diff
@@ -22,9 +25,16 @@
     @Override
     protected void configure() {
         bind(AmazonAutoScaling.class).toProvider(AmazonAutoScalingProvider.class);
+        bind(AmazonS3Client.class).toProvider(AmazonS3ClientProvider.class);
 
-        Multibinder<OverridesService> multibinder = Multibinder.newSetBinder(binder(), OverridesService.class);
-        multibinder.addBinding().to(AsgOverrideService.class);
+        bind(InstanceStatusOverridesView.class).to(AsgStatusOverridesView.class);
+        bind(InstanceStatusOverridesView.class).to(S3StatusOverridesRegistry.class);
```
tbak: Binding twice to the same InstanceStatusOverridesView class.

eureka2-core/src/main/java/com/netflix/eureka2/transport/EurekaTransports.java
qiangdavidliu: This is legacy from the earlier pull when we still had the Overrides datastructure. Should be removed.

eureka2-ext/eureka2-aws/src/main/java/com/netflix/eureka2/ext/aws/S3OverridesConfiguration.java
qiangdavidliu: In this case for the S3 specific use case we need to have some S3 specific properties (bucketName and prefix). It extends AwsConfiguration.

eureka2-write-server/src/main/java/com/netflix/eureka2/server/EurekaWriteServerModule.java
qiangdavidliu: We had specific bindings for override stuff that breaks the write server when removed. What we need is some no-op bindings.

PR #566, opened 2015-06-30
2.x Migrate to archaius2 configuration proxies.

tbak:


eureka2-bridge/src/main/resources/eureka-bridge-server.properties
```diff
@@ -1,19 +1,21 @@
...
 
-# Eureka 2 admin UI properties
-#eureka.client.discovery-endpoint.dns=localhost
-eureka.client.discovery-endpoint.port=12303
+eurekaRegistry.evictionTimeoutMs=30000
+eurekaRegistry.evictionStrategyType=PercentageDrop
+eurekaRegistry.evictionStrategyValue=5
 
-eureka.services.bridge.refreshRateSec=30
+refreshRateSec=30
```
qiangdavidliu: Add a sensible prefix to this config?

qiangdavidliu:
Much cleaner:)

eureka2-bridge/src/main/resources/eureka-bridge-server.properties
tbak: I have forgotten to add prefix here. The default is "eureka2.".

PR #592, opened 2015-07-23
Binding exception due EurekaInstanceConfig was already configured in …

xbaran:
ExampleEurekaGovernatedService does not work because a wrong guice binding. 

The EurekaInstanceConfig is already configured in EurekaModule and also in anonymous AbstractModule. So, it causes guice to throw already configured exception. I just override EurekaModule with AbstractModule via Modules.override() method.


qiangdavidliu:
@xbaran thanks for the fix.

PR #601, opened 2015-08-03
Remove unused reference to DynamicPropertyFactory

elandau:
This triggers Archaius static initialization, which is not necessary here.


tbak:
Thanks

PR #612, opened 2015-08-06
Adding server side capability to return compact json data

qiangdavidliu:


eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
```diff
@@ -326,10 +325,10 @@ public synchronized BackupRegistry get() {
             }
 
             DiscoveryJerseyProvider discoveryJerseyProvider = new DiscoveryJerseyProvider(
-                    CodecWrapper.get(CodecType.from(clientConfig.getJsonCodecName())),
-                    CodecWrapper.get(CodecType.from(clientConfig.getXmlCodecName()))
+                    CodecWrappers.getEncoder(clientConfig.getEncoderName()),
```
tbak: After stabilization period we shall remove all legacy codecs which will leave us with options:
* XML vs JSON (== MediaType)
* full vs compact1 vs compact2 (== scope)
I think public API shall be aligned with this target, while we will have to provide temporary means to specify an exact codec to use.

eureka-client/src/main/java/com/netflix/discovery/converters/wrappers/DecoderWrapper.java
```diff
@@ -0,0 +1,14 @@
+package com.netflix.discovery.converters.wrappers;
+
+import java.io.IOException;
+import java.io.InputStream;
+
+/**
+ * @author David Liu
+ */
+public interface DecoderWrapper extends CodecWrapperBase {
```
tbak: Shall we drop 'wrapper' term, as this is more like Eureka codec API now? (for example EurekaCodecBase, ...)

eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
qiangdavidliu: Agree. The public REST API current does not care about codecs, but takes a X-Eureka-Accept header. DiscoveryClient (for now, during the transition) translates between the specified decoder and the appropriate X-Eureka-Accept (translation in EurekaAccept.java)

eureka-client/src/main/java/com/netflix/discovery/converters/wrappers/DecoderWrapper.java
qiangdavidliu: Either way works. Once we move beyond the transitions phase and deprecate the older codecs, the wrappers can be removed and the actual codec classes be moved to implement the Encoder and Decoder interface.

PR #614, opened 2015-08-07
New sub-project to configure eureka using archaius2

elandau:


eureka-client-archaius2/build.gradle
```diff
@@ -0,0 +1,14 @@
+apply plugin: 'nebula-test-jar'
+
+dependencies {
+    compile project(':eureka-core')
```
qiangdavidliu: eureka (1.x) has a weird structure where eureka-client is the base module. eureka-core is really for server side only.

brharrington:
Can the legacy archaius bindings be pulled out to a separate lib so that if using archaius2 we don't have to pull in the legacy version?

elandau:
Removing the legacy libs would require a bit more work.  @qiangdavidliu may take a stab at it.

eureka-client-archaius2/src/main/java/com/netflix/discovery/Ec2EurekaArchaius2InstanceConfig.java
```diff
@@ -0,0 +1,112 @@
...
+ * 
+ * 
+ * @author elandau
+ *
+ */
+@Singleton
+public class Ec2EurekaArchaius2InstanceConfig extends EurekaArchaius2InstanceConfig {
+    private static final Logger LOG = LoggerFactory.getLogger(Ec2EurekaArchaius2InstanceConfig.class);
+    
+    private static final String DEFAULT_NAMESPACE = "eureka";
```
qiangdavidliu: move this to the super class and then both can reference it instead of having a magic "eureka" in EurekaArchaius2InstanceConfig?

eureka-client-archaius2/src/main/java/com/netflix/discovery/EurekaArchaius2InstanceConfig.java
```diff
@@ -0,0 +1,146 @@
...
+
+    private final String defaultAppGroup;
+    
+    @Inject
+    public EurekaArchaius2InstanceConfig(Config config) {
+        this(config, "eureka");
+    }
+    
+    public EurekaArchaius2InstanceConfig(Config config, String namespace) {
+        this(config, "eureka", new DataCenterInfo() {
```
qiangdavidliu: use namespace instead of "eureka".

eureka-client-archaius2/src/main/java/com/netflix/discovery/EurekaArchaius2ClientConfig.java
```diff
@@ -0,0 +1,190 @@
...
+    private final String defaultRegion;
+
+    @Inject
+    public EurekaArchaius2ClientConfig(Config config) {
+        this(config, DEFAULT_NAMESPACE);
+    }
+
+    public EurekaArchaius2ClientConfig(Config config, String namespace) {
+        this.defaultRegion = config.getString("@region", null);
+        this.config = config.getPrefixedView(DEFAULT_NAMESPACE);
```
qiangdavidliu: use namespace instead of DEFAULT

eureka-client/src/main/java/com/netflix/discovery/providers/DefaultEurekaClientConfigProvider.java
```diff
@@ -17,21 +16,10 @@
...
-            } else {
-                config = new DefaultEurekaClientConfig(namespace);
-            }
-
-            // TODO: Remove this when DiscoveryManager is finally no longer used
-            DiscoveryManager.getInstance().setEurekaClientConfig(config);
-        }
-
-        return config;
+        return namespace == null
```
qiangdavidliu: why remove the singleton check?

PR #618, opened 2015-08-18
Features/jackson codec ng

bondj:
Add some interns to save memory.

### DIFF
--- a/eureka-client/src/main/java/com/netflix/appinfo/InstanceInfo.java
+++ b/eureka-client/src/main/java/com/netflix/appinfo/InstanceInfo.java
@@ -37,6 +37,7 @@ import com.netflix.discovery.converters.EurekaJacksonCodec.InstanceInfoSerialize
 import com.netflix.discovery.provider.Serializer;
 import com.thoughtworks.xstream.annotations.XStreamAlias;
 import com.thoughtworks.xstream.annotations.XStreamOmitField;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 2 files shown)


eureka-client/src/main/java/com/netflix/appinfo/InstanceInfo.java
```diff
@@ -155,23 +159,23 @@ public InstanceInfo(
             @JsonProperty("lastDirtyTimestamp") Long lastDirtyTimestamp,
             @JsonProperty("actionType") ActionType actionType,
             @JsonProperty("asgName") String asgName) {
-        this.appName = appName;
-        this.appGroupName = appGroupName;
-        this.ipAddr = ipAddr;
-        this.sid = sid;
+        this.appName = appName!=null ? appName.intern():null;
+        this.appGroupName = appGroupName!=null? appGroupName.intern():null;
+        this.ipAddr = ipAddr!=null?ipAddr.intern():null;
```
brharrington: Is the savings mostly because this data is repeated for many instances of the sample application? Why would you have many InstanceInfo objects with the same ip address? With recent jvms the default string table size is around 60k and I would expect interning the ips would quickly lead to collisions causing the performance to go down.

eureka-client/src/main/java/com/netflix/appinfo/InstanceInfo.java
```diff
@@ -155,23 +159,23 @@ public InstanceInfo(
...
         this.isUnsecurePortEnabled = portEnabled;
         this.securePort = securePort;
         this.isSecurePortEnabled = securePortEnabled;
-        this.homePageUrl = homePageUrl;
-        this.statusPageUrl = statusPageUrl;
-        this.healthCheckUrl = healthCheckUrl;
-        this.secureHealthCheckUrl = secureHealthCheckUrl;
-        this.vipAddress = vipAddress;
-        this.secureVipAddress = secureVipAddress;
+        this.homePageUrl = homePageUrl!=null?homePageUrl.intern():null;
```
brharrington: Looking at the discovery metadata for my apps these urls are absolute and contain the hostname. Would there actually be many copies of the same value?

PR #630, opened 2015-09-02
EurekaNameSpace not used in eureka.region

jaume-pinyol:
Hello,

I've seen that the eureka.region property in DefaultEurekaClientConfig.class it's not tied to the namespace, and when using a different namespace there is the need to hardcode this property.

Regards,
Jaume


tbak:
Thank you for your PR. To be backward compatible, can you add a fallback logic, so if a property with a namespace does not exist, it should read "eureka.region" as previously.

jaume-pinyol:
Done! Do you want to join commits?

tbak:
Thanks!

PR #644, opened 2015-09-18
Feature/DI friendly server

qiangdavidliu:


eureka-client/src/main/java/com/netflix/discovery/EurekaClient.java
```diff
@@ -94,23 +94,31 @@
...
 
     /**
+     * @deprecated see {@link com.netflix.discovery.endpoint.EndpointUtils} for replacement
+     *
      * Get the list of all eureka service urls for the eureka client to talk to.
      *
      * @param zone the zone in which the client resides
      * @return The list of all eureka service urls for the eureka client to talk to.
      */
+    @Deprecated
```
tbak: The replacement is not one to one, as EndpointUtils expects more arguments. Shall we create a new interface  for zone management? I am working now on transport refactoring, where I have introduced `ClusterResolver` abstraction. Zone analysis is done actually because of the transport needs, so probably lots of this logic will be used there.

eureka-client/src/main/java/com/netflix/discovery/NotImplementedRegistryImpl.java
```diff
@@ -12,11 +12,11 @@
 
     @Override
     public Applications fetchRegistry() {
-        throw new UnsupportedOperationException("Backup registry not implemented.");
+        return null;
```
tbak: As we change the behavior lets specify the contract in more details in `BackupRegistry` class.

eureka-client/src/test/java/com/netflix/discovery/converters/wrappers/CodecWrappersTest.java
```diff
@@ -0,0 +1,61 @@
...
+        Assert.assertNull(CodecWrappers.getEncoder(testWrapperName));
+        Assert.assertNull(CodecWrappers.getDecoder(testWrapperName));
+
+        CodecWrappers.registerWrapper(new TestWrapper());
+
+        Assert.assertNotNull(CodecWrappers.getEncoder(testWrapperName));
+        Assert.assertNotNull(CodecWrappers.getDecoder(testWrapperName));
+    }
+
+    private final class TestWrapper implements EncoderDecoderWrapper {
```
tbak: Is EncoderDecoderWrapper becoming our public API? The original intent was to support the migration process and erase it. For long-term public API we should make a few adjustments.

eureka-core/src/main/java/com/netflix/eureka/PeerAwareInstanceRegistry.java
```diff
@@ -1,10 +0,0 @@
-package com.netflix.eureka;
-
-import com.netflix.appinfo.InstanceInfo;
-
-/**
- * @author Tomasz Bak
- */
-public interface PeerAwareInstanceRegistry extends InstanceRegistry {
```
tbak: These abstractions were introduced for sake of unit testing, so we do not have to instantiate concrete objects. I think there is value in keeping those.

eureka-core/src/main/java/com/netflix/eureka/aws/EIPManager.java
```diff
@@ -63,29 +71,109 @@
...
+    private void scheduleEIPBindTask() {
+        timer.schedule(new TimerTask() {
+                           @Override
+                           public void run() {
+                               try {
+                                   // If the EIP is not bound, the registry could  be stale
+                                   // First sync up the registry from the neighboring node before
+                                   // trying to bind the EIP
+                                   if (!isEIPBound()) {
+                                       registry.clearRegistry();
```
tbak: Do we have to do that? If there is false positive, we will wipe out whole registry, and force all clients to do the full registry fetch. They service is up for the whole time so clients may get empty/partial registry content. As we have eventually consistent system, there is no harm in skipping this step.

eureka-core/src/main/java/com/netflix/eureka/registry/AbstractInstanceRegistry.java
```diff
@@ -86,32 +82,51 @@
...
-    protected AbstractInstanceRegistry() {
-        recentCanceledQueue = new CircularQueue<Pair<Long, String>>(1000);
-        recentRegisteredQueue = new CircularQueue<Pair<Long, String>>(1000);
-        deltaRetentionTimer.schedule(getDeltaRetentionTask(),
-                EUREKA_CONFIG.getDeltaRetentionTimerIntervalInMs(),
-                EUREKA_CONFIG.getDeltaRetentionTimerIntervalInMs());
+    protected AbstractInstanceRegistry(EurekaServerConfig serverConfig, ServerCodecs serverCodecs) {
+        this.serverConfig = serverConfig;
+        this.serverCodecs = serverCodecs;
+        this.recentCanceledQueue = new CircularQueue<Pair<Long, String>>(1000);
```
tbak: Not used in the code. Can be removed with its corresponding method.

eureka-core/src/main/java/com/netflix/eureka/registry/AbstractInstanceRegistry.java
```diff
@@ -1236,67 +1215,11 @@ public boolean offer(E e) {
...
-            }
-        }
-        logger.debug("Returning the default instance status {} for instance {}",
-                r.getStatus(), r.getId());
-        return r.getStatus();
-    }
-
+    // FIXME this should really be part of peerAware registry as it has knowledge of replication ...
+    // fix in next round of code clean up
+    protected abstract InstanceInfo.InstanceStatus getOverriddenInstanceStatus(InstanceInfo r,
```
tbak: The replication flag comes from REST API call, and we pass it down from there. I think it is fine to have this logic here.

eureka-core/src/main/java/com/netflix/eureka/registry/AbstractInstanceRegistry.java
```diff
@@ -1315,21 +1238,4 @@ public void run() {
 
         };
     }
-
-    protected void initRemoteRegionRegistry() throws MalformedURLException {
```
tbak: Remote region abstraction is parallel to PeerAwareInstanceRegistry, I would leave it here as we actually use the remote data in this class.

eureka-core/src/main/java/com/netflix/eureka/registry/AwsInstanceRegistry.java
```diff
@@ -0,0 +1,122 @@
...
+                               EurekaClientConfig clientConfig,
+                               ServerCodecs serverCodecs,
+                               EurekaClient eurekaClient) {
+        super(serverConfig, clientConfig, serverCodecs, eurekaClient);
+    }
+
+    @Override
+    public void init(PeerEurekaNodes peerEurekaNodes) throws Exception {
+        super.init(peerEurekaNodes);
+        this.awsAsgUtil = new AwsAsgUtil(serverConfig, clientConfig, this);
```
tbak: It is no longer util. Can we convert this into a regular service with an interface and implementation, and inject it via constructor?

eureka-core/src/main/java/com/netflix/eureka/PeerAwareInstanceRegistry.java
tbak: Never mind, I have seen it was just moved.

eureka-core/src/main/java/com/netflix/eureka/registry/AbstractInstanceRegistry.java
```diff
@@ -86,32 +82,51 @@
...
+        this.recentCanceledQueue = new CircularQueue<Pair<Long, String>>(1000);
+        this.recentRegisteredQueue = new CircularQueue<Pair<Long, String>>(1000);
+        this.deltaRetentionTimer.schedule(getDeltaRetentionTask(),
+                serverConfig.getDeltaRetentionTimerIntervalInMs(),
+                serverConfig.getDeltaRetentionTimerIntervalInMs());
+
+    }
+
+    @Override
+    public synchronized void initializedResponseCache() {
```
tbak: Do we need this lazy initialization?

eureka-core/src/main/java/com/netflix/eureka/registry/PeerAwareInstanceRegistryImpl.java
```diff
@@ -115,38 +115,80 @@ public int compare(Application l, Application r) {
...
+    @Override
+    public void shutdown() {
+        try {
+            DefaultMonitorRegistry.getInstance().unregister(Monitors.newObjectMonitor(this));
+        } catch (Throwable t) {
+            logger.error("Cannot shutdown monitor registry", t);
+        }
+        try {
+            if (peerEurekaNodes != null) {
+                for (PeerEurekaNode node : this.peerEurekaNodes.getPeerEurekaNodes()) {
```
tbak: PeerEurekaNodes has shutdown method.

eureka-core/src/main/java/com/netflix/eureka/registry/ResponseCache.java
```diff
@@ -0,0 +1,42 @@
...
+
+import javax.annotation.Nullable;
+import java.util.concurrent.atomic.AtomicLong;
+
+/**
+ * @author David Liu
+ */
+public interface ResponseCache {
+
+    public enum KeyType {
```
tbak: Shall be in Key class?

eureka-core/src/main/java/com/netflix/eureka/registry/ResponseCache.java
```diff
@@ -0,0 +1,42 @@
...
+     * <p>
+     * If the cached information is not available it is generated on the first
+     * request. After the first request, the information is then updated
+     * periodically by a background thread.
+     * </p>
+     *
+     * @param key the key for which the cached information needs to be obtained.
+     * @return payload which contains information about the applications.
+     */
+    <KEY> String get(KEY key);
```
tbak: Why not class level type parameter?

eureka-core/src/test/java/com/netflix/eureka/resources/ApplicationResourceTest.java
```diff
@@ -32,12 +32,12 @@ public void setUp() throws Exception {
...
 
-        applicationResource = new ApplicationResource(testApplication.getName(), registry, new ResponseCache(registry));
+        applicationResource = new ApplicationResource(testApplication.getName(), serverContext.getServerConfig(), serverContext.getRegistry());
 
         for (InstanceInfo instanceInfo : testApplication.getInstances()) {
-            InstanceInfo changed = new InstanceInfo.Builder(instanceInfo)
-                    .setASGName(null).build();  // null asgName to get around AwsAsgUtil check
-            registry.register(changed, false);
+//            InstanceInfo changed = new InstanceInfo.Builder(instanceInfo)
+//                    .setASGName(null).build();  // null asgName to get around AwsAsgUtil check
```
tbak: Dead code

eureka-core/src/test/java/com/netflix/eureka/resources/ApplicationsResourceTest.java
```diff
@@ -33,13 +33,13 @@ public void setUp() throws Exception {
...
-        applicationsResource = new ApplicationsResource(registry, new ResponseCache(registry));
+        applicationsResource = new ApplicationsResource(serverContext);
 
         for (Application application : testApplications.getRegisteredApplications()) {
             for (InstanceInfo instanceInfo : application.getInstances()) {
-                InstanceInfo changed = new InstanceInfo.Builder(instanceInfo)
-                        .setASGName(null).build();  // null asgName to get around AwsAsgUtil check
-                registry.register(changed, false);
+//                // null asgName to get around AwsAsgUtil check
+//                InstanceInfo changed = new InstanceInfo.Builder(instanceInfo).setASGName(null).build();
```
tbak: Remove?

eureka-core/src/test/java/com/netflix/eureka/util/AwsAsgUtilTest.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    @After
+    public void tearDown() throws Exception {
+        ConfigurationManager.getConfigInstance().clear();
+    }
+
+    @Test
+    public void testAsyncLoadingFromCache() {
+
```
tbak: No actual test

eureka-core/src/main/java/com/netflix/eureka/registry/AbstractInstanceRegistry.java
tbak: Actually it is used by UI. Please, ignore it

eureka-client/src/main/java/com/netflix/discovery/EurekaClient.java
qiangdavidliu: The PeerEurekaNodes on the server side also uses this method, so it should really not be dependent on EurekaClient for this functionality. Moving it to EndpointUtils was just a convenient way of moving this out of the client side interface. Feel free to enhance this to as part of the transport work.

eureka-client/src/test/java/com/netflix/discovery/converters/wrappers/CodecWrappersTest.java
qiangdavidliu: Good point. I think we do need an abstraction over our codecs, as even when we standardize on a single codec in the future, there needs to be a way to add additional codecs for extension purposes.

The APIs as they exist today are pretty bare and generic. We can always add additional ones in the future as needed.

eureka-core/src/main/java/com/netflix/eureka/aws/EIPManager.java
qiangdavidliu: This code has always been here ... But a good point on what the correct action should be here. As this is code that has worked well for a long time now, let's not change it immediately without doing some more testing.

eureka-core/src/main/java/com/netflix/eureka/registry/AbstractInstanceRegistry.java
qiangdavidliu: Agree with the logic placement. The comment was more of a "comment" on the structure of the interfaces (e.g. the non-peer aware registry should not understand replication). Unfortunately we are stuck with some of this without doing major surgery.

eureka-core/src/main/java/com/netflix/eureka/registry/AbstractInstanceRegistry.java
qiangdavidliu: I buy that.

eureka-core/src/main/java/com/netflix/eureka/registry/AbstractInstanceRegistry.java
qiangdavidliu: Yes, because the cache needs a reference to the registry unfortunately.

eureka-core/src/main/java/com/netflix/eureka/registry/AwsInstanceRegistry.java
qiangdavidliu: I agree. But we can fix this in the next round of refactoring.

eureka-core/src/main/java/com/netflix/eureka/registry/PeerAwareInstanceRegistryImpl.java
qiangdavidliu: indeed. Thanks.

eureka-core/src/main/java/com/netflix/eureka/registry/ResponseCache.java
qiangdavidliu: Thanks. Forgot to move it.

eureka-core/src/test/java/com/netflix/eureka/util/AwsAsgUtilTest.java
qiangdavidliu: Urg must have forgotten to do this last time I added the test class.

eureka-core/src/main/java/com/netflix/eureka/registry/ResponseCache.java
qiangdavidliu: If it at the class level then it'll need to be propagated all the way up to the PeerAwareRegistry, which gets messy. I'm not happy with the way either. Maybe this is over wrapping it and we should just let class Key be part of the interface.

tbak: In discovery-readonly I am doing exactly this. It is very unlikely we will
reuse this code in different scenarios.

eureka-client/src/test/java/com/netflix/discovery/converters/wrappers/CodecWrappersTest.java
jkschneider: I'm curious what the purpose of supporting multiple codecs is.  Is it to be less opinionated about dependencies provided to downstreams?  If so, could you pick one and shade it instead?

qiangdavidliu: The initial purpose of adding this was to aid in the migration of legacy codecs to new codecs, with the ability to switch between codecs quickly (without having rebuild and redeploy) for compatibility purposes. In the case of the Jackson json codecs for example, the two codecs does not actually have any dependency differences (both use jackson), but they handle the encode and decode differently.

qiangdavidliu:
Closed in favour of #652 which merges into a branch first.

PR #646, opened 2015-09-19
Enable custom metadata-aware DataCenterInfo implementations

jkschneider:
Currently, `DataCenterInfo` metadata is only serialized/deserialized if it is an instance of `AmazonInfo`.  It is sometimes useful to be able to hang metadata on non-AmazonInfo `DataCenterInfo` instances (e.g. MyDataCenterInfo) as well.

This PR introduces a new `MetadataAware` interface, and both `AmazonInfo` and `MyDataCenterInfo` are made to extend it.  Any `DataCenterInfo` implementation that is also `MetadataAware` will have its metadata serialized/deserialized.

Since `MyDataCenterInfo` did not previously support metadata, its metadata map is initialized to an empty map (effectively preserving its state prior to the PR).

This PR blocks progress in Spring Cloud Netflix on adoption of eureka-client 1.2.2 and later.  After merge, we can effectively get rid of the [custom codec](https://github.com/spring-cloud/spring-cloud-netflix/blob/master/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/eureka/DataCenterAwareJacksonCodec.java) in Spring Cloud Netflix.

### DIFF
--- a/eureka-client/src/main/java/com/netflix/discovery/converters/Converters.java
+++ b/eureka-client/src/main/java/com/netflix/discovery/converters/Converters.java
@@ -25,13 +25,10 @@ import java.util.Iterator;
 import java.util.Map;
 import java.util.Map.Entry;
 
-import com.netflix.appinfo.AmazonInfo;
-import com.netflix.appinfo.DataCenterInfo;
+import com.netflix.appinfo.*;
 import com.netflix.appinfo.DataCenterInfo.Name;
-import com.netflix.appinfo.InstanceInfo;
 import com.netflix.appinfo.InstanceInfo.InstanceStatus;
 import com.netflix.appinfo.InstanceInfo.PortType;
-import com.netflix.appinfo.LeaseInfo;
 import com.netflix.discovery.shared.Application;
 import com.netflix.discovery.shared.Applications;
 import com.netflix.discovery.util.StringCache;
@@ -450,15 +447,15 @@ public final class Converters {
             writer.setValue(info.getName().name());
             writer.endNode();
 
-            if (info.getName() == Name.Amazon) {
-                AmazonInfo aInfo = (AmazonInfo) info;
+            if (info instanceof MetadataAware) {
+                MetadataAware metadataAwareInfo = (MetadataAware) info;
                 writer.startNode(NODE_METADATA);
                 // for backward compat. for now
-                if (aInfo.getMetadata().size() == 0) {
+                if (metadataAwareInfo.getMetadata().size() == 0) {
                     writer.addAttribute("class",
                             "java.util.Collections$EmptyMap");
                 }
-                context.convertAnother(aInfo.getMetadata());
+                context.convertAnother(metadataAwareInfo.getMetadata());
                 writer.endNode();
             }
         }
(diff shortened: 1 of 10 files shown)


eureka-client/src/main/java/com/netflix/discovery/converters/Converters.java
```diff
@@ -475,40 +472,35 @@ public void marshal(Object source, HierarchicalStreamWriter writer,
...
-                        final DataCenterInfo.Name name =
-                                DataCenterInfo.Name.valueOf(dataCenterName);
-                        info = new DataCenterInfo() {
-
-                            @Override
-                            public Name getName() {
-                                return name;
-                            }
-                        };
+                        info = new MyDataCenterInfo(DataCenterInfo.Name.valueOf(dataCenterName));
```
spencergibb: If it's newing up `MyDataCenterInfo` how can a custom one be used?

jkschneider: I suppose we would use `MyDataCenterInfo` rather than a custom type.  Because of all that silliness around serializing any non-AmazonInfo with a @class attribute of  `com.netflix.appinfo.InstanceInfo$DefaultDataCenterInfo`, we are down to a binary choice: AmazonInfo or not.  At least this way we can hang metadata on our non-AmazonInfo choice.

jkschneider: We could be certain that we preserve existing behavior AND support a fully custom DataCenterInfo implementation if we add a fourth `Custom` type to the `DataCenterInfo.Name` enum, where we serialize the fully qualified class name of the custom type.  

@tbak Would this break any eureka-clients in a cloud using a version less than 1.2.5+?

spencergibb: OK, if that's the case, the point of this was so we could set the id of the `DataCenterInfo`. `MyDataCenterInfo` doesn't implement `UniqueIdentifier`.

jkschneider: 87de5c9 makes `MyDataCenterInfo` a `UniqueIdentifier`, but alters the contract of `UniqueIdentifier` to make its id nullable.  `InstanceInfo.getId` has been updated accordingly.

qiangdavidliu:
@jkschneider thanks for the pull. Regarding your question on compatibility of adding a new Name enum, it would break older clients who while fail at Name.valueOf("ANewCustomName") unfortunately.

tbak:
Eureka1 default serialization engine is limited in practice to DataCenterInfo and AmazonInfo types. If a custom implementation is required, it should come with its serialization engine, possibly extending the default one.

We are in the process of migration from legacy encoding mechanism, that was both slow and memory hungry. We will remove the old code, as soon as we have enough confidence that all corner cases are covered in the new implementation. For the purpose of the migration we introduced an internal CodecWrapper API, but I would not  recommend to use it may be changed or removed in the near future.

In Eureka2 we provide BasicDataCenterInfo, in parallel with AWS implementation, which is aimed at DC deployments. I believe it makes sense to have it in Eureka1 as well. If this would solve your problem, lets go that route.

eureka-client/src/main/java/com/netflix/discovery/converters/Converters.java
tbak: Converters provides XStream based serialization that we want to remove in the near future.

jkschneider:
@tbak If BasicDataCenterInfo contains a customizable ID, then that is sufficient.  Custom metadata is icing on the cake, but there is no immediate need for it in Spring Cloud.

qiangdavidliu:
@jkschneider @spencergibb #649 should provide a way to specify custom ids for InstanceInfo without having to go through DataCenterInfo.

PR #649, opened 2015-09-23
Decoupling the id of InstanceInfo from the id of the DataCenterInfo.

qiangdavidliu:
For compatibility ease, reusing the deprecated "sid" (serviceId) field.
Also some minor reformatting.


spencergibb:
:+1: 
@dsyer this will eliminate the need for us to do any custom codec stuff.

tbak:
Looks good

qiangdavidliu:
Turns out there are some lingering internal compatibility issues with reusing sid. Going to rework this pull and add a new field (client usage should be similar).

PR #650, opened 2015-09-23
1.x Encode @class entry in InstanceInfo.metadata for backward compatibility

tbak:


eureka-client/src/main/java/com/netflix/discovery/converters/jackson/InstanceInfoJsonBeanSerializer.java
```diff
@@ -29,8 +32,18 @@
  * @author Tomasz Bak
  */
 class InstanceInfoJsonBeanSerializer extends BeanSerializer {
-    InstanceInfoJsonBeanSerializer(BeanSerializerBase src) {
+
+    private static final Map<String, String> EMPTY_MAP = Collections.singletonMap("@class", "java.util.Collections$EmptyMap");
```
qiangdavidliu: add a comment about why this is necessary?

qiangdavidliu: nvm saw it was commented on later.

eureka-client/src/main/java/com/netflix/discovery/converters/jackson/EurekaJsonJacksonCodec.java
```diff
@@ -37,7 +39,7 @@ public EurekaJsonJacksonCodec() {
     public EurekaJsonJacksonCodec(final KeyFormatter keyFormatter, boolean compact) {
```
qiangdavidliu: I can imaging potential future cases where we may have multiple different compact types. As this is exposed as public API, can we change this from a boolean to a more flexible type?

tbak: I wouldn't say this is public API. All codec stuff is internal
implementation until we provide "official" codec API.

qiangdavidliu: It's in a public constructor and there are already cases of extensions of the Codec classes.

qiangdavidliu:
ship it.

Issue #647, opened 2015-09-21
Eureka gives 500 when app registered without Datacenter info


diasjorge:
Hi,

I was testing Eureka and found that if I do a push to register and app to:

SERVER/v2/apps/APP_NAME and in the json data I only have something like:

```
{
    "instance": {
        "hostName":"eureka-testing.example.com",
        "app":"eureka-testing",
        "ipAddr": "127.0.0.1",
        "vipAddress":"eureka-testing",
        "status":"UP",
        "port":"8080"
    }
}
```

The status page then gives a 500.

I think the submitted data should be validated against the DTD shown in the wiki or at least the page should not fail.

tbak:
We do not do any schema based validation (it is JSON), but the request should not fail with 500 in such case. As each instance record must have associated data center info, an HTTP response with status code 400 would be more meaningful here.
Thank you for reporting this.

tbak:
Fixed in PR https://github.com/Netflix/eureka/pull/651

PR #651, opened 2015-09-23
1.x Better error handling for invalid content in request

tbak:
issue: https://github.com/Netflix/eureka/issues/647

### DIFF
--- a/eureka-client/src/main/java/com/netflix/discovery/provider/DiscoveryJerseyProvider.java
+++ b/eureka-client/src/main/java/com/netflix/discovery/provider/DiscoveryJerseyProvider.java
@@ -91,7 +91,10 @@ public class DiscoveryJerseyProvider implements MessageBodyWriter, MessageBodyRe
 
     @Override
     public boolean isReadable(Class serializableClass, Type type, Annotation[] annotations, MediaType mediaType) {
-        return checkForAnnotation(serializableClass);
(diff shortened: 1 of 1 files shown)


eureka-client/src/main/java/com/netflix/discovery/provider/DiscoveryJerseyProvider.java
```diff
@@ -263,4 +218,33 @@ private static ISerializer getSerializer(@SuppressWarnings("rawtypes") Class ser
...
+    }
+
+    private static Response createErrorReply(int status, String errorMessage, MediaType mediaType) {
+        String message;
+        if (MediaType.APPLICATION_JSON_TYPE.equals(mediaType)) {
+            message = "{\"error\": \"" + errorMessage + "\"}";
+        } else {
+            message = "<error><message>" + errorMessage + "</message></error>";
+        }
+        return Response.status(status).entity(message).type(mediaType).build();
```
qiangdavidliu: The else here assumes xml. This code will do funny stuff in other mediaTypes are supported in the future. How about if(json)elseif(xml)else(rawString) instead?

tbak: We should verify media type in isReadable method, but there is no proper check there. I will add it.

PR #656, opened 2015-09-28
1.x merge Refactorings/transport to features/di-friendly-server

tbak:


eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
```diff
@@ -418,6 +422,17 @@ public synchronized BackupRegistry get() {
             fetchRegistryFromBackup();
         }
 
+        // Configure new transport layer (candidate for injecting in the future)
+        EurekaHttpClient newEurekaHttpClient = null;
+        try {
+            String[] availZones = clientConfig.getAvailabilityZones(clientConfig.getRegion());
+            String myZone = InstanceInfo.getZone(availZones, instanceInfo);
+            newEurekaHttpClient = EurekaHttpClients.createStandardClient(clientConfig, applicationInfoManager, myZone);
```
qiangdavidliu: Can probably move the myZone initialization into createStandardClient, since it already takes clientConfig and appInfoManager as params.

eureka-client/src/main/java/com/netflix/discovery/shared/resolver/EurekaEndpoint.java
```diff
@@ -0,0 +1,132 @@
...
+package com.netflix.discovery.shared.resolver;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * @author Tomasz Bak
+ */
+public class EurekaEndpoint implements Comparable<EurekaEndpoint> {
```
qiangdavidliu: This Endpoint class seems pretty aws specific (references to zones etc). Can we interface this and then move the aws specific implementations of endpoint + resolver into a .aws package?

eureka-client/src/main/java/com/netflix/discovery/shared/resolver/LegacyClusterResolver.java
```diff
@@ -0,0 +1,115 @@
...
+
+            return newResolver;
+        }
+
+        private List<EurekaEndpoint> createEurekaEndpointsFromConfig() {
+            List<String> serviceUrls = EndpointUtils.getServiceUrlsFromConfig(clientConfig, myZone, clientConfig.shouldPreferSameZoneEureka());
+            List<EurekaEndpoint> endpoints = new ArrayList<>(serviceUrls.size());
+            for (String serviceUrl : serviceUrls) {
+                try {
+                    URI serviceURI = new URI(serviceUrl);
```
qiangdavidliu: Wouldn't this potentially resolve to and cache private hostnames in an aws environment?

eureka-client/src/main/java/com/netflix/discovery/shared/resolver/ReloadingClusterResolver.java
```diff
@@ -0,0 +1,141 @@
...
+        ClusterResolver delegate = delegateRef.get();
+        return delegate == null ? null : delegate.getRegion();
+    }
+
+    @Override
+    public List<EurekaEndpoint> getClusterEndpoints() {
+        long expiryTime = lastUpdateTime + currentReloadIntervalMs;
+        if (expiryTime <= System.currentTimeMillis()) {
+            try {
+                ClusterResolver newDelegate = reload();
```
qiangdavidliu: Can this block?

eureka-client/src/main/java/com/netflix/discovery/shared/transport/decorator/RebalancingEurekaHttpClient.java
```diff
@@ -0,0 +1,66 @@
...
+        long now = System.currentTimeMillis();
+        long delay = now - lastReconnectTimeStamp;
+        if (delay >= reconnectIntervalMs) {
+            lastReconnectTimeStamp = now;
+            TransportUtils.shutdown(eurekaHttpClientRef.getAndSet(null));
+        }
+
+        EurekaHttpClient eurekaHttpClient = eurekaHttpClientRef.get();
+        if (eurekaHttpClient == null) {
+            eurekaHttpClient = TransportUtils.getOrSetAnotherClient(eurekaHttpClientRef, clientFactory.create());
```
qiangdavidliu: the transportUtils method already keeps track of the client in an AtomicRef. There is no need to check again here.

eureka-client/src/main/java/com/netflix/discovery/shared/transport/decorator/MetricsCollectingEurekaHttpClient.java
```diff
@@ -0,0 +1,201 @@
...
+            case 1:
+                return Status.x100;
+            case 2:
+                return Status.x100;
+            case 3:
+                return Status.x100;
+            case 4:
+                return Status.x100;
+            case 5:
+                return Status.x100;
```
qiangdavidliu: Typos?

eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
tbak: I did not want to change DiscoveryClient constructor parameter list. I would revisit this when removing legacy code.

eureka-client/src/main/java/com/netflix/discovery/shared/resolver/EurekaEndpoint.java
tbak: Good point, I will refactor that.

eureka-client/src/main/java/com/netflix/discovery/shared/resolver/LegacyClusterResolver.java
tbak: I am not sure if I understand. Host part will be identical to what was provided in service URI, without further resolve. We could provide ClusterResolver decorator that would resolve any host name to its IP value, to avoid transient DNS issues.

eureka-client/src/main/java/com/netflix/discovery/shared/resolver/ReloadingClusterResolver.java
tbak: Yes. Certain resolvers take time to get their data (DNS), so this could block, and this would be in the request path. In the first iteration I have implemented reloading resolver with a background thread. Probably I should get it back.

eureka-client/src/main/java/com/netflix/discovery/shared/transport/decorator/RebalancingEurekaHttpClient.java
tbak: The logic in if branch should be called only if atomic ref contains null.

eureka-client/src/main/java/com/netflix/discovery/shared/transport/decorator/MetricsCollectingEurekaHttpClient.java
tbak: Copy paste.

eureka-client/src/main/java/com/netflix/discovery/shared/resolver/EurekaEndpoint.java
tbak: On a second thought, I am not sure about that. In ClusterResolver we have 'region' attribute. This is useful metadata, and it is convenient to carry it together with the ClusterResolver object, just like zone with EndpointResolver. In Eureka server we will have multiple copies of ClusterResolver, one for local, and a few for remote regions. We could provide AWS specific abstraction here as well, but in this case region name serves as an id.
Other cloud providers use similar terms (Google Compute Engine region/zone, Azure/RackSpace region). At least 'region' seems to be a common one.

PR #659, opened 2015-09-29
add route53 binding alternative to eip

aroger-r7:
with the eureka.bindingStrategy allow to switch between a eip binding or route53.

### DIFF
--- a/eureka-client/src/main/java/com/netflix/appinfo/AmazonInfo.java
+++ b/eureka-client/src/main/java/com/netflix/appinfo/AmazonInfo.java
@@ -68,6 +68,7 @@ public class AmazonInfo implements DataCenterInfo, UniqueIdentifier {
         instanceId("instance-id"),
         instanceType("instance-type"),
         localIpv4("local-ipv4"),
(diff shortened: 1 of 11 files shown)


qiangdavidliu:
@aroger-r7 thanks for the pull, this looks quite useful indeed. I apologies for this but we are close to checking in some large server side refactoring (see the feature/di-friendly-server branch) that affects this pull. Would you mind considering rebasing this change w.r.t. to that branch instead? 

The overall pull request should not need to change much, and in particular in the refactoring branch some similar was work done (EIP binding was moved out of the bootstrap class for example). The main class you may want to look at is [EIPManager.java](https://github.com/Netflix/eureka/blob/feature/di-friendly-server/eureka-core/src/main/java/com/netflix/eureka/aws/EIPManager.java). You may want to just provide a route53 parallel to this and at bootstrap/setup time, we can pick between the right ones to use.

aroger-r7:
no problem. will try do that asap

aroger-r7:
closed in favor of one on  di-friendly-server

PR #660, opened 2015-09-29
1.x Eviction mechanism update

tbak:
Experimental feature with no cancellation replication for evicted items.
Imposing additional eviction constraints, based on registry size not
received heartbeats.


qiangdavidliu:
Looks good.

PR #665, opened 2015-10-01
1.x Refactorings/transport

tbak:


eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
```diff
@@ -418,21 +420,23 @@ public synchronized BackupRegistry get() {
...
-        }
 
         // Configure new transport layer (candidate for injecting in the future)
+        EurekaHttpClientFactory newEurekaHttpClientFactory = null;
         EurekaHttpClient newEurekaHttpClient = null;
         try {
-            String[] availZones = clientConfig.getAvailabilityZones(clientConfig.getRegion());
-            String myZone = InstanceInfo.getZone(availZones, instanceInfo);
-            newEurekaHttpClient = EurekaHttpClients.createStandardClient(clientConfig, applicationInfoManager, myZone);
+            newEurekaHttpClientFactory = EurekaHttpClients.createStandardClientFactory(clientConfig, applicationInfoManager);
```
spencergibb: Why not inject the factory rather than having a static initializer?

tbak: Before that we need to get rid of legacy transport related code in DiscoveryClient. We want to keep it that way for some time, until we have enough confidence in the new transport implementation.

PR #678, opened 2015-10-09
Intelligent resolvers for read operations

qiangdavidliu:
(WIP) initial commit of intelligent resolvers for read operations
Some interface refactoring


eureka-client/src/main/java/com/netflix/discovery/endpoint/EndpointUtils.java
```diff
@@ -322,4 +323,38 @@ private static int getZoneOffset(String myZone, boolean preferSameZone, String[]
         return 0;
     }
 
+
+    public static EurekaEndpoint fromTargetUrl(final String targetUrl) {
```
tbak: This should be added as static constructor in EurekaEndpoint.

eureka-client/src/main/java/com/netflix/discovery/endpoint/EndpointUtils.java
```diff
@@ -322,4 +323,38 @@ private static int getZoneOffset(String myZone, boolean preferSameZone, String[]
...
+        return new EurekaEndpoint() {
+            @Override
+            public String getServiceUrl() {
+                return targetUrl;
+            }
+
+            @Override
+            public String getHostName() {
+                throw new UnsupportedOperationException("not implemented");
```
tbak: All these values can be taken out of URI.

eureka-client/src/main/java/com/netflix/discovery/shared/resolver/AsyncResolver.java
```diff
@@ -0,0 +1,65 @@
...
+    @Override
+    public String getRegion() {
+        return delegate.getRegion();
+    }
+
+    @Override
+    public List<T> getClusterEndpoints() {
+        if (!warmedUp) {
+            updateTask.run();
+            executorService.schedule(backgroundTask, 0, TimeUnit.SECONDS);
```
tbak: Shouldn't this be delayed?

eureka-client/src/main/java/com/netflix/discovery/shared/resolver/AsyncResolver.java
```diff
@@ -0,0 +1,65 @@
...
+            warmedUp = true;
+        }
+        return resultsRef.get();
+    }
+
+    private final Runnable updateTask = new Runnable() {
+        @Override
+        public void run() {
+            List<T> newList = delegate.getClusterEndpoints();
```
tbak: It is a good practice to wrap Runnable.run in try/catch, so we do not silently swallow exceptions.

eureka-client/src/main/java/com/netflix/discovery/shared/resolver/ClusterResolver.java
```diff
@@ -21,9 +21,9 @@
 /**
  * @author Tomasz Bak
  */
-public interface ClusterResolver {
+public interface ClusterResolver<T extends EurekaEndpoint> {
```
tbak: Do we want to have extended EurekaEndpoint interfaces, or just different implementations? In the latter case, I would drop the generic type parameter and use straight EurekaEndpoint as previously.

eureka-client/src/main/java/com/netflix/discovery/shared/resolver/aws/ApplicationsResolver.java
```diff
@@ -0,0 +1,64 @@
...
+                List<InstanceInfo> validInstanceInfos = application.getInstances();
+                for (InstanceInfo instanceInfo : validInstanceInfos) {
+                    if (instanceInfo.getStatus() == InstanceInfo.InstanceStatus.UP) {
+                        result.add(ResolverUtils.instanceInfoToEndpoint(clientConfig, instanceInfo));
+                    }
+                }
+            }
+        }
+
+        logger.debug("Retrieved endpoint list {}", result);
```
tbak: For stale data we will return an empty list, while the contract is that we should return previous value.

eureka-client/src/main/java/com/netflix/discovery/shared/resolver/aws/ApplicationsResolver.java
```diff
@@ -0,0 +1,64 @@
...
+        return result;
+
+    }
+
+    public interface ApplicationsSource {
+        /**
+         * @return the known set of Applications, or null if the data is beyond the stalenss threshold
+         */
+        Applications getApplications(int stalenessThreshold, TimeUnit timeUnit);
```
tbak: Why not returning List<InstanceInfo>? Applications/Application are mutable and inconvenient to use.

eureka-client/src/main/java/com/netflix/discovery/shared/resolver/aws/EurekaHttpResolver.java
```diff
@@ -0,0 +1,121 @@
...
+                              InstanceInfo myInstanceInfo,
+                              ClusterResolver<EurekaEndpoint> bootstrapResolver,
+                              String appName) {
+        this.clientConfig = clientConfig;
+        this.myInstanceInfo = myInstanceInfo;
+        this.appName = appName;
+        this.clientFactory = RetryableEurekaHttpClient.createFactory(
+                bootstrapResolver,
+                JerseyEurekaHttpClientFactory.create(clientConfig, myInstanceInfo, resolverIdentity),
+                new ServerStatusEvaluator() {
```
tbak: It could be added to ServerStatusEvaluators. Might be useful in other places.

eureka-client/src/main/java/com/netflix/discovery/shared/resolver/aws/EurekaHttpResolver.java
```diff
@@ -0,0 +1,121 @@
...
+    public EurekaHttpResolver(EurekaClientConfig clientConfig,
+                              InstanceInfo myInstanceInfo,
+                              ClusterResolver<EurekaEndpoint> bootstrapResolver,
+                              String appName) {
+        this.clientConfig = clientConfig;
+        this.myInstanceInfo = myInstanceInfo;
+        this.appName = appName;
+        this.clientFactory = RetryableEurekaHttpClient.createFactory(
+                bootstrapResolver,
+                JerseyEurekaHttpClientFactory.create(clientConfig, myInstanceInfo, resolverIdentity),
```
tbak: This factory creates Jersey connection pool, and must be shutdown. We could create a factory in getClusterEndpoints, and shut id down immediately after usage.

eureka-client/src/main/java/com/netflix/discovery/shared/transport/jersey/JerseyEurekaHttpClientFactory.java
```diff
@@ -64,9 +71,10 @@ public static EurekaHttpClientFactory create(EurekaClientConfig clientConfig,
                 .withMaxTotalConnections(clientConfig.getEurekaServerTotalConnections())
                 .withConnectionIdleTimeout(clientConfig.getEurekaConnectionIdleTimeoutSeconds())
                 .withEncoder(clientConfig.getEncoderName())
-                .withDecoder(clientConfig.getDecoderName(), clientConfig.getClientDataAccept());
+                .withDecoder(clientConfig.getDecoderName(), clientConfig.getClientDataAccept())
+                .withClientIdentity(clientIdentity);
 
-        if (isSecure && "true".equals(System.getProperty("com.netflix.eureka.shouldSSLConnectionsUseSystemSocketFactory"))) {
+        if ("true".equals(System.getProperty("com.netflix.eureka.shouldSSLConnectionsUseSystemSocketFactory"))) {
             clientBuilder.withClientName("DiscoveryClient-HTTPClient-System").withSystemSSLConfiguration();
```
tbak: Maybe we should only set base name, and append protocol specific parts in the builder? So all those `if`s can be replaced with single `clientBuilder.withClientName("DiscoveryClient")`?

eureka-client/src/main/java/com/netflix/discovery/shared/resolver/AsyncResolver.java
qiangdavidliu: Good catch

eureka-client/src/main/java/com/netflix/discovery/shared/resolver/ClusterResolver.java
qiangdavidliu: There is a difference that matter for some resolvers (e.g. zoneAffinity)

eureka-client/src/main/java/com/netflix/discovery/shared/resolver/aws/ApplicationsResolver.java
qiangdavidliu: The contract (as per the interface of ApplicationsSource) is to return no data if too stale.

eureka-client/src/main/java/com/netflix/discovery/shared/resolver/aws/ApplicationsResolver.java
qiangdavidliu: I wanted to keep the logic of which "name" the read cluster has inside the Resolver. If we were to return a list of InstanceInfos, it'll either be inefficient (scan) or bleed the logic out to the implementation of ApplicationsSource to understand and filter on the name.

eureka-client/src/main/java/com/netflix/discovery/shared/resolver/aws/EurekaHttpResolver.java
qiangdavidliu: Sure

eureka-client/src/main/java/com/netflix/discovery/shared/resolver/aws/EurekaHttpResolver.java
qiangdavidliu: Good catch.

eureka-client/src/main/java/com/netflix/discovery/shared/transport/jersey/JerseyEurekaHttpClientFactory.java
qiangdavidliu: Yeah let me play around with that.

eureka-client/src/main/java/com/netflix/discovery/endpoint/EndpointUtils.java
qiangdavidliu: I was actually going to move EndpointUtils to the transport package:)

eureka-client/src/main/java/com/netflix/discovery/endpoint/EndpointUtils.java
qiangdavidliu: True. I was lazy. Let me do the right thing.

PR #682, opened 2015-10-12
Splitting query and registration clients for experimental transport

qiangdavidliu:


eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
```diff
@@ -185,10 +191,14 @@
...
     private final ApacheHttpClient4 discoveryApacheClient;
-    private EurekaHttpClient eurekaHttpClient;
-    private EurekaHttpClientFactory eurekaHttpClientFactory;
     private EurekaJerseyClient discoveryJerseyClient;
 
+    private EurekaHttpClient registrationClient;
+    private EurekaHttpClientFactory registrationClientFactory;
+
+    private EurekaHttpClient queryClient;
+    private EurekaHttpClientFactory queryClientFactory;
```
tbak: Why not using `SplitEurekaHttpClient` like it was before?

qiangdavidliu: We still could. This change was made as the two (reg and query) clients need different resolvers that may internally be backed by threadpools, and given that we give the option of disabling reg and query, it makes sense to separate them so init can be skipped.

eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
```diff
@@ -354,12 +401,13 @@ public synchronized BackupRegistry get() {
...
                     new SynchronousQueue<Runnable>());  // use direct handoff
 
-            scheduleServerEndpointTask(zone);
-
             fetchRegistryGeneration = new AtomicLong(0);
 
             clientAccept = EurekaAccept.fromString(clientConfig.getClientDataAccept());
 
+            eurekaTransport = new EurekaTransport();
+            scheduleServerEndpointTask(eurekaTransport, zone);
```
tbak: As most clients will not use this feature, lets create this on demand. Ideally, if turned off we should release resources as well. What do you think?

eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
```diff
@@ -453,13 +501,38 @@ private void scheduleServerEndpointTask(String zone) {
                 clientConfig.getEurekaServiceUrlPollIntervalSeconds(),
                 clientConfig.getEurekaServiceUrlPollIntervalSeconds(), TimeUnit.SECONDS);
 
+        eurekaTransport.bootstrapResolver = EurekaHttpClients
+                .newBootstrapResolver(clientConfig, applicationInfoManager.getInfo());
+
+        eurekaTransport.transportClientFactory = EurekaHttpClients
+                .newTransportClientFactory(clientConfig, applicationInfoManager.getInfo());
+
+        if (clientConfig.shouldRegisterWithEureka()) {
```
tbak: If this method is called during initialization, the new transport must be enabled before the service is started?

eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
qiangdavidliu: All the new transport stuff are created lazily so will not use any resources unless triggered. We do need them to be pre-created though, as currently the option to use/no-use them are dynamic and not tied to a restart.

Having said that, the latest change to the AsyncResolver may have broken the lazy guarantee. Will fix that.

eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
qiangdavidliu: Correct. There should be no external dependencies (e.g. network calls) needed for the new transport init, they should all be lazy.

PR #683, opened 2015-10-15
Various updates for refactorings/transport

qiangdavidliu:
- update staleness monitor levels if registry fetch never worked
- Provide a mechanism to use new transport with bootstrap servers for query
- Tune down some logging

### DIFF
--- a/eureka-client/src/main/java/com/netflix/discovery/util/ThresholdLevelsMetric.java
+++ b/eureka-client/src/main/java/com/netflix/discovery/util/ThresholdLevelsMetric.java
@@ -78,6 +78,10 @@ public class ThresholdLevelsMetric {
         }
     }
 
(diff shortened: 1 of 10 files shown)


eureka-client/src/main/java/com/netflix/discovery/util/ThresholdLevelsMetric.java
```diff
@@ -78,6 +78,10 @@ public void update(long delayMs) {
         }
     }
 
+    /* visible for testing */ LongGauge[] getGauges() {
```
tbak: I do not see any usage of this method in this PR.

qiangdavidliu: Cruft code I forgot to delete. Will remove

PR #687, opened 2015-10-22
1.x new transport enhancements

qiangdavidliu:
- add config to define quarantine clear condition
- better metrics and logging
- adding jitter to sessionedClient's session duration

### DIFF
--- a/eureka-client/src/main/java/com/netflix/discovery/shared/transport/decorator/RetryableEurekaHttpClient.java
+++ b/eureka-client/src/main/java/com/netflix/discovery/shared/transport/decorator/RetryableEurekaHttpClient.java
@@ -161,10 +161,12 @@ public class RetryableEurekaHttpClient extends EurekaHttpClientDecorator {
 
         // If enough hosts are bad, we have no choice but start over again
         int threshold = candidateHosts.size() * transportConfig.getRetryableClientQuarantineRefreshPercentage() / 100;
-        if (quarantineSet.size() >= threshold) {
+        if (quarantineSet.isEmpty()) {
+            // no-op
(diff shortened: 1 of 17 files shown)


eureka-client/src/main/java/com/netflix/discovery/shared/transport/decorator/RetryableEurekaHttpClient.java
```diff
@@ -152,10 +159,14 @@ public void shutdown() {
         List<EurekaEndpoint> candidateHosts = clusterResolver.getClusterEndpoints();
         quarantineSet.retainAll(candidateHosts);
 
-        // If all hosts are bad, we have no choice but start over again
-        if (quarantineSet.size() == candidateHosts.size()) {
+        // If enough hosts are bad, we have no choice but start over again
+        int threshold = candidateHosts.size() * transportConfig.getRetryableClientQuarantineRefreshPercentage() / 100;
+        if (quarantineSet.isEmpty()) {
+            // no-op
+        } else if (quarantineSet.size() >= threshold) {
```
tbak: I am not sure if this is good approach. If first 50% servers are down, and threshold is 50%, the client will never be able to connect to any healthy instance, as quarantine set will be just before the first live instance reset to zero.

qiangdavidliu: Good point. However we should also also not have it so that if there is only a single good server left, then we send a thundering herd there.

YunaiV:
@qiangdavidliu 

hello, why add "adding jitter to sessionedClient's session duration" .

thank you!

PR #693, opened 2015-10-26
1.x Fix CLOSE_WAIT issue in new transport (use existing connection cleaner).

tbak:


eureka-client/src/main/java/com/netflix/discovery/shared/transport/jersey/JerseyEurekaHttpClientFactory.java
```diff
@@ -153,7 +162,7 @@ private JerseyEurekaHttpClientFactory buildLegacy() {
                     .withReadTimeout(readTimeout)
                     .withMaxConnectionsPerHost(maxConnectionsPerHost)
                     .withMaxTotalConnections(maxTotalConnections)
-                    .withConnectionIdleTimeout(connectionIdleTimeout)
+                    .withConnectionIdleTimeout((int) connectionIdleTimeout)
```
qiangdavidliu: If this is the majority use case and inside EurekaJerseyClientImpl we use the idle timeout as an int, why use long in the first place?

tbak: We use consistently long for any time related properties. Here we map to legacy that will be removed in the future.

qiangdavidliu:
ship it!

Issue #711, opened 2015-11-18
License violation suspicion - CDDL/GPLv2+CE reused code


vlna:
File `eureka-core/src/main/java/com/netflix/eureka/cluster/DynamicGZIPContentEncodingFilter.java` contains substantial part of http://grepcode.com/file/repo1.maven.org/maven2/com.sun.jersey/jersey-bundle/1.17.1/com/sun/jersey/api/client/filter/GZIPContentEncodingFilter.java

Original file is CDDL/GPLv2+CE dual-licensed and CDDL/GPLv2+CE is not possible to relicense to Apache 2.0.

I was not able to find this code under Apache 2.0 license and it looks like license violation.

Suspicious code was commited by @tbak in commit Netflix/eureka@1475b3b6b9cb6a9a31f01b287a58c1c699fc91d4

https://www.apache.org/legal/resolved.html#category-b describes proper handling of CDDL licensed code to be reused in Apache 2.0 licensed product.

tbak:
Thank you for spotting this. I will provide fresh implementation for this class, without basing it on the existing jersey code.

tbak:
Fixed in #713

PR #713, opened 2015-11-19
1.x Fix for issue https://github.com/Netflix/eureka/issues/711

tbak:


eureka-core/src/main/java/com/netflix/eureka/cluster/DynamicGZIPContentEncodingFilter.java
```diff
@@ -27,37 +29,29 @@ public DynamicGZIPContentEncodingFilter(EurekaServerConfig config) {
...
-
-            if (encodings.equals("gzip")) {
-                response.getHeaders().remove(HttpHeaders.CONTENT_ENCODING);
-                try {
-                    response.setEntityInputStream(new GZIPInputStream(response.getEntityInputStream()));
-                } catch (IOException ex) {
-                    throw new ClientHandlerException(ex);
-                }
-            }
+        String responseEncoding = response.getHeaders().getFirst(HttpHeaders.CONTENT_ENCODING);
```
qiangdavidliu: the gzip encoding could be specified as a later param, which would fail the getFirst here.

PR #716, opened 2015-12-08
1.x Client InstanceInfo default address resolution

qiangdavidliu:
Adding ability to define resolution order of the local InstanceInfo's "default" network address based on data available in AmazonInfo.

### DIFF
--- a/eureka-client/src/main/java/com/netflix/appinfo/InstanceInfo.java
+++ b/eureka-client/src/main/java/com/netflix/appinfo/InstanceInfo.java
@@ -861,6 +861,22 @@ public class InstanceInfo {
         return hostName;
     }
 
+    /**
+     * Return the default network address to connect to this instance.
+     *
+     * The address can either be a hostname or an ip and there is no guarantee which will be returned.
+     * Assume the address can change dynamically over time.
+     * If a usecase need more specific hostnames or ips, please use data from {@link #getDataCenterInfo()}.
+     *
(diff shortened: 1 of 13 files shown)


eureka-client/src/main/java/com/netflix/appinfo/InstanceInfo.java
```diff
@@ -861,6 +861,22 @@ public String getHostName() {
...
+     * The address can either be a hostname or an ip and there is no guarantee which will be returned.
+     * Assume the address can change dynamically over time.
+     * If a usecase need more specific hostnames or ips, please use data from {@link #getDataCenterInfo()}.
+     *
+     * For legacy reasons, the data backing this field is extracted from the hostname field
+     *
+     * @return either a hostname or an ipAddress
+     */
+    @JsonIgnore
+    public String getDefaultAddress() {
```
tbak: Why this need method is needed, if it just returns hostName?

qiangdavidliu: Ideally, it would be nice to get rid of the toplevel hostname/ipAddr fields and just have users access them from DataCenterInfo. At the toplevel, what's nice to have is a default "address" that can be either hostname or ip, as defined by the registering service itself. Unfortunately, for backwards compatibility, the hostname field is today the realistic data that's used for this. Introducing this new accessor for a cleaner client API.

tbak:
Looks good

PR #721, opened 2015-12-22
1.x Improve backwards compatibility with legacy eureka-client,

tbak:
that ignore the new InstanceInfo.instanceId field.

### DIFF
--- a/eureka-client/src/main/java/com/netflix/appinfo/InstanceInfo.java
+++ b/eureka-client/src/main/java/com/netflix/appinfo/InstanceInfo.java
@@ -50,7 +50,6 @@ import org.slf4j.LoggerFactory;
  * </p>
  *
  * @author Karthik Ranganathan, Greg Kim
- *
  */
 @ProvidedBy(EurekaConfigBasedInstanceInfoProvider.class)
 @Serializer("com.netflix.discovery.converters.EntityBodyConverter")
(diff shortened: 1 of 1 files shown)


eureka-client/src/main/java/com/netflix/appinfo/InstanceInfo.java
```diff
@@ -190,7 +189,21 @@ public InstanceInfo(
...
+
+        // Address the issue when new instanceId field is set, and Amazon metadata do not include instanceId.
+        // In this case legacy decoder will ignore instanceId value, and try to get the id from Amazon metadata.
+        // As it was not set, null will be returned which is returned further to the caller without check. This
+        // breaks registry deserialization process.
+        if (dataCenterInfo instanceof AmazonInfo && instanceId != null) {
+            AmazonInfo amazonInfo = (AmazonInfo) dataCenterInfo;
+            String effectiveId = amazonInfo.get(AmazonInfo.MetaDataKey.instanceId);
+            if (effectiveId == null) {
+                amazonInfo.getMetadata().put(AmazonInfo.MetaDataKey.instanceId.getName(), instanceId);
```
qiangdavidliu: This is not entirely correct. Since with the new changes registering entities are now able to specify their own (potentially custom and not related to EC2 instanceId) instanceId, we cannot make this assumption here and assume this put into the metadata map is correct.

tbak: I agree, and this patch should be removed in the future. Now it is the only way to prevent legacy client from crashing.

PR #730, opened 2016-01-11
1.x Refactor Jackson serializer, so it does not depend on internal Jackson code.

tbak:
This was source of incompatibility with Jackson versions higher than 2.5.


eureka-client/src/main/java/com/netflix/discovery/converters/jackson/serializer/ApplicationXmlDeserializer.java
```diff
@@ -0,0 +1,59 @@
...
+            jp.nextToken(); // to point to value
+            if ("name".equals(fieldName)) {
+                name = jp.getValueAsString();
+            } else if ("instance".equals(fieldName)) {
+                instances.add(jp.readValueAs(InstanceInfo.class));
+            } else {
+                throw new JsonMappingException("Unexpected field " + fieldName, jp.getCurrentLocation());
+            }
+        }
+        return new Application(name, instances);
```
qiangdavidliu: Check for  and handle null name if it exist?

tbak: JSON version is not doing it, so NPE will be thrown. If we want to check it for XML documents, we have two choices:
* throw an exception
* provide default application name (like 'default') to allow deserialization to proceed

Given that we do explicit data validation on the server side that checks for application name existence, I would not do any more checking here.

eureka-client/src/main/java/com/netflix/discovery/converters/jackson/builder/ApplicationsJacksonBuilder.java
```diff
@@ -0,0 +1,55 @@
...
+    private long version;
+    private String appsHashCode;
+
+    @JsonProperty("application")
+    public void withApplication(List<Application> applications) {
+        this.applications = applications;
+    }
+
+    @JsonAnySetter
+    public void with(String fieldName, Object value) {
```
qiangdavidliu: do we need to do null check for value here?

eureka-client/src/main/java/com/netflix/discovery/converters/jackson/serializer/ApplicationXmlDeserializer.java
qiangdavidliu: sounds good.

eureka-client/src/main/java/com/netflix/discovery/converters/jackson/builder/ApplicationsJacksonBuilder.java
tbak: I would assume Jackson is doing the necessary checking, but I will add extra guards.

eureka-client/src/main/java/com/netflix/discovery/converters/jackson/serializer/ApplicationXmlDeserializer.java
```diff
@@ -0,0 +1,59 @@
...
+    public Application deserialize(JsonParser jp, DeserializationContext ctxt) throws IOException, JsonProcessingException {
+        String name = null;
+        List<InstanceInfo> instances = new ArrayList<>();
+        while (jp.nextToken() == JsonToken.FIELD_NAME) {
+            String fieldName = jp.getCurrentName();
+            jp.nextToken(); // to point to value
+            if ("name".equals(fieldName)) {
+                name = jp.getValueAsString();
+            } else if ("instance".equals(fieldName)) {
+                instances.add(jp.readValueAs(InstanceInfo.class));
```
qiangdavidliu: it might be safer to try/catch here (and also in json version) to filter out any bad instance level data and not just exception out the entire application.

tbak: JSON version is not present, and it is fully based on Jackson annotations. XML version did not work correctly with some Jackson versions, hence own deserializer.

PR #733, opened 2016-01-15
Fix build error

jylin:
The putIfAbsent method is used on this variable which is not part of the Map interface.


qiangdavidliu:
@jylin putIfAbsent is part of the map API since java8.

jylin:
This is the only thing that prevents building with Java 7 though.

qiangdavidliu:
In that case, sure. However please do note that we are developing on java8 and may introduce other java8 only functionality in the future.

PR #734, opened 2016-01-19
blitz4j is no longer included in eureka server as part of 1.3.5...

dcaba:
... #714 . Default log4j properties still points to a class that is part of blitz4j, generating an exception at boot time.


tbak:
Thank you for spotting this.

dcaba:
you're welcome!

Issue #702, opened 2015-11-06
DiscoveryJerseyProvider: Error creating a serializer.  ClassNotFoundException: jackson


spencergibb:
```
2015-11-06 12:50:15.910 ERROR 81323 --- [et_localhost-13] c.n.d.provider.DiscoveryJerseyProvider   : Error creating a serializer.

java.lang.ClassNotFoundException: jackson
	at java.net.URLClassLoader$1.run(URLClassLoader.java:372) ~[na:1.8.0_05]
	at java.net.URLClassLoader$1.run(URLClassLoader.java:361) ~[na:1.8.0_05]
	at java.security.AccessController.doPrivileged(Native Method) ~[na:1.8.0_05]
	at java.net.URLClassLoader.findClass(URLClassLoader.java:360) ~[na:1.8.0_05]
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424) ~[na:1.8.0_05]
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:308) ~[na:1.8.0_05]
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357) ~[na:1.8.0_05]
	at java.lang.Class.forName0(Native Method) ~[na:1.8.0_05]
	at java.lang.Class.forName(Class.java:259) ~[na:1.8.0_05]
	at com.netflix.discovery.provider.DiscoveryJerseyProvider.getSerializer(DiscoveryJerseyProvider.java:207) [eureka-client-1.3.4.jar:1.3.4]
	at com.netflix.discovery.provider.DiscoveryJerseyProvider.readFrom(DiscoveryJerseyProvider.java:118) [eureka-client-1.3.4.jar:1.3.4]
	at com.sun.jersey.api.client.ClientResponse.getEntity(ClientResponse.java:634) [jersey-client-1.19.jar:1.19]
	at com.sun.jersey.api.client.ClientResponse.getEntity(ClientResponse.java:586) [jersey-client-1.19.jar:1.19]
	at com.netflix.eureka.transport.JerseyReplicationClient.submitBatchUpdates(JerseyReplicationClient.java:120) [eureka-core-1.3.4.jar:1.3.4]
	at com.netflix.eureka.cluster.ReplicationTaskProcessor.process(ReplicationTaskProcessor.java:71) [eureka-core-1.3.4.jar:1.3.4]
[...]

PR #739, opened 2016-02-04
Fix for issue #702 (handle charset provided in Content-Type header)

tbak:


eureka-client/src/main/java/com/netflix/discovery/provider/DiscoveryJerseyProvider.java
```diff
@@ -44,92 +43,81 @@
...
-        LOGGER.info("Using encoding codec {}", this.encoder.codecName());
-        LOGGER.info("Using decoding codec {}", this.decoder.codecName());
-    }
-
-    public EncoderWrapper getEncoder() {
-        return encoder;
-    }
+        // If jackson-dataformat-xml available, initialize XML codecs
+        if (isJacksonXmlOnClasspath()) {
+            this.xmlEncoder = CodecWrappers.getEncoder("JacksonXml");
```
qiangdavidliu: Use class instead of raw strings here

tbak: This might cause ClassNotFoundException if jackson-dataformat-xml is not on classpath.

qiangdavidliu:
ship it

PR #741, opened 2016-02-05
1.x Removing old client transport in favour of the (no longer) experimental transport

qiangdavidliu:


eureka-client/src/main/java/com/netflix/discovery/shared/transport/EurekaHttpClients.java
```diff
@@ -35,7 +37,10 @@
 import com.netflix.discovery.shared.transport.decorator.RedirectingEurekaHttpClient;
 import com.netflix.discovery.shared.transport.decorator.RetryableEurekaHttpClient;
 import com.netflix.discovery.shared.transport.decorator.ServerStatusEvaluators;
+import com.netflix.discovery.shared.transport.jersey.EurekaJerseyClient;
```
tbak: EurekaHttpClients should have no dependency on any transport implementation, as it is part of transport API.

eureka-client/src/main/java/com/netflix/discovery/shared/transport/jersey/JerseyEurekaHttpClientFactory.java
```diff
@@ -244,6 +276,14 @@ private void addFilters(ApacheHttpClient4 discoveryApacheClient) {
             String ip = myInstanceInfo == null ? null : myInstanceInfo.getIPAddr();
             AbstractEurekaIdentity identity = clientIdentity == null ? new EurekaClientIdentity(ip) : clientIdentity;
             discoveryApacheClient.addFilter(new EurekaIdentityHeaderFilter(identity));
+
+            if (additionalFilters != null) {
```
tbak: Do we have any integration test that creates a custom filter? If not, it may be worth to add it.

qiangdavidliu: added

PR #745, opened 2016-02-12
1.x Add composite strategy bootstrap resolver 

qiangdavidliu:
Two features for multi-account, multi vpc type support:
- an option for (all) resolvers to use hostnames or ipAddresses for resolution
- an option to bootstrap with a composite strategy, where the bootstrap resolver prefers to load data from local registry, else attempts to make a remote get vips call for write servers. The remote call will make a dns/config look up for target hosts.

### DIFF
--- a/eureka-client/src/main/java/com/netflix/discovery/shared/resolver/LegacyClusterResolver.java
+++ b/eureka-client/src/main/java/com/netflix/discovery/shared/resolver/LegacyClusterResolver.java
@@ -30,7 +30,8 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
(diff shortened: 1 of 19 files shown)


eureka-client/src/main/java/com/netflix/discovery/shared/resolver/LegacyClusterResolver.java
```diff
@@ -30,6 +30,8 @@
 import org.slf4j.LoggerFactory;
 
 /**
+ * @deprecated as of 2016-02-11. Will be deleted in an upcoming release
```
tbak: Lets add information about recommended way of resolving cluster information.

tbak:
Looks good

PR #746, opened 2016-02-12
Move back to java7 compatibility.

spencergibb:
Spring Cloud Netflix is still java7 compatible.

When using java7 to build, the following error occurs because servo is java8 only.

java.lang.UnsupportedClassVersionError: com/netflix/servo/monitor/Monitors
	at com.netflix.discovery.DiscoveryClient.<init>(DiscoveryClient.java:126)
	at com.netflix.discovery.BackUpRegistryTest.setUp(BackUpRegistryTest.java:71)

Spring Cloud Netflix still uses the java7 servo.

Building with java8 and targeting java7 still works and maintains the required compatibility.

@twicksell

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,11 @@ At Netflix, Eureka is used for the following purposes apart from playing a criti
 * For carrying other additional application specific metadata about services for various other reasons.
 
 
+Building
+----------
+The build requires java8 because of some required libraries that are java8 (servo), but the source and target compatibility are still set to 1.7.
+
+
 Support
 ----------
 [Eureka Google Group] (https://groups.google.com/forum/?fromgroups#!forum/eureka_netflix)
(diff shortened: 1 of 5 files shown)


spencergibb:
Is there anything I need to do to make sure it's built with java8?

spencergibb:
This was set in 78677c43800e715a3e861536ba84e5ff82510090 by @qiangdavidliu

qiangdavidliu:
@spencergibb looks good, I'll merge this in. Will spring cloud netflix be moving to java8 servo soon?

spencergibb:
@qiangdavidliu spring Cloud will start moving to java 8 with spring 5 and boot 2. Spring 5 is due next year, milestones starting this year. We would be OK staying on an the 1.3.x line if you move to java 8 in 1.4.x

qiangdavidliu:
Great, thanks for letting me know. There is no great need to enforce java8 for now, so we can try to make sure we can build against java7 unless anything comes up.

PR #749, opened 2016-02-16
New EurekaEventListener API

elandau:
Introduce a new EurekaEventListener API for getting EurekaClient state change notification.  

This API is cleaner than forcing the Netflix EventBus on users of EurekaClient.


eureka-client/src/main/java/com/netflix/discovery/CacheRefreshedEvent.java
```diff
@@ -6,6 +6,9 @@
  * 
  * @author brenuart
  */
-public class CacheRefreshedEvent extends DiscoveryEvent {
-
+public class CacheRefreshedEvent extends AbstractEurekaEvent {
+    @Override
+    public String toString() {
+        return "CacheRefreshedEvent[]";
```
tbak: We have timestamp attribute, so it should be printed here.

spencergibb:
This will be very nice! /cc @dsyer

eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
```diff
@@ -603,6 +624,11 @@ public void registerHealthCheck(HealthCheckHandler healthCheckHandler) {
         }
     }
 
+    @Override
+    public void registerEventListener(EurekaEventListener eventListener) {
```
tbak: As all listeners can be injected via constructor, there is no pressing need to have this method. If we need it, we should add unregisterEventListener as well.

eureka-client/src/main/java/com/netflix/discovery/CacheRefreshedEvent.java
```diff
@@ -6,6 +6,9 @@
  * 
  * @author brenuart
  */
-public class CacheRefreshedEvent extends DiscoveryEvent {
-
+public class CacheRefreshedEvent extends AbstractEurekaEvent {
```
tbak: Currently event listeners Depend on DiscoveryEvent <- (CacheRefreshedEvent, StatusChangeEvent) hierarchy. The new API is not backwards compatible, and make break some clients.

eureka-client/src/main/java/com/netflix/discovery/DiscoveryEvent.java
```diff
@@ -3,31 +3,11 @@
...
-		this.timestamp = System.currentTimeMillis();
-	}
-	
-	/**
-	 * Return the system time in milliseconds when the event happened.
-	 */
-	public final long getTimestamp() {
-		return this.timestamp;
-	}
+public abstract class DiscoveryEvent extends AbstractEurekaEvent {
```
tbak: By reverting this inheritance, we may fix the compatibility issue (CacheRefreshedEvent and StatusChangeEvent would still depend on DiscoveryEvent). However in this case why not stay with the original event model?

eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
tbak: On a second thought. Since the new event model is not backward compatible, we should stay with the old one, and depend on constructor injection only. We can introduce the new event model with new EurekaClient implementation, and make it parallel to the old one.

eureka-client/src/main/java/com/netflix/discovery/DiscoveryEvent.java
elandau: I'm using EurekaEvent since it is reference transitively through the EurekaClient API and is in the spirit of renaming 'discovery' to 'eureka'.

eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
```diff
@@ -1380,60 +1406,65 @@ public HealthCheckHandler getHealthCheckHandler() {
...
+                allAppsHashCodes.append(localRegionApps.get().getAppsHashCode());
+                allAppsHashCodes.append(", is fetching remote regions? ");
+                allAppsHashCodes.append(isFetchingRemoteRegionRegistries);
+                for (Map.Entry<String, Applications> entry : remoteRegionVsApps.entrySet()) {
+                    allAppsHashCodes.append(", Remote region: ");
+                    allAppsHashCodes.append(entry.getKey());
+                    allAppsHashCodes.append(" , apps hashcode: ");
+                    allAppsHashCodes.append(entry.getValue().getAppsHashCode());
                 }
-            } catch (Throwable th) {
```
tbak: As this is thread outermost loop try/catch we should catch Throwable exceptions. Otherwise the registry refresher may silently fail without us knowing about it.

eureka-client/src/main/java/com/netflix/discovery/DiscoveryEvent.java
tbak: I agree that this would be better, but it is not backward compatible change.
We plan to re-write EurekaClient, so we can leave the old event model with current DiscoveryClient implementation, and introduce new event model for the new implementation.
Alternatively we can introduce now the new model in parallel with the old one.

eureka-client/src/main/java/com/netflix/discovery/CacheRefreshedEvent.java
qiangdavidliu: Agree here. It is preferable to run with the two different event mechanisms side-by-side for now, until future client refactors.

elandau: I updated the PR in a way that should be backwards compatible.  I'd rather not have two different event mechanisms unless absolutely necessary, especially since the older one can be built on top of the new one.  Let me know if the change addresses your backwards compatibility concerns.

eureka-client/src/main/java/com/netflix/discovery/DiscoveryEvent.java
qiangdavidliu: If the aim of this is to move from discovery* to eureka* in terms of naming, then I would propose that it's fine to keep the old event model as there are functionally no difference. We already have .discovery. in our package names and it would introduce more api complexity with the renaming, than keeping with the current.

PR #752, opened 2016-02-23
Only append port to serviceUrl if port >= 0

spencergibb:
Fixes a problem communicating with an eureka server that is running on port 80 or 443 (such as the case in cloud foundry). serviceUrl looked like this http://eurekahost:-1

I can confirm this works on a local build.

See https://github.com/spring-cloud/spring-cloud-netflix/issues/840 for the original issue.


spencergibb:
FYI @tbak

tbak:
Thanks!

spencergibb:
@tbak thanks for the quick merge! Any idea on a release date?

tbak:
@spencergibb It is done.

spencergibb:
Sweet! @tbak you've been super helpful.

PR #754, opened 2016-02-24
Sleep during syncUp only if necessary

### LINKED ISSUE #753: PeerAwareInstanceRegistryImpl.syncUp blocks unecessarily
https://github.com/Netflix/eureka/blob/f2d9dbd886c2644d07ed5de6451c142a995af225/eureka-core/src/main/java/com/netflix/eureka/registry/PeerAwareInstanceRegistryImpl.java#L206

If its the last (or only) iteration of the loop, there's no need to sleep, because you won't be retrying again. To highlight the issue, if I wanted to implement a server that tries to fetch registry from peers once (with eureka.servier.registrySyncRetries=1), but in case their are no other peers alive, continue startup, we'd block for an additional 30 seconds before startup can complete.

william-tran:
Fixes #753


tbak:
Looks good. Thanks

PR #758, opened 2016-02-29
Self-preservation mode logged and announced by JMX

dcaba:
A new schedule is generated to announce if the instance is in self-preservation using servo

### DIFF
--- a/eureka-core/src/main/java/com/netflix/eureka/DefaultEurekaServerConfig.java
+++ b/eureka-core/src/main/java/com/netflix/eureka/DefaultEurekaServerConfig.java
@@ -218,6 +218,13 @@ public class DefaultEurekaServerConfig implements EurekaServerConfig {
     }
 
     @Override
+    public int getSelfPreservationStatusUpdateIntervalMs() {
+        return configInstance.getIntProperty(
+                namespace + "renewalSelfPreservationStatusUpdateIntervalMs",
+                (60 * 1000)).get();
+    }
+
+    @Override
     public double getRenewalPercentThreshold() {
         return configInstance.getDoubleProperty(
                 namespace + "renewalPercentThreshold", 0.85).get();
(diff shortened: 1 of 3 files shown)


tbak:
Servo can do probing automatically. Please, check how it is implemented in DiscoveryClient:
```java
    @com.netflix.servo.annotations.Monitor(name = METRIC_REGISTRATION_PREFIX + "lastSuccessfulHeartbeatTimePeriod",
            description = "How much time has passed from last successful heartbeat", type = DataSourceType.GAUGE)
    private long getLastSuccessfulHeartbeatTimePeriodInternal() {
        long delay = getLastSuccessfulHeartbeatTimePeriod();
        heartbeatStalenessMonitor.update(computeStalenessMonitorDelay(delay));
        return delay;
    }
```
Can you reimplement it using this approach?

dcaba:
Hi!

  I will assess this in the next days! The dedicated scheduler (plus polling) just for monitoring purposes looked a bit overkilling, but I thought may be the cleanest one until you presented this servo feature to me ;) . I will keep you posted,

     Daniel

dcaba:
Hi,

  I've been taking a look, and I guess we have two options:

* discard this PR. I've been browsing the current metrics you are exporing, and bearing in mind the method "isBelowRenewThreshold" is also there, the additional metric this PR incorporates adds small value.
* keep the suggested approach. Thanks to the additional schedule/thread, eureka will be logging if the instance is in self-preservation mode, or not. This could be interesting staff for admins. If we let servo to prove this, I understand the schedule (and the log generation) will be driven by external JMX polls, that does not look ideal. One point that can be interesting, though, is to log only transitions, instead of every renewalSelfPreservationStatusUpdateIntervalMs.

  Thanks!

tbak:
@dcaba I propose to close this PR. As you have stated we have roughly equivalent metric already, and if we would like to implement this anyway, we should use servo annotations.

Issue #756, opened 2016-02-26
Restore HTTP authentication support in eureka client


tbak:
Check https://github.com/Netflix/eureka/issues/755 for the context.

kennedyoliveira:
Hello! Any news on this topic?

tbak:
Sorry, we have been busy recently. Would you like to make a PR for this?

kennedyoliveira:
I would love to, but don't know how can i make it, if you give me some hints, i don't understand the eureka server, well didn't looked at the source through.

Anyway, i'll try, but would be very helpful if you could give me some advices.

Thank you!

YangGuang001:
i use jks credential, i create it on jvm (build 1.8.0_151-b12),  in the server i use server.keystore,  in the client i use server.cer.
but error message:
Caused by: java.io.IOException: Invalid keystore format
	at sun.security.provider.JavaKeyStore.engineLoad(JavaKeyStore.java:658) ~[na:1.8.0_111]
	at sun.security.provider.JavaKeyStore$JKS.engineLoad(JavaKeyStore.java:56) ~[na:1.8.0_111]
	at sun.security.provider.KeyStoreDelegator.engineLoad(KeyStoreDelegator.java:224) ~[na:1.8.0_111]
	at sun.security.provider.JavaKeyStore$DualFormatJKS.engineLoad(JavaKeyStore.java:70) ~[na:1.8.0_111]

server.keystore:
keytool -validity 365 -genkey -v -alias server -keyalg RSA -keystore server.keystore -dname "CN=127.0.0.1,OU=icesoft,O=icesoft,L=Haidian,ST=Beijing,c=cn" -storepass 123456 -keypass 123456 
server.cer:
keytool -export -v -alias server -keystore server.keystore -storepass 123456 -rfc -file server.cer

```java
@Configuration
public class EurekaJerseyClientAdapter
{
    @Bean
    public EurekaJerseyClient eurekaJerseyClient()
    {
        EurekaJerseyClientImpl.EurekaJerseyClientBuilder builder = new
                EurekaJerseyClientImpl.EurekaJerseyClientBuilder()
                .withTrustStoreFile("server.cer","123456");
        return builder.build();
    }
}
```

PR #765, opened 2016-03-15
Fix issue #756 (restore HTTP authentication support)

tbak:


tbak:
@kennedyoliveira Here is the fix restoring basic authentication. Most of the code there is the test code, to make sure we do not miss this feature in the future. We do not use HTTP basic authentication as it is inherently insecure.

spencergibb:
Thanks @tbak, we have folks that use it in conjunction with SSL.  Looks good!

kennedyoliveira:
@tbak, thank you very much! I know it's not safe, but i was planing to use with SSL as @spencergibb, there is any alternative to this method for eureka?

spencergibb:
There's some SSL certificate security, but other than that...

tbak:
@kennedyoliveira We are using client side certificates, to authenticate both sides of the connection.

qiangdavidliu:
:+1:

kennedyoliveira:
@tbak, sounds interesting, how is that?

tbak:
@kennedyoliveira Sorry, I have to take it back about the client side certificate validation. There are however some ways to do that (via SSL system properties or creating custom EurekaJerseyClient).

kennedyoliveira:
@tbak, hmm i will search about that, i don't know much about SSL, just the basic, anyway, thank you for fixing it! If you have any source of documentation/books/ anything about this case and could tell me, i appreciate :D

PR #767, opened 2016-03-23
1.x Add test showing issue with OUT_OF_SERVICE status registration.

tbak:


tbak:
During internal discussion we decided to handle this at REST layer, by not allowing registrations with OUT_OF_SERVICE status set.

PR #773, opened 2016-04-06
1.x EIPManager: if instance is not associated with one EIP, do not try to unbind

hieurl:
Instead, throw exception
Reason: DescribeAddressRequest with null public ip returns list of all public address
If instance is not associated with one eip, Eureka will dissociate first ip on list which does not belong to the instance

### DIFF
--- a/eureka-core/src/main/java/com/netflix/eureka/aws/EIPManager.java
+++ b/eureka-core/src/main/java/com/netflix/eureka/aws/EIPManager.java
@@ -260,7 +260,8 @@ public class EIPManager implements AwsBinder {
             myPublicIP = ((AmazonInfo) myInfo.getDataCenterInfo())
                     .get(MetaDataKey.publicIpv4);
             if (myPublicIP == null) {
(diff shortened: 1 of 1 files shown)


hieurl:
Quick check using python awscli

>aws ec2 describe-addresses --public-ips | jq '.Addresses[0]'
> {
>   "InstanceId": "i-xxxxx",
>   "PublicIp": "xxx.xxx.xxx.xxx",
>   "Domain": "standard"
> }

qiangdavidliu:
@hieurl thanks for the pull. Looking at the change, unbindEIP() is currently really used at shutdown time, so a more optimized behaviour when unbindEIP() is called when no EIP is bound, is to perhaps log.info() and then just return from the function, instead of throwing an RuntimeException. What do you think?

hieurl:
@qiangdavidliu seem reasonable to me. I'm going to push a commit.

qiangdavidliu:
@hieurl thanks for the contribution.

PR #784, opened 2016-04-28
Support healthy status on server StatusInfo

mattnelson:
When one of the replicas is down the eureka server status should reflect that in the healthy flag on `StatusInfo` and `/eureka/v2/status` endpoint.

### DIFF
--- a/eureka-core/src/main/java/com/netflix/eureka/EurekaServerConfig.java
+++ b/eureka-core/src/main/java/com/netflix/eureka/EurekaServerConfig.java
@@ -411,10 +411,18 @@ public interface EurekaServerConfig {
     
     /**
      * Get the minimum number of available peer replication instances
+     * for this instance to be considered healthy. The design of eureka allows
+     * for an instance to continue operating with zero peers, but that would not
+     * be ideal.
+     * <p>
+     * The default value of -1 is interpreted as a marker to not compare
+     * the number of replicas. This would be done to either disable this check
+     * or to run eureka in a single node configuration.
      * 
      * @return minimum number of available peer replication instances
+     *         for this instance to be considered healthy.
      */
-    int getMinNumberOfAvailablePeers();
+    int getHealthStatusMinNumberOfAvailablePeers();
 
     /**
      * Get the time in milliseconds to try to replicate before dropping
--- a/eureka-core/src/main/java/com/netflix/eureka/EurekaServerConfig.java
+++ b/eureka-core/src/main/java/com/netflix/eureka/EurekaServerConfig.java
@@ -408,6 +408,13 @@ public interface EurekaServerConfig {
      * @return maximum number of threads to be used for replication.
      */
     int getMaxThreadsForPeerReplication();
+    
+    /**
+     * Get the minimum number of available peer replication instances
+     * 
+     * @return minimum number of available peer replication instances
+     */
+    int getMinNumberOfAvailablePeers();
 
     /**
      * Get the time in milliseconds to try to replicate before dropping
--- a/eureka-core/src/main/java/com/netflix/eureka/util/StatusUtil.java
+++ b/eureka-core/src/main/java/com/netflix/eureka/util/StatusUtil.java
@@ -54,7 +54,7 @@ public class StatusUtil {
         builder.add("unavailable-replicas", downReplicas.toString());
         
         // Only set the healthy flag if a threshold has been configured.
-        if (peerEurekaNodes.getMinNumberOfAvailablePeers() > 0) {
+        if (peerEurekaNodes.getMinNumberOfAvailablePeers() > -1) {
             builder.isHealthy(upReplicasCount >= peerEurekaNodes.getMinNumberOfAvailablePeers());
         }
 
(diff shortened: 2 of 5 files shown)


qiangdavidliu:
@mattnelson sorry about missing this PR, not sure how I missed this.

The changes in general looks fine, though taking a look at the information exposed by StatusInfo (which is code that has been moved around a bunch of times now and is pretty old), the contract for isHealthy seems poorly defined. The isHealthy field should really only be applied to the local server itself, as at a cluster level, eureka is can still be considered "healthy" even with some down replicas.

mattnelson:
The cluster health would depend on the deployment configuration? Using the EIP/DNS options would provide multiple levels of resiliency. Using the `eureka.serviceUrl` property could have huge impacts with a down replica.

I think the `isHealthy` contract could be defined with the introduction of a new property which is the threshold of down replicas in which is it unhealthy due to the inability to replicate confidently.

qiangdavidliu:
If you are accessing eureka (server) through the provided eureka-client, then the theoretical threshold for DOWN replicas is pretty much n-1. There is a some load balancing built into the provided client such that they can handle parts of the server cluster being unavailable.

The replication protocol within eureka is a broadcast model, and hence as long as you have a single server standing, the cluster is "ok" (provided that this single server can handle the incoming requests of the world of course).

qiangdavidliu:
Hi @mattnelson , eureka is designed to be able to operate fully even when replicas are missing (as long as not all replicas are missing, and of course there are enough replicas to satisfy the data output bandwidth requirements). Replication is not tied to the number of replicas available so having downed replicas does not affect replication at a local level between the healthy replicas.

I do agree fully with you that it would be very useful to have, from the status endpoint, a view of the cluster's health. However the current PR results in an "unhealthy" status for a given server, if peers of this server are unhealthy. This would be incorrect information per our discussions in this thread.

How would you like to proceed? You are more than welcome to update this PR to expand on the health status to expose cluster level health if you would like to define a cluster level down threshold. Alternatively we can close this PR for now, and start a new PR for that.

Thanks.

mattnelson:
For the sake of not losing the context of this discussion. I will make the changes and update the PR. I will try to get the proposed changes done in the next couple of days.

mattnelson:
Finally got around to making these changes in bf6aa56945abbb3c067f6c4ae492fbe5f95bf489

* Went with an integer for the number of required up replicas since I think the common uses cases would be a relatively low number of replicas.
* Couldn't come up with a good name for the property so looking for suggestions there.
* Handles the default property value of 0 to remain passive with the current functionality of an unset healthy flag.

eureka-core/src/main/java/com/netflix/eureka/EurekaServerConfig.java
```diff
@@ -408,6 +408,13 @@
      * @return maximum number of threads to be used for replication.
      */
     int getMaxThreadsForPeerReplication();
+    
+    /**
+     * Get the minimum number of available peer replication instances
+     * 
+     * @return minimum number of available peer replication instances
+     */
+    int getMinNumberOfAvailablePeers();
```
qiangdavidliu: Looks like this config is designed for the statusPage only. Perhaps rename the method to reflect that? As it stands, it (name + comment) feels more like something that is system critical.

eureka-core/src/main/java/com/netflix/eureka/util/StatusUtil.java
```diff
@@ -50,11 +52,16 @@ public StatusInfo getStatusInfo() {
         builder.add("registered-replicas", replicaHostNames.toString());
         builder.add("available-replicas", upReplicas.toString());
         builder.add("unavailable-replicas", downReplicas.toString());
+        
+        // Only set the healthy flag if a threshold has been configured.
+        if (peerEurekaNodes.getMinNumberOfAvailablePeers() > 0) {
```
qiangdavidliu: What about a not configured default of -1? Technically you can run with a single eureka server (obv not recommended), in which case you would legitimately have 0 peers.

qiangdavidliu:
@mattnelson sorry didn't see this PR getting updated, some minor quibbles but LGTM.

mattnelson:
@qiangdavidliu Addressed PR comments on bc336a27b5cb6ef45b4499c89f4e25d28ee07b9f

mattnelson:
Not sure what caused the failure. I don't seem to be able to see the output of the travis build. I built the project locally several times to see if I could reproduce, but all the builds passed.

qiangdavidliu:
@mattnelson thanks, the failures are not your fault. Ever since we migrated to travisCI we've had these intermittent failures. Current theory is that the limited CPU that the travis containers are configured for are not enough for some of the multi-treading tests. I'll trigger some reruns for you.

qiangdavidliu:
@mattnelson the failures are actually due to one of the tests added:
```
com.netflix.eureka.util.StatusUtilTest > testGetStatusInfoHealthy FAILED
    java.lang.AssertionError at StatusUtilTest.java:28
```

mattnelson:
Figured out the failure reason. This branch wasn't rebased with the changes I introduced on #831.

I cherry picked that commit over and fixed the tests.

qiangdavidliu:
Looks good, merged. Thanks.

PR #797, opened 2016-05-18
Add fail fast option to AmazonInfo init

qiangdavidliu:
Add fail fast option to AmazonInfo init (esp at bootstrap time) so that running with VPN will not unncecessarily cause long hang times waiting on each key's ReadTimeout.


eureka-client/src/main/java/com/netflix/appinfo/AmazonInfo.java
```diff
@@ -184,15 +185,28 @@ public AmazonInfo build() {
             return result;
         }
 
+        public AmazonInfo autoBuild(String namespace) {
+            return this.autoBuild(namespace, false);
```
twicksell: Should consider making this true by default.

PR #800, opened 2016-05-27
Handle on demand update within InstanceInfoReplicator initial delay

mattnelson:
When performing an on demand status update within the InstanceInfoReplicator start's initial delay the originally scheduled task is not cancelled which results in two scheduled tasks being executed for the life of the process.


qiangdavidliu:
@mattnelson thanks for the fix!

PR #805, opened 2016-06-13
! Fix erroneous connectionIdleTimeout value

vseguin:
This commit is related to issue [https://github.com/spring-cloud/spring-cloud-netflix/issues/729](https://github.com/spring-cloud/spring-cloud-netflix/issues/729).

When overriding the connectionIdleTimeout value, it was erroneously using a seconds value, although default value is in milliseconds.


qiangdavidliu:
@vseguin thank you for the fix!

Issue #803, opened 2016-06-06
Eureka Peer Server throwing java.lang.NullPointerException: null at com.netflix.eureka.registry.AbstractInstanceRegistry.renew


codependent:
I am using Spring Cloud Netflix to run two Eureka (1.4.6) servers with peer awareness and found this issue after their last API upgrade:

```
java.lang.NullPointerException: null
    at com.netflix.eureka.registry.AbstractInstanceRegistry.renew(AbstractInstanceRegistry.java:379) ~[eureka-core-1.4.6.jar:1.4.6]
    at com.netflix.eureka.registry.PeerAwareInstanceRegistryImpl.renew(PeerAwareInstanceRegistryImpl.java:401) ~[eureka-core-1.4.6.jar:1.4.6]
    at org.springframework.cloud.netflix.eureka.server.InstanceRegistry.renew(InstanceRegistry.java:127) ~[spring-cloud-netflix-eureka-server-1.1.0.RELEASE.jar:1.1.0.RELEASE]
    at com.netflix.eureka.resources.InstanceResource.renewLease(InstanceResource.java:112) ~[eureka-core-1.4.6.jar:1.4.6]
    at com.netflix.eureka.resources.PeerReplicationResource.handleHeartbeat(PeerReplicationResource.java:148) ~[eureka-core-1.4.6.jar:1.4.6]
    at com.netflix.eureka.resources.PeerReplicationResource.dispatch(PeerReplicationResource.java:113) ~[eureka-core-1.4.6.jar:1.4.6]
    at com.netflix.eureka.resources.PeerReplicationResource.batchReplication(PeerReplicationResource.java:85) ~[eureka-core-1.4.6.jar:1.4.6]
    at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[na:1.8.0_74]
    at sun.reflect.NativeMethodAccessorImpl.invoke(Unknown Source) ~[na:1.8.0_74]
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(Unknown Source) ~[na:1.8.0_74]
    at java.lang.reflect.Method.invoke(Unknown Source) ~[na:1.8.0_74]
```

[Here are the details](https://github.com/spring-cloud/spring-cloud-netflix/issues/1065) to reproduce the problem.

spencergibb:
I sent him here as the NPE is in pure eureka code

qiangdavidliu:
[...]

PR #806, opened 2016-06-14
Fix (harmless) NPE at server startup due to init order on a metrics counter

qiangdavidliu:
#803

### DIFF
--- a/eureka-core/src/main/java/com/netflix/eureka/util/MeasuredRate.java
+++ b/eureka-core/src/main/java/com/netflix/eureka/util/MeasuredRate.java
@@ -31,30 +31,45 @@ public class MeasuredRate {
     private static final Logger logger = LoggerFactory.getLogger(MeasuredRate.class);
     private final AtomicLong lastBucket = new AtomicLong(0);
     private final AtomicLong currentBucket = new AtomicLong(0);
+
+    private final long sampleInterval;
     private final Timer timer;
 
-    // private ctor for NO_OP instance
-    private MeasuredRate() {
-        this.timer = null;
-    }
+    private volatile boolean isActive;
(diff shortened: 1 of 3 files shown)


tcellucci:
hey David, I can see that this change will fix the NPE problem, but wanted to suggest that an underlying problem is still present.  It looks like the root issue is that MeasuredRate doesn't manage its resources (Timer, TimerTask) in a way that's compatible with AbstractInstanceRegistry.  Since MeasuredRate starts its internal timer in construction, AbstractInstanceRegistry can't instantiate it until 'postInit' and that led to the NPE you fixed.  
If MeasuredRate had lifecycle methods (start, stop) to handle timer activity, then it could be constructed immediately by AbstractInstanceRegistry and avoid NPEs altogether.  Additionally the 'stop' method would enable AbstractInstanceRegistry to free up its resources on shutdown().

eureka-core/src/main/java/com/netflix/eureka/util/MeasuredRate.java
```diff
@@ -31,30 +31,45 @@
...
     private final AtomicLong currentBucket = new AtomicLong(0);
+
+    private final long sampleInterval;
     private final Timer timer;
 
-    // private ctor for NO_OP instance
-    private MeasuredRate() {
-        this.timer = null;
-    }
+    private volatile boolean isActive;
```
tcellucci: volatile is not necessary here, since both start and stop methods are synchronized and isStarted is used only in these methods.   (synchronized methods force a sync with object's entire state prior to execution)

tcellucci:
looks good.  I had one comment on use of 'volatile' but not a blocker.

Issue #818, opened 2016-07-19
IndexOutOfBounds Exception thrown repeatedly when Eureka has no EIPs containing "ec2-"


robertnosburn:
The EIPManager has a running task that looks to bind Eureka to EIPs.  In our case we are writing an application which is hosted in AWS but will not be publicly available and will be accessible ONLY from our corporate network.

We have an AWS Direct Connect to our corporate environment.  My EC2 container running my docker images, including Eureka, does not have a public nor elastic IP address.  In this case the only "ec2Urls" value is "http://localhost:8761".

EIPManager.getEIPsFromServiceUrls(List<String> ec2Urls) (line:346-349) does not handle this case and line 347 will throw:
java.lang.StringIndexOutOfBoundsException: String index out of range: -4
        at java.lang.String.substring(String.java:1967) ~[na:1.8.0_91]
        at com.netflix.eureka.aws.EIPManager.getEIPsFromServiceUrls(EIPManager.java:342) ~[eureka-core-1.4.6.jar!/:1.4.6]
        at com.netflix.eureka.aws.EIPManager.getEIPsForZoneFromConfig(EIPManager.java:322) ~[eureka-core-1.4.6.jar!/:1.4.6]
        at com.netflix.eureka.aws.EIPManager.getCandidateEIPs(EIPManager.java:304) ~[eureka-core-1.4.6.jar!/:1.4.6]
        at com.netflix.eureka.aws.EIPManager.isEIPBound(EIPManager.java:157) ~[eureka-core-1.4.6.jar!/:1.4.6]
        at com.netflix.eureka.aws.EIPManager$EIPBindingTask.run(EIPManager.java:412) ~[eureka-core-1.4.6.jar!/:1.4.6]
        at java.util.TimerThread.mainLoop(Timer.java:555) [na:1.8.0_91]
        at java.util.TimerThread.run(Timer.java:505) [na:1.8.0_91]

I propose a check to ensure that the index is greater than -1 before parsing the substrings.

See #819  as a potential fix.

Thanks,

Robert

qiangdavidliu:
Hi @robertnosburn thank you for the PR. Looking at your usecase, it seems like what you would prefer is that the EIPBinder not run at all?

Eureka servers need to have fixed IPs, and EIP is one mechanism that we use to guarantee that in a cloud environment. However if you are able to guarantee that via some other mechanism, then possibly a better way to disable the errors from the EIPBinder is to not initialize the EIPBinder at all.

[...]

PR #819, opened 2016-07-19
Check the sub-string index value before trying to sub-string on it.

robertnosburn:
See issue #818 -

In a private application with no elastic IPs exceptions will be thrown when there are no URLs containing "ec2-".  With this modification, the exception will not be thrown and the rest of the process will simply not bind to any EIPs.


qiangdavidliu:
@robertnosburn thanks for the PR.

Issue #600, opened 2015-07-31
Eureka Client dependency update to Jersey 2.x


cforce:
We have to add the spring-boot-starter-jersey to put Jersyey 2.14 on the classpath. The mix of both versions lead to strange NoSuchMethodErrors during application startup! The only solution seems to be to use JAX-RX 1.x (Jersey 1) . This is maybe related to https://github.com/Netflix/eureka/issues/376

qiangdavidliu:
@cforce eureka-client depends on jersey 1.x only and we have not tested (or guarantee) compatibility with jersey 2.x. We also do not have any plans on upgrading to jersey 2.x in the short term.

foo4u:
@qiangdavidliu can we shade the dependency on Jersey 1.x? Causing issues since most other frameworks have moved on to JAX-RS 2.0.

tbak:
We implement a new transport that will allow one to plug-in different HTTP client implementations. It is still on branch, but we will merge this to master probably next week. Once it is done, you can exclude jersey 1.x, and provide equivalent jersey 2.x implementation
Check: https://github.com/Netflix/eureka/tree/refactorings/transport/eureka-client/src/main/java/com/netflix/discovery/shared/transport

If we have time, we might provide a new module (eureka-client-jersey2), that would do exactly this.

efenderbosch:
Would that make it possible to use something other than Jersey? It is just required to be a JAX-RS implementation?

tbak:
For eureka-client, yes this could be anything that implements EurekaHttpClient API. But on the server side we depend heavily on JAX-RS.

cforce:
How its solved now? Do i still need to pimp like this?
 <plugins>
            <plugin>
                <groupId>com.spotify</groupId>
                <artifactId>docker-maven-plugin</artifactId>
            </plugin>
            <plugin>
                <groupId>org.springframework.boot</groupId>
                <artifactId>spring-boot-maven-plugin</artifactId>
                
                <configuration>
                    <requiresUnpack>
                        <dependency>
                            <groupId>com.netflix.eureka</groupId>
                            <artifactId>eureka-core</artifactId>
                        </dependency>
                        <dependency>
                            <groupId>com.netflix.eureka</groupId>
                            <artifactId>eureka-client</artifactId>
                        </dependency>
                    </requiresUnpack>
                </configuration>
            </plugin>
        </plugins>

or 

<dependency>
        <groupId>org.springframework.cloud</groupId>
[...]

PR #821, opened 2016-07-20
Support jersey2 for client/server

mattnelson:
Initial support for jersey2 for #600 

Deployed these changes to our local snapshot repository. I've been running our builds with the snapshot for the last week. I believe the only non-passive change for existing jersey1 consumers is the relocation of `DiscoveryClientOptionalArgs` from an inner class of `DiscoveryClient` this was because the optional args has a method to supply additional filters which the interface was changed with jaxrs2.

### client
Changed `DiscoveryClient` to accept a `TransportClientFactories` on the optional args class to support supplying the jersey2 transport client factory.

```java
Jersey2DiscoveryClientOptionalArgs args = new Jersey2DiscoveryClientOptionalArgs();
args.setTransportClientFactories(Jersey2TransportClientFactories.getInstance());
DiscoveryManager.getInstance().initComponent(new MyDataCenterInstanceConfig(), new DefaultEurekaClientConfig(), args);
```

### server
Added `EurekaServerContextBinder` to use the hk2 injectors for `EurekaServerContext`. The `GuiceFilter` is no longer required to be on the filter chain. I did not create a `eureka-server-jersey2` module since I am not deploying eureka-server as a war. Added `Jersey2EurekaBootStrap` which uses the client/server jersey2 classes for peer replication.

```java
// Expects that you have already initialized the discovery client
new Jersey2EurekaBootStrap(DiscoveryManager.getInstance().getDiscoveryClient()
```


qiangdavidliu:
@mattnelson thank you for the PR! Much appreciated. I'll take a look at this as soon as I can.

eureka-core/src/main/java/com/netflix/eureka/EurekaBootStrap.java
```diff
@@ -82,6 +81,12 @@
 
     protected volatile EurekaServerContext serverContext;
     protected volatile AwsBinder awsBinder;
+    
+    private DiscoveryClient discoveryClient;
```
qiangdavidliu: I think it is reasonable to add getEurekaClientConfig() to the EurekaClient interface, which will then allow the reference here to be to the EurekaClient interface instead of the DiscoveryClient implementation.

eureka-core/src/main/java/com/netflix/eureka/EurekaBootStrap.java
```diff
@@ -82,6 +81,12 @@
...
+    
+    protected void setDiscoveryClient(DiscoveryClient discoveryClient) {
```
qiangdavidliu: Rather than providing overrideable ways to set the client, maybe we can add the necessary constructors to EurekaBootstrap:
```
public EurekaBootstrap(ApplicationInfoManager infoManager, EurekaClient client) {
    this... = ...
}

public EurekaBootstrap() {
    this(null, null);
}

File: eureka-client-jersey2/src/main/java/com/netflix/discovery/shared/transport/jersey2/Jersey2TransportClientFactories.java [hunk 3]
```diff
@@ -0,0 +1,57 @@
...
+
+    @Override
+    public TransportClientFactory newTransportClientFactory(final EurekaClientConfig clientConfig,
+                                                                   final Collection<ClientRequestFilter> additionalFilters,
+                                                                   final InstanceInfo myInstanceInfo) {
+        final TransportClientFactory jerseyFactory = Jersey2ApplicationClientFactory.create(
+                clientConfig,
+                additionalFilters,
+                myInstanceInfo,
+                new EurekaClientIdentity(myInstanceInfo.getIPAddr())
```
qiangdavidliu: It might be worthwhile creating a jersey2 version of the EurekaClientIdentity here such that they identify as jersey2 variants of the eureka-client.

eureka-client-jersey2/src/main/java/com/netflix/discovery/shared/transport/jersey2/EurekaJersey2ClientImpl.java
```diff
@@ -0,0 +1,354 @@
...
+        return apacheHttpClient;
+    }
+
+    /**
+     * Clean up resources.
+     */
+    @Override
+    public void destroyResources() {
+        if (eurekaConnCleaner != null) {
+            eurekaConnCleaner.shutdown();
```
qiangdavidliu: In the existing code the connCleaner is called one last time before it is shutdown. Do you want to replicate that here too?

eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
```diff
@@ -463,19 +404,26 @@ public synchronized BackupRegistry get() {
...
+        Collection<?> additionalFilters = args == null
+                ? Collections.emptyList()
                 : args.additionalFilters;
 
         EurekaJerseyClient providedJerseyClient = args == null
                 ? null
                 : args.eurekaJerseyClient;
-
+        
+        TransportClientFactories transportClientFactories = TransportClientFactories.INSTANCE;
```
qiangdavidliu: Can we try another way to do this here instead of having a default static instance reference of a child class in the TransportClientFactories interface? One suggestion could be:
1. define the interface for TransportClientFactories as is, but with no default static instance
2. have a TransportClientFactoriesProvider, e.g. 
```java
public class TransportClientFactoriesProvider implements Provider<TransportClientFactories<?>> {

    private volatile TransportClientFactories<?> transportClientFactories;

    public TransportClientFactoriesProvider(TransportClientFactories<?> transportClientFactories) {
        this.transportClientFactories = transportClientFactories;
    }

    @Override
    public synchronized TransportClientFactories get() {
        if (transportClientFactories == null) {
            transportClientFactories = new Jersey1TransportClientFactories();
        }

        return transportClientFactories;
    }
}
```

3. in DiscoveryClient, initialize this provider with args.getTransportClientFactories(), then just get the actual impl out here in the same way.

qiangdavidliu:
@mattnelson are you running both jersey2 client and server in your set up, or just the client?

qiangdavidliu:
@spencergibb you'll be interested in this PR too.

mattnelson:
@qiangdavidliu 
> are you running both jersey2 client and server in your set up, or just the client?

I am running both client and server. The environment I'm using has 4 clients and 1 server (I'll work on getting another server integrated to verify cluster replication) that are able to register and use ribbon with `DiscoveryEnabledNIWSServerList`.

qiangdavidliu:
@mattnelson much appreciated. It would be good to have some confirmation that replication is also unaffected once you have >1 server running.

mattnelson:
Addressed comments on 7dbea2d1e1b5d3d2e816593f8542a27107cdc10d

mattnelson:
@qiangdavidliu Finally had a chance to validate replication.

Had 36 clients register with a cluster of 3 eureka servers. Then had an additional 2 eureka servers that were only configured with the other 3 eureka servers to ensure that the apps they got were only through replication. Took a while because I had to trace down a bug in `PeerEurekaNodes. isThisMyUrl`[1] and `EurekaTransportConfig.applicationsResolverUseIp`[2]. The virtualization environment I was using only supports IP resolution. I'll get a PR out for those changes as it will be a blocker for me to validate a release candidate of these changes.

[1] https://github.com/Netflix/eureka/blob/v1.4.10/eureka-core/src/main/java/com/netflix/eureka/cluster/PeerEurekaNodes.java#L230-L234
[2] https://github.com/Netflix/eureka/blob/v1.4.10/eureka-client/src/main/java/com/netflix/discovery/shared/transport/EurekaTransportConfig.java#L31

qiangdavidliu:
@mattnelson great to hear, thanks!

mattnelson:
@qiangdavidliu Is there anything else you need in order to get this PR merged?

qiangdavidliu:
@mattnelson nothing major, I just thought (per your last message) that you were making some additional changes to the PR for replication?

mattnelson:
I can include them with this PR if you want, but they are not related to the jersey2 uplift. They are isolated to running eureka in IP address resolver transport mode.

qiangdavidliu:
Ah. In that case then all good. A separate PR would be better if they are unrelated. Let me make a release of eureka without this going in first to have an artifact for the current head before merging this in.

mattnelson:
Created #831 for the IP address transport issue.

mattnelson:
@qiangdavidliu Now that https://github.com/Netflix/eureka/releases/tag/v1.4.11 is released, are there any other blockers?

qiangdavidliu:
@mattnelson I'm looking at merging this in and doing a release. There is one last issue where due to the move of "DiscoveryClientOptionalArgs" it creates a minor api incompatibility. I am assessing the blast radius of this at the moment.

qiangdavidliu:
@mattnelson there are a few small changes that I would like to make to this PR, but in the interest of not blocking you further, what do you say to moving this PR to against the (newly created) "contrib/jersey2" branch, where we can merge it in, and then I will make some PRs for the minor changes before we moving it back into master?

srinivasnagesh:
@mattnelson Can you please share any documentation available to use Eureka with Jersey2. I am looking to run Eureka Server as a packaged web app in a servlet container which isn't compatible with Jersey 1.

PR #831, opened 2016-08-17
Support eureka servers configured with IP addresses

mattnelson:
When configuring the `eureka.serviceUrl.default` with IP addresses the comparisons performed in `PeerEurekaNodes` and `StatusUtil` prevent the instances from recognizing themselves and therefore never report themselves as available.

As discussed on #821 (https://github.com/Netflix/eureka/pull/821#issuecomment-237416131) follow PR.


qiangdavidliu:
Thanks for the fix!

PR #833, opened 2016-08-30
Archaius2 client compatibility

qiangdavidliu:


eureka-client-archaius2/build.gradle
```diff
@@ -1,7 +1,14 @@
 apply plugin: 'nebula.test-jar'
 
+def archaius2Version = '2.1.4'
+
 dependencies {
-    compile project(':eureka-client')
+    compile (project(':eureka-client')) {
+        // archaius1
+        exclude group: 'com.netflix.archaius', module: 'archaius-core'
```
twicksell: If eureka-client depends directly on archaius-core, then this exclude will not prevent them from getting it. Excludes are not transitive, so while archaius1 may not appear in this project, it will be present for everyone who depends on it. We may need to consider refactoring eureka-client to shift this dependency to another place.

eureka-client-archaius2/src/main/java/com/netflix/appinfo/Ec2EurekaArchaius2InstanceConfig.java
```diff
@@ -75,7 +80,10 @@ else if ((amazonInfo.get(MetaDataKey.publicHostname) == null)
             // public hostname might be null
             amazonInfo.getMetadata().put(MetaDataKey.publicHostname.getName(),
                     (amazonInfo.get(MetaDataKey.localIpv4)));
-        }        
+        }
+
+        // TODO: Remove this when DiscoveryManager is finally no longer used
+        DiscoveryManager.getInstance().setEurekaInstanceConfig(this);
```
twicksell: What will be the replacement for this?

eureka-client-archaius2/src/main/java/com/netflix/appinfo/EurekaArchaius2InstanceConfig.java
```diff
@@ -38,88 +39,90 @@ public Name getName() {
...
+        this.defaultAppGroup = config.getString(FALLBACK_APP_GROUP_KEY, Values.UNKNOWN_APPLICATION);
         
         this.namespace = namespace;
         this.config = config.getPrefixedView(namespace);
         this.dcInfo = dcInfo;
-        
-        // TODO: archaius.deployment.environment=${env}?
+
+        // TODO: Remove this when DiscoveryManager is finally no longer used
+        DiscoveryManager.getInstance().setEurekaInstanceConfig(this);
```
twicksell: If we need to do this, I feel like we can at least do this somewhere else. A binding for something like DiscoveryManagerInitializer with dependencies on these Config classes might work.

eureka-client-archaius2/src/main/java/com/netflix/discovery/guice/LocalDevEurekaClientModule.java
```diff
@@ -1,29 +1,29 @@
...
 import com.netflix.discovery.shared.transport.EurekaArchaius2TransportConfig;
 import com.netflix.discovery.shared.transport.EurekaTransportConfig;
-import com.netflix.eventbus.impl.EventBusImpl;
-import com.netflix.eventbus.spi.EventBus;
 
 /**
  * @author David Liu
  */
-public class LocalDevEurekaClientModule extends AbstractModule {
+public final class LocalDevEurekaClientModule extends AbstractModule {
```
twicksell: Don't see a lot of differences between this and Ec2EurekaClientModule. Should one install the other, or have some kind of inheritance to avoid the duplicate code?

eureka-client-archaius2/src/main/java/com/netflix/appinfo/providers/Archaius2VipAddressResolver.java
```diff
@@ -0,0 +1,49 @@
...
+    @Inject
+    public Archaius2VipAddressResolver(Config config) {
+        this.config = config;
+    }
+
+    @Override
+    public String resolveDeploymentContextBasedVipAddresses(String vipAddressMacro) {
+        String result = vipAddressMacro;
+
+        if (vipAddressMacro == null) {
```
elandau: Feels like this check should come first before assigning result

eureka-client-archaius2/src/main/java/com/netflix/appinfo/providers/Archaius2VipAddressResolver.java
```diff
@@ -0,0 +1,49 @@
...
+            return null;
+        }
+
+        Matcher matcher = VIP_ATTRIBUTES_PATTERN.matcher(result);
+        while (matcher.find()) {
+            String key = matcher.group(1);
+            String value = config.getString(key, "");
+
+            logger.debug("att:" + matcher.group());
```
elandau: Can these four lines be consolidated into one?

eureka-client-archaius2/src/main/java/com/netflix/discovery/EurekaArchaius2ClientConfig.java
```diff
@@ -28,119 +30,127 @@ public EurekaArchaius2ClientConfig(Config config, EurekaTransportConfig transpor
...
-        return config.getString("region", defaultRegion);
+        return config.getString(
+                CLIENT_REGION_KEY,
+                config.getString(CLIENT_REGION_FALLBACK_KEY, Values.DEFAULT_CLIENT_REGION)
+        );
     }
 
     public String[] getAvailabilityZones(String region) {
-        return config.getString(String.format("%s.availabilityZones", region), DEFAULT_ZONE).split(",");
+        return config.getString(String.format("%s." + CONFIG_AVAILABILITY_ZONE_PREFIX, region), DEFAULT_ZONE).split(",");
```
elandau: Trim the results.  This can be done easily using Guava's Splitter.

eureka-client-archaius2/src/main/java/com/netflix/discovery/guice/LocalDevEurekaClientModule.java
elandau: +1 but don't use inheritance.  Create a new internal module with the common bindings

eureka-client/src/main/java/com/netflix/appinfo/InstanceInfo.java
```diff
@@ -375,6 +380,11 @@ public static Builder newBuilder() {
             return new Builder();
         }
 
+        public Builder withVipAddressResolver(VipAddressResolver vipAddressResolver) {
```
elandau: Looks like order is important when calling withVipAddressResolver and setVIPAddress.  A better approach would be to cache the value passed to setVIPAddress and call the resolver in build().  You can also set vipAddressResolver = LazyHolder.DEFAULT_VIP_ADDRESS_RESOLVER in the constructor if it's not set right before resolving the address.

eureka-client/src/main/java/com/netflix/appinfo/providers/Archaius1VipAddressResolver.java
```diff
@@ -0,0 +1,39 @@
...
+
+    private static final Logger logger = LoggerFactory.getLogger(Archaius1VipAddressResolver.class);
+
+    private static final Pattern VIP_ATTRIBUTES_PATTERN = Pattern.compile("\\$\\{(.*?)\\}");
+
+    @Override
+    public String resolveDeploymentContextBasedVipAddresses(String vipAddressMacro) {
+        String result = vipAddressMacro;
+
+        if (vipAddressMacro == null) {
```
elandau: This should be called before result is assigned

eureka-client/src/main/java/com/netflix/appinfo/providers/Archaius1VipAddressResolver.java
```diff
@@ -0,0 +1,39 @@
+package com.netflix.appinfo.providers;
+
+import com.netflix.config.DynamicPropertyFactory;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class Archaius1VipAddressResolver implements VipAddressResolver {
```
elandau: This code repeats for Archaius1 and Archaius2.  Any thoughts on creating an AbstractVipAddressResolver to hold the common code or put this in a utility class.

eureka-client-archaius2/src/main/java/com/netflix/appinfo/providers/Archaius2VipAddressResolver.java
```diff
@@ -0,0 +1,49 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.inject.Inject;
+import javax.inject.Singleton;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+@Singleton
+public class Archaius2VipAddressResolver implements VipAddressResolver {
```
elandau: Both archaius1 and archaius2 already support string interpolation.  This is code is completely unnecessary.

eureka-client-archaius2/build.gradle
qiangdavidliu: Yeah, but maybe in a separate PR for that.

eureka-client-archaius2/src/main/java/com/netflix/appinfo/Ec2EurekaArchaius2InstanceConfig.java
qiangdavidliu: Replacement is to not use the Manager and just create things either via DI or directly.

eureka-client-archaius2/src/main/java/com/netflix/appinfo/EurekaArchaius2InstanceConfig.java
qiangdavidliu: Yeah good idea, I'll take a stab at that.

eureka-client-archaius2/src/main/java/com/netflix/discovery/guice/LocalDevEurekaClientModule.java
qiangdavidliu: sure

eureka-client/src/main/java/com/netflix/appinfo/InstanceInfo.java
qiangdavidliu: yeah, changed.

eureka-client/src/main/java/com/netflix/appinfo/providers/Archaius1VipAddressResolver.java
qiangdavidliu: I don't think these two classes are really used now, only keeping them around for legacy support (and also making this change as they have config references).

eureka-client/src/main/java/com/netflix/appinfo/providers/VipAddressResolver.java
```diff
@@ -0,0 +1,15 @@
...
+ */
+public interface VipAddressResolver {
+
+    /**
+     * Convert <code>VIPAddress</code> by substituting environment variables if necessary.
+     *
+     * @param vipAddressMacro the macro for which the interpolation needs to be made.
+     * @return a string representing the final <code>VIPAddress</code> after substitution.
+     */
+    String resolveDeploymentContextBasedVipAddresses(String vipAddressMacro);
```
elandau: I realize this class is temporary but the class name and especially the method name seem extremely specific.  Why not create an interpolator api instead.  It greatly clarifies the purpose of this class.

```java
public interface StringInterpolator {
    String interpolate(String source);
}
```

eureka-client-archaius2/src/main/java/com/netflix/appinfo/providers/CompositeInstanceConfigFactory.java
```diff
@@ -0,0 +1,68 @@
...
+ * Guice treats Providers specially.
+ *
+ * @author David Liu
+ */
+@Singleton
+@ConfigurationSource("eureka-client")
+public class CompositeInstanceConfigFactory implements EurekaInstanceConfigFactory {
+    private static final Logger logger = LoggerFactory.getLogger(CompositeInstanceConfigFactory.class);
+
+    private static final String INIT_WITH_EC2_INSTANCE_CONFIG_KEY = "shouldInitAsEc2";
```
qiangdavidliu: open to other suggestions for this :).

eureka-client/src/main/java/com/netflix/discovery/DiscoveryManager.java
```diff
@@ -40,31 +46,80 @@
...
     private static final Logger logger = LoggerFactory.getLogger(DiscoveryManager.class);
-    private DiscoveryClient discoveryClient;
 
-    private EurekaInstanceConfig eurekaInstanceConfig;
-    private EurekaClientConfig eurekaClientConfig;
-    private static final DiscoveryManager s_instance = new DiscoveryManager();
+    private static volatile DiscoveryManager INSTANCE;
+
+    @Inject
+    private volatile Provider<EurekaInstanceConfig> eurekaInstanceConfig = new Provider<EurekaInstanceConfig>() {
```
elandau: What's this?

eureka-client/src/main/java/com/netflix/discovery/DiscoveryManager.java
```diff
@@ -40,31 +46,80 @@
...
+                public EurekaClientConfig get() {
+                    return config;
+                }
+            };
+        }
     }
 
-    public void setEurekaInstanceConfig(EurekaInstanceConfig eurekaInstanceConfig) {
-        this.eurekaInstanceConfig = eurekaInstanceConfig;
+    public synchronized void setEurekaInstanceConfig(final EurekaInstanceConfig config) {
```
elandau: Instead of using providers why not just mark these methods with @Inject?

eureka-client/src/main/java/com/netflix/discovery/DiscoveryManager.java
qiangdavidliu: Moving all of these to be providers so that they are still lazy even if DiscoveryManager is eagerly instantiated due to requestStaticInjection

eureka-client/src/main/java/com/netflix/discovery/DiscoveryManager.java
qiangdavidliu: These methods were not here to support DI, they are legacy setters that just need to be supported.

elandau: My point is that you don't need to write any code here.  Annotating this method with @Inject can be done instead of annotating the 'eurekaInstanceConfig' field with @Inject.

eureka-client/src/main/java/com/netflix/discovery/DiscoveryManager.java
elandau: This code is jumping through hoops to try to address the eager vs. lazy behavior of DiscoveryManager.  This should be handled outside of this class.

eureka-client-archaius2/build.gradle
```diff
@@ -1,7 +1,11 @@
 apply plugin: 'nebula.test-jar'
 
+def archaius2Version = '2.1.4'
```
twicksell: Some known issues in 2.1.4. Best to wait for a patch release or go back to 2.0.6

eureka-client-archaius2/src/main/java/com/netflix/discovery/guice/EurekaClientModule.java
```diff
@@ -0,0 +1,62 @@
...
+ * Custom {@link EurekaInstanceConfigFactory} may be registered as follows:
+ * <code>
+ * InjectorBuilder.fromModules(new EurekaClientModule() {
+ *      protected void configureEureka() {
+ *          bindEurekaInstanceConfigFactory().to(MyEurekaInstanceConfigFactory.class);
+ *      }
+ * }).createInjector()
+ * </code>
+ *
+ * Note that this module is NOT compatible with the archaius1 based {@link com.netflix.discovery.guice.EurekaModule}
```
twicksell: Pretty difficult to distinguish this module from the other. Slightly different names, same packages. Bound to confuse someone.

eureka-client-archaius2/build.gradle
qiangdavidliu: Sigh. I'll down rev to 2.0.6.

PR #835, opened 2016-08-31
Remove use of deprecated static accessor.

spencergibb:
Add a builder method and check that instanceInfo is not null in build().

Found this in a test where the static singleton was null :-(

### DIFF
--- a/eureka-core/src/main/java/com/netflix/eureka/util/StatusInfo.java
+++ b/eureka-core/src/main/java/com/netflix/eureka/util/StatusInfo.java
@@ -42,6 +42,11 @@ public class StatusInfo {
             return this;
         }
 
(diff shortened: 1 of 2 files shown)


eureka-core/src/main/java/com/netflix/eureka/util/StatusInfo.java
```diff
@@ -42,6 +42,11 @@ public Builder isHealthy(boolean b) {
             return this;
         }
 
+        public Builder instanceInfo(InstanceInfo instanceInfo) {
```
qiangdavidliu: minor quibble, change the method name to withInstanceInfo()?

spencergibb: sure, np

qiangdavidliu:
Thanks @spencergibb

PR #841, opened 2016-09-06
remove governator dep for eureka-client

qiangdavidliu:


eureka-client/src/main/java/com/netflix/discovery/InternalEurekaStatusModule.java
```diff
@@ -26,6 +26,7 @@
  * @author elandau
  *
  */
+@Deprecated
```
elandau: Just delete it!  It probably doesn't work anyway!

PR #846, opened 2016-09-15
Remove uses of archaius2 config.getPrefixedView()

qiangdavidliu:
It currently has issues when combined with config interpolation

### DIFF
--- a/eureka-client-archaius2/build.gradle
+++ b/eureka-client-archaius2/build.gradle
@@ -2,12 +2,6 @@ apply plugin: 'nebula.test-jar'
 
 def archaius2Version = '2.1.5'
 
-sourceSets {
-    test {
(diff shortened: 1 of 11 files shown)


twicksell:
:thumbsup:

brharrington:
Can you elaborate on what the issues are? Why not fix the prefixed views in archaius2?

qiangdavidliu:
The issue is to do with config interpolation when using prefixedView, where if prefixedView is used, then the interpolator automatically expect the same prefix to be attached to the interpolated config.

e.g. if my.prefix.config=${ENV_VAR}, then using prefixed view with prefix=my.prefix, ENV_VAR will not be loaded as the interpolator will expect my.prefix.ENV_VAR instead.

We will fix archaius2, however as the fix is less clear right now we are doing this to be unblocked.

PR #849, opened 2016-09-19
Add optional args to ApplicationInfoManager

qiangdavidliu:


eureka-client/src/test/java/com/netflix/appinfo/ApplicationInfoManagerTest.java
```diff
@@ -3,12 +3,16 @@
 import com.netflix.discovery.util.InstanceInfoGenerator;
 import org.junit.Before;
 import org.junit.Test;
+import org.omg.CORBA.UNKNOWN;
```
spencergibb: Whoops

qiangdavidliu: :) woops indeed. Thanks for the check!

PR #855, opened 2016-09-30
Contrib/jersey2 compatibility

qiangdavidliu:


qiangdavidliu:
@mattnelson going into master

ayusun:
We don't have any example on how to use it? Can you share some, or guide us to some documents, so that it will be easier for us.
Thanks.

mattnelson:
@ayusun 

There are some notes on the initial jersey2 PR #821. The readmes for eureka-client-jersey2/eureka-core-jersey2 could be updated before 1.6.0 is released.

PR #808, opened 2016-06-15
Prefer javax.inject Provider over guice Provider for #807

mattnelson:
Made the changes to prefer the javax.inject Provider over the guice Provider where possible. The remaining usages of the guice provider are blocked by https://github.com/google/guice/pull/926.

PR #862, opened 2016-10-20
Convert all remaining guice provider references to javax

mattnelson:
Follow up to #808 now that guice has been released and pulled in.


qiangdavidliu:
thanks @mattnelson

PR #869, opened 2016-11-07
Remove assignment of instanceInfo from static instance.

spencergibb:
f40f54a03d135379845aeb8c060b4f768fe7e7b5 made instanceInfo a required field to be set, but an assignment was still happening from `ApplicationInfoManager.getInstance().getInfo()` overwriting assigned field.


qiangdavidliu:
thanks @spencergibb

PR #874, opened 2016-11-15
Support HTTPS requests to servers that use multiple certificates

### LINKED ISSUE #873: Eureka Client cannot make HTTPS requests to servers that use multiple certificates via SNI
EurekaJerseyClientImpl uses deprecated, pre HTTPClient 4.3 APIs to create the SSLSocketFactory used in MonitoredConnectionManager [[1]](https://github.com/Netflix/eureka/blob/v1.6.0/eureka-client/src/main/java/com/netflix/discovery/shared/transport/jersey/EurekaJerseyClientImpl.java#L222). Support for https://en.wikipedia.org/wiki/Server_Name_Indication was introduced in Apache HttpClient 4.3.2 [[2]](https://hc.apache.org/news.html). When using the deprecated APIs, requests to servers that are protected with multiple certificates, which is typical of environments with many applications behind a single load balancer, will fail.

This can be demonstrated by setting the serviceUrl to `https://api.google.com/`. If the SSL handshake doesn't do SNI properly, the client will get the certificate for `www.google.com`, and hostname verification will fail with this exception. If SNI succeeds we won't see this exception (but we'll see another Eureka specific exception obviously).

```
2016-11-15 17:12:33.841 ERROR 65408 --- [nfoReplicator-0] c.n.d.s.t.d.RedirectingEurekaHttpClient  : Request execution error

com.sun.jersey.api.client.ClientHandlerException: javax.net.ssl.SSLException: Certificate for <api.google.com> doesn't match any of the subject alternative names: [www.google.com]
	at com.sun.jersey.client.apache4.ApacheHttpClient4Handler.handle(ApacheHttpClient4Handler.java:187) ~[jersey-apache-client4-1.19.1.jar:1.19.1]
	at com.sun.jersey.api.client.filter.GZIPContentEncodingFilter.handle(GZIPContentEncodingFilter.java:123) ~[jersey-client-1.19.1.jar:1.19.1]
	at com.netflix.discovery.EurekaIdentityHeaderFilter.handle(EurekaIdentityHeaderFilter.java:27) ~[eureka-client-1.4.11.jar:1.4.11]

...

Caused by: javax.net.ssl.SSLException: Certificate for <api.google.com> doesn't match any of the subject alternative names: [www.google.com]
	at org.apache.http.conn.ssl.AbstractVerifier.verify(AbstractVerifier.java:165) ~[httpclient-4.5.2.jar:4.5.2]
	at org.apache.http.conn.ssl.BrowserCompatHostnameVerifier.verify(BrowserCompatHostnameVerifier.java:61) ~[httpclient-4.5.2.jar:4.5.2]
	at org.apache.http.conn.ssl.AbstractVerifier.verify(AbstractVerifier.java:141) ~[httpclient-4.5.2.jar:4.5.2]
	at org.apache.http.conn.ssl.AbstractVerifier.verify(AbstractVerifier.java:114) ~[httpclient-4.5.2.jar:4.5.2]
	at org.apache.http.conn.ssl.SSLSocketFactory.verifyHostname(SSLSocketFactory.java:580) ~[httpclient-4.5.2.jar:4.5.2]
	at org.apache.http.conn.ssl.SSLSocketFactory.connectSocket(SSLSocketFactory.java:554) ~[httpclient-4.5.2.jar:4.5.2]
	at org.apache.http.conn.ssl.SSLSocketFactory.connectSocket(SSLSocketFactory.java:412) ~[httpclient-4.5.2.jar:4.5.2]
	at 
...
``` 

[1] https://github.com/Netflix/eureka/blob/v1.6.0/eureka-client/src/main/java/com/netflix/discovery/shared/transport/jersey/EurekaJerseyClientImpl.java#L222
[2] https://hc.apache.org/news.html

william-tran:
Use Apache HttpClient 4.3+ SSL handling which supports SNI, and
adapt it to the deprecated pre 4.3 API.

Fixes #873

### DIFF
--- a/eureka-client/src/main/java/com/netflix/discovery/shared/transport/jersey/EurekaJerseyClientImpl.java
+++ b/eureka-client/src/main/java/com/netflix/discovery/shared/transport/jersey/EurekaJerseyClientImpl.java
@@ -17,9 +17,13 @@ import com.sun.jersey.client.apache4.ApacheHttpClient4;
 import com.sun.jersey.client.apache4.config.ApacheHttpClient4Config;
 import com.sun.jersey.client.apache4.config.DefaultApacheHttpClient4Config;
 import org.apache.http.client.params.ClientPNames;
+import org.apache.http.conn.scheme.PlainSocketFactory;
 import org.apache.http.conn.scheme.Scheme;
 import org.apache.http.conn.scheme.SchemeRegistry;
+import org.apache.http.conn.ssl.SSLConnectionSocketFactory;
 import org.apache.http.conn.ssl.SSLSocketFactory;
+import org.apache.http.conn.ssl.X509HostnameVerifier;
+import org.apache.http.impl.conn.SchemeRegistryFactory;
 import org.apache.http.params.CoreProtocolPNames;
 import org.apache.http.params.HttpConnectionParams;
 import org.apache.http.params.HttpParams;
@@ -180,7 +184,7 @@ public class EurekaJerseyClientImpl implements EurekaJerseyClient {
                 } else if (trustStoreFileName != null) {
                     cm = createCustomSslCM();
                 } else {
-                    cm = new MonitoredConnectionManager(clientName);
+                    cm = createDefaultSslCM();
                 }
 
                 if (proxyHost != null) {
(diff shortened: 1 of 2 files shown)


william-tran:
It would be great if we could get this backported to 1.4.x like in my fork, and release a v1.4.12 so we can include this fix in a Spring Cloud Camden SR.

PR #876, opened 2016-11-17
use shutdownNow() instead of shutdown()

klues:
shutdown() does not prevent execution of previously submitted tasks, which can cause that AsyncResolver.shutdown() does not really shutdown all tasks -> memory leak


klues:
see https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ThreadPoolExecutor.html#shutdown() for effects of shutdown() and shutdownNow()

pauldburton:
This solved the same issue in our shared tomcat environment is there any chance of it being merged?

klues:
I built my own release containing this fix and we are using it in our project. I don't know why it is not being merged. If you need the release containing the fix and you don't want to build it on your own just contact me.

TatsianaKTV:
Hi @klues . If it possible give me please credentials for your eureka realise fix. Thanks.

klues:
see https://github.com/klues/eureka/tree/jar-containing-memory-leak-fix/bin for my built release containing this fix

TatsianaKTV:
@klues thanks.

qiangdavidliu:
@klues thanks for the contrib.

PR #895, opened 2017-01-31
Fixed missing url in log output

RichiMueller:
We get a lot of these log messages in our logfiles. We can't see which URL is causing the log message, as the placeholder {} is missing in the text. This pull request fixes this.


qiangdavidliu:
Thanks @RichiMueller

PR #897, opened 2017-02-03
Scope governator dependencies to reflect actual usage

mattnelson:
Governator was scoped as a compile dependency of eureka-core incorrectly. A few other modules needed it for testing. This fixes consumers from pulling in this dependency transitively when it is not used.


qiangdavidliu:
Thanks @mattnelson

PR #922, opened 2017-04-04
Improving conditional statement for map

bistros:
remoteRegionUrlsWithName can never null, so change null check condition

According to EurekaServerConfig.java's remoteRegionUrlsWithName comment
  Empty map if no remote region url is defined.

Now, Always running
https://github.com/Netflix/eureka/blob/4f24d9dcf90713b421f6c092fa63e707cbf74df3/eureka-core/src/main/java/com/netflix/eureka/registry/AbstractInstanceRegistry.java#L138-L151


qiangdavidliu:
Thanks for the PR @bistros

PR #931, opened 2017-04-21
Fix link rendering in README.md

macsj200:
There's a space in between the brackets and parentheses in the readme, which causes the links to not render properly. I've removed the space so the links look like links now.


qiangdavidliu:
Thank you @macsj200

PR #943, opened 2017-05-19
Allow for independent SSL context for EurekaJerseyClient

weswalker125:
The builder only allows SSL configuration through system settings, though it can be useful to operate the discovery client on a separate SSL context from the server/system level.  This exposes a method to supply an SSLContext object and use it in the "createCustomSslCM" method.

### DIFF
--- a/eureka-client/src/main/java/com/netflix/discovery/shared/transport/jersey/EurekaJerseyClientImpl.java
+++ b/eureka-client/src/main/java/com/netflix/discovery/shared/transport/jersey/EurekaJerseyClientImpl.java
@@ -93,6 +93,7 @@ public class EurekaJerseyClientImpl implements EurekaJerseyClient {
         private int connectionIdleTimeout;
         private EncoderWrapper encoderWrapper;
         private DecoderWrapper decoderWrapper;
+        private SSLContext sslContext;
 
         public EurekaJerseyClientBuilder withClientName(String clientName) {
             this.clientName = clientName;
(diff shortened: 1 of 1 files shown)


spencergibb:
👍

qiangdavidliu:
Thanks for the contrib @weswalker125

eureka-client/src/main/java/com/netflix/discovery/shared/transport/jersey/EurekaJerseyClientImpl.java
```diff
@@ -234,18 +240,20 @@ private MonitoredConnectionManager createSystemSslCM() {
...
-                    SSLContext sslContext = SSLContext.getInstance(PROTOCOL_SCHEME);
-                    KeyStore sslKeyStore = KeyStore.getInstance(KEYSTORE_TYPE);
+                    if (sslContext == null) {
+                        sslContext = SSLContext.getInstance(PROTOCOL_SCHEME);
+                        KeyStore sslKeyStore = KeyStore.getInstance(KEYSTORE_TYPE);
 
-                    fin = new FileInputStream(trustStoreFileName);
-                    sslKeyStore.load(fin, trustStorePassword.toCharArray());
+                        fin = new FileInputStream(trustStoreFileName);
+                        sslKeyStore.load(fin, trustStorePassword.toCharArray());
```
YangGuang001: Invalid Keystore format

i use jks,

PR #960, opened 2017-06-29
Upgrade XStream to 1.4.10, added tests around CVE-2017-7957

borlafu:
Applied http://x-stream.github.io/CVE-2017-7957.html#workarounds and added some tests around it.

After upgrading to 'com.thoughtworks.xstream:xstream:1.4.9' the workaround can be removed.

### DIFF
--- a/eureka-client/build.gradle
+++ b/eureka-client/build.gradle
@@ -8,7 +8,7 @@ configurations.all {
 
 dependencies {
     compile "com.netflix.netflix-commons:netflix-eventbus:0.3.0"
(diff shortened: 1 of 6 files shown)


spencergibb:
Why not upgrade to 1.4.10 instead of putting a workaround in?

borlafu:
+1: Makes sense. Had to modify the tests as the exception thrown is different to the one from the workaround.

qiangdavidliu:
Thanks @borlafu for the PR.

PR #965, opened 2017-07-13
initial commit

tcellucci:
1. memory tuning for deserialization
2. java 8

### DIFF
--- a/eureka-client/src/main/java/com/netflix/appinfo/InstanceInfo.java
+++ b/eureka-client/src/main/java/com/netflix/appinfo/InstanceInfo.java
@@ -57,6 +57,8 @@ import org.slf4j.LoggerFactory;
 @JsonRootName("instance")
 public class InstanceInfo {
 
+    private static final String VERSION_UNKNOWN = "unknown";
+
     /**
      * {@link InstanceInfo} JSON and XML format for port information does not follow the usual conventions, which
      * makes its mapping complicated. This class represents the wire format for port information.
@@ -142,18 +144,21 @@ public class InstanceInfo {
     @Auto
     private volatile Boolean isCoordinatingDiscoveryServer = Boolean.FALSE;
     @XStreamAlias("metadata")
-    private volatile Map<String, String> metadata = new ConcurrentHashMap<String, String>();
+    private volatile Map<String, String> metadata;
     @Auto
-    private volatile Long lastUpdatedTimestamp = System.currentTimeMillis();
+    private volatile Long lastUpdatedTimestamp;
     @Auto
-    private volatile Long lastDirtyTimestamp = System.currentTimeMillis();
+    private volatile Long lastDirtyTimestamp;
     @Auto
     private volatile ActionType actionType;
     @Auto
     private volatile String asgName;
-    private String version = "unknown";
+    private String version = VERSION_UNKNOWN;
 
     private InstanceInfo() {
+        this.metadata = new ConcurrentHashMap<String, String>();
+        this.lastUpdatedTimestamp = System.currentTimeMillis();
+        this.lastDirtyTimestamp = lastUpdatedTimestamp;
     }
 
     @JsonCreator
@@ -314,6 +319,7 @@ public class InstanceInfo {
                     return InstanceStatus.valueOf(s.toUpperCase());
                 } catch (IllegalArgumentException e) {
                     // ignore and fall through to unknown
+                    if (logger.isDebugEnabled()) logger.debug("illegal argument supplied to InstanceStatus.valueOf: {}, defaulting to {}", s, UNKNOWN);
                 }
             }
             return UNKNOWN;
@@ -322,10 +328,8 @@ public class InstanceInfo {
 
     @Override
     public int hashCode() {
-        final int prime = 31;
-        int result = 1;
-        result = prime * result + ((getId() == null) ? 0 : getId().hashCode());
-        return result;
+        String id = getId();
+        return (id == null) ? 31 : (id.hashCode() + 31);
     }
 
     @Override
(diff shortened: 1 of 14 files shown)


eureka-client/src/test/java/com/netflix/discovery/converters/EurekaJacksonCodecTest.java
```diff
@@ -226,4 +241,63 @@ public void testJacksonWrite() throws Exception {
...
+        for (int i=0; i < 30; i++) {
+            Future<Applications> appsFuture = executor.submit(task);
+            if (i%3 < 2) {
+                Thread.sleep(500);
+                System.out.println("cancelling..." + " i: " + i + " - " + (i%3));
+                appsFuture.cancel(true);
+            }
+            try {
+                Applications apps = appsFuture.get();
+               System.out.println("found some applications: " + apps.toString() + ":" + apps.getRegisteredApplications().size() + " i: " + i + " - " + (i%3));
```
qiangdavidliu: nit: there are some weird spacing in this file.

eureka-client/src/test/java/com/netflix/discovery/converters/EurekaJacksonCodecTest.java
```diff
@@ -226,4 +241,63 @@ public void testJacksonWrite() throws Exception {
 
         assertTrue(EurekaEntityComparators.equal(decoded, INSTANCE_INFO_1_A1));
     }
+    
+    @Test
```
qiangdavidliu: Are these two tests intended to be `@Ignore`'d?

eureka-client/src/main/java/com/netflix/appinfo/InstanceInfo.java
```diff
@@ -308,9 +309,11 @@ public InstanceInfo(InstanceInfo ii) {
...
 
         public static InstanceStatus toEnum(String s) {
-            for (InstanceStatus e : InstanceStatus.values()) {
-                if (e.name().equalsIgnoreCase(s)) {
-                    return e;
+            if (s != null) {
+                try {
+                    return InstanceStatus.valueOf(s.toUpperCase());
+                } catch (IllegalArgumentException e) {
+                    // ignore and fall through to unknown
```
mattnelson: Can a debug/trace level message be logged so this defaulting to unknown isn't completely ignored?

eureka-client/src/main/java/com/netflix/discovery/util/DeserializerStringCache.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.Optional;
+import java.util.function.Function;
+
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.ObjectReader;
+import com.fasterxml.jackson.databind.cfg.ContextAttributes;
+
+/**
+ * A non-locking alternative to {@link String#intern()} and {@link StringCache} that works with DeserializationContext.  Definitely
+ * NOT thread-safe.
```
mattnelson: Should ConcurrentHashMap be used instead? or if thread safety doesn't matter since this is a cache and misses here aren't a big deal update the docs to call that out. As it is written it implies that the lack safety could be an issue.

eureka-client/src/main/java/com/netflix/discovery/converters/EurekaJacksonCodec.java
```diff
@@ -393,152 +400,167 @@ protected void autoMarshalEligible(Object o, JsonGenerator jgen) {
...
-                        f.set(o, value);
-
-                    }
-                }
-            } catch (Throwable th) {
-                logger.error("Error in unmarshalling the object:", th);
-            }
+                        return (t, v) -> tryCatchLog(()->{ setterField.set(t, v); return null; });
+                    }  
+                } catch (Throwable th) {
```
mattnelson: Catch `Exception` instead.

eureka-client/src/main/java/com/netflix/discovery/converters/EurekaJacksonCodec.java
```diff
@@ -393,152 +400,167 @@ protected void autoMarshalEligible(Object o, JsonGenerator jgen) {
...
+                }                
+            });
+            action.accept(o, value);
+        }
+    }
+    
+    private static void tryCatchLog(Callable<Void> callable) {
+        try {    
+            callable.call();
+        } catch (Throwable th) {
```
mattnelson: Catch `Exception` instead.

eureka-client/src/main/java/com/netflix/discovery/util/DeserializerStringCache.java
tcellucci: the raison d'etre of DeserializerStringCache is to cache in a single-threaded context and thereby avoid the expense of locking.  I'll update the class-level documentation to call this out more clearly.

eureka-client/src/test/java/com/netflix/discovery/converters/EurekaJacksonCodecTest.java
tcellucci: added comment explaining why ignored

eureka-client/src/main/java/com/netflix/discovery/shared/Application.java
```diff
@@ -216,14 +216,15 @@ private void _shuffleAndStoreInstances(boolean filterUpInstances, boolean indexB
         synchronized (instances) {
             instanceInfoList = new ArrayList<InstanceInfo>(instances);
         }
-        if (indexByRemoteRegions || filterUpInstances) {
+        boolean remoteIndexingActive = indexByRemoteRegions && null != instanceRegionChecker && null != clientConfig
```
jhspaybar: The behavior through here doesn't look quite equivalent anymore.  It seems the old behavior would have gotten non-UP instances filtered out even when an instanceRegionChecker was null for example, but that the new behavior won't end up filtering out non-UP instances?

tcellucci: code will iterate and filter whether remote indexing is active or not; here's the guard condition:
if (remoteIndexingActive OR filterUpInstances) {
 if filterUpInstances [do filtering]
 if remoteIndexingActive [do remote indexing]
}

.

tcellucci:
had some more changes, going to close this PR and resubmit

PR #969, opened 2017-07-23
Misc. fixes

dtrebbien:
In testing my [SLF4J Helper](http://plugins.netbeans.org/plugin/72557/) plugin for NetBeans IDE, I found a few minor improvements that can be made.

### DIFF
--- a/eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
+++ b/eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
@@ -426,7 +426,7 @@ public class DiscoveryClient implements EurekaClient {
                     throw new IllegalStateException("Registration error at startup. Invalid server response.");
                 }
             } catch (Throwable th) {
-                logger.error("Registration error at startup.", th.getMessage());
+                logger.error("Registration error at startup: {}", th.getMessage());
                 throw new IllegalStateException(th);
             }
         }
--- a/eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
+++ b/eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
@@ -845,7 +845,7 @@ public class DiscoveryClient implements EurekaClient {
             logger.debug(PREFIX + "{} - Heartbeat status: {}", appPathIdentifier, httpResponse.getStatusCode());
             if (httpResponse.getStatusCode() == 404) {
                 REREGISTER_COUNTER.increment();
-                logger.info("{} - Re-registering apps/{}", PREFIX + appPathIdentifier, instanceInfo.getAppName());
+                logger.info(PREFIX + "{} - Re-registering apps/{}", appPathIdentifier, instanceInfo.getAppName());
                 long timestamp = instanceInfo.setIsDirtyWithTime();
                 boolean success = register();
                 if (success) {
--- a/eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
+++ b/eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
@@ -276,7 +276,7 @@ public class DiscoveryClient implements EurekaClient {
                     if (null != backupRegistryClassName) {
                         try {
                             backupRegistryInstance = (BackupRegistry) Class.forName(backupRegistryClassName).newInstance();
-                            logger.info("Enabled backup registry of type " + backupRegistryInstance.getClass());
+                            logger.info("Enabled backup registry of type {}", backupRegistryInstance.getClass());
                         } catch (InstantiationException e) {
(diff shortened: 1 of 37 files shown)


dtrebbien:
Rebased to fix the two conflicts.

qiangdavidliu:
@dtrebbien thanks for the PR, we will take a look at this.

eureka-client/src/main/java/com/netflix/discovery/shared/transport/decorator/RetryableEurekaHttpClient.java
```diff
@@ -126,7 +126,7 @@ public void shutdown() {
                 }
                 logger.warn("Request execution failure with status code {}; retrying on another server if available", response.getStatusCode());
             } catch (Exception e) {
-                logger.warn("Request execution failed with message: {}", e.getMessage());  // just log message as the underlying client should log the stacktrace
+                logger.warn("Request execution failed with message: {}", e.getMessage(), e);  // just log message as the underlying client should log the stacktrace
```
qiangdavidliu: The full stacktrace is explicitly not logged here (see comment).

eureka-resources/src/main/resources/css/main.css
```diff
@@ -146,7 +146,7 @@ tr.odd {
 span.hlist {
   padding-right: 10px;
   margin-right: 10px;
-  #background-color: #ffffcc;
+  background-color: #ffffcc;
```
qiangdavidliu: Let's keep the existing background-color please. Can just delete this line.

eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
```diff
@@ -822,7 +820,7 @@ boolean renew() {
         EurekaHttpResponse<InstanceInfo> httpResponse;
         try {
             httpResponse = eurekaTransport.registrationClient.sendHeartBeat(instanceInfo.getAppName(), instanceInfo.getId(), instanceInfo, null);
-            logger.debug("{} - Heartbeat status: {}", PREFIX + appPathIdentifier, httpResponse.getStatusCode());
+            logger.debug(PREFIX + "{} - Heartbeat status: {}", appPathIdentifier, httpResponse.getStatusCode());
```
qiangdavidliu: All the `PREFIX +` in this class can also be moved to `PREFIX + appPathIdentifier`

dtrebbien:
Hello David,

Thank you for reviewing this pull request.

I am currently traveling.  It might take me several days to make these
changes, but I will do so the soonest that I am able.

Daniel

eureka-resources/src/main/resources/css/main.css
dtrebbien: Done

eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
dtrebbien: In this case, because `PREFIX` is a constant variable (as defined in the Java Language Specification, section 4.12.4 `final` Variables), having `PREFIX` concatenated in the format string avoids a runtime string concatenation as the compiler will compute the constant expression and treat the same way as a literal; from section 3.10.5 String Literals:

> Strings computed by constant expressions (§15.28) are computed at compile time and then treated as if they were literals.

eureka-client/src/main/java/com/netflix/discovery/shared/transport/decorator/RetryableEurekaHttpClient.java
dtrebbien: Oops, I misread the comment. Sorry about that!

While rebasing to resolve the conflict with d8acdba71706ded8dd5613f98936211d471c6f16, I dropped the "Also log the Exception" commit.

dtrebbien:
Also, the latest 5 commits are new to the PR.

eureka-client/src/main/java/com/netflix/discovery/util/DeserializerStringCache.java
```diff
@@ -36,7 +36,6 @@
     }
 
     private static final Logger logger = LoggerFactory.getLogger(DeserializerStringCache.class);
-    private static final boolean logEnabled = logger.isTraceEnabled();
```
qiangdavidliu: This is a typo, the actual intent here is `debugLogEnabled =logger.isDebugEnabled(); `

dtrebbien: Done

dtrebbien:
I have now restored the is*Enabled() checks that guard against wasted computation.

qiangdavidliu:
Looks good, thanks @dtrebbien .

PR #970, opened 2017-07-24
 memory tuning

tcellucci:
moved to Java 8
extensive changes to jackson deserialization to reduce heap usage

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -52,8 +52,8 @@ subprojects {
 
     group = "com.netflix.${githubProjectName}"
 
-    sourceCompatibility = 1.7
-    targetCompatibility = 1.7
+    sourceCompatibility = 1.8
+    targetCompatibility = 1.8
 
     repositories {
         jcenter()
--- a/eureka-client/build.gradle
+++ b/eureka-client/build.gradle
@@ -1,5 +1,5 @@
 apply plugin: 'nebula.test-jar'
-apply plugin: 'provided-base'
+apply plugin: 'nebula.provided-base'
 
 configurations.all {
     // jersey2
@@ -18,6 +18,8 @@ dependencies {
     compile "org.apache.httpcomponents:httpclient:${apacheHttpClientVersion}"
     compile "com.google.inject:guice:${guiceVersion}"
 
+    compile "com.github.vlsi.compactmap:compactmap:1.2.1"
+    
     compile "com.fasterxml.jackson.core:jackson-annotations:${jacksonVersion}"
     compile "com.fasterxml.jackson.core:jackson-core:${jacksonVersion}"
     compile "com.fasterxml.jackson.core:jackson-databind:${jacksonVersion}"
--- a/eureka-client/src/main/java/com/netflix/appinfo/InstanceInfo.java
+++ b/eureka-client/src/main/java/com/netflix/appinfo/InstanceInfo.java
@@ -22,6 +22,7 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Function;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonIgnore;
@@ -55,6 +56,8 @@ import org.slf4j.LoggerFactory;
 @XStreamAlias("instance")
 @JsonRootName("instance")
 public class InstanceInfo {
+
+    private static final String VERSION_UNKNOWN = "unknown";
 
     /**
      * {@link InstanceInfo} JSON and XML format for port information does not follow the usual conventions, which
@@ -141,18 +144,21 @@ public class InstanceInfo {
     @Auto
     private volatile Boolean isCoordinatingDiscoveryServer = Boolean.FALSE;
     @XStreamAlias("metadata")
-    private volatile Map<String, String> metadata = new ConcurrentHashMap<String, String>();
+    private volatile Map<String, String> metadata;
     @Auto
-    private volatile Long lastUpdatedTimestamp = System.currentTimeMillis();
+    private volatile Long lastUpdatedTimestamp;
     @Auto
-    private volatile Long lastDirtyTimestamp = System.currentTimeMillis();
+    private volatile Long lastDirtyTimestamp;
     @Auto
     private volatile ActionType actionType;
     @Auto
     private volatile String asgName;
-    private String version = "unknown";
+    private String version = VERSION_UNKNOWN;
 
     private InstanceInfo() {
+        this.metadata = new ConcurrentHashMap<String, String>();
+        this.lastUpdatedTimestamp = System.currentTimeMillis();
+        this.lastDirtyTimestamp = lastUpdatedTimestamp;
     }
 
     @JsonCreator
@@ -308,9 +314,12 @@ public class InstanceInfo {
         UNKNOWN;
 
         public static InstanceStatus toEnum(String s) {
-            for (InstanceStatus e : InstanceStatus.values()) {
-                if (e.name().equalsIgnoreCase(s)) {
-                    return e;
+            if (s != null) {
+                try {
+                    return InstanceStatus.valueOf(s.toUpperCase());
+                } catch (IllegalArgumentException e) {
+                    // ignore and fall through to unknown
+                    if (logger.isDebugEnabled()) logger.debug("illegal argument supplied to InstanceStatus.valueOf: {}, defaulting to {}", s, UNKNOWN);
                 }
             }
             return UNKNOWN;
@@ -319,10 +328,8 @@ public class InstanceInfo {
 
     @Override
     public int hashCode() {
-        final int prime = 31;
-        int result = 1;
-        result = prime * result + ((getId() == null) ? 0 : getId().hashCode());
-        return result;
+        String id = getId();
+        return (id == null) ? 31 : (id.hashCode() + 31);
     }
 
     @Override
@@ -337,11 +344,12 @@ public class InstanceInfo {
             return false;
         }
         InstanceInfo other = (InstanceInfo) obj;
-        if (getId() == null) {
+        String id = getId();
+        if (id == null) {
             if (other.getId() != null) {
                 return false;
             }
-        } else if (!getId().equals(other.getId())) {
+        } else if (!id.equals(other.getId())) {
             return false;
         }
         return true;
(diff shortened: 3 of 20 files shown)


build.gradle
```diff
@@ -52,8 +52,8 @@ subprojects {
 
     group = "com.netflix.${githubProjectName}"
 
-    sourceCompatibility = 1.7
-    targetCompatibility = 1.7
+    sourceCompatibility = 1.8
```
spencergibb: I would hope a change like this would require a significant version change and not a patch version.

eureka-client/src/main/java/com/netflix/discovery/converters/EurekaJacksonCodec.java
```diff
@@ -94,22 +103,30 @@
...
     private final String appHashCodeKey;
 
     private final ObjectMapper mapper;
 
-    private final Map<Class<?>, ObjectReader> objectReaderByClass;
+    private final Map<Class<?>, Supplier<ObjectReader>> objectReaderByClass;
     private final Map<Class<?>, ObjectWriter> objectWriterByClass;
 
+    static EurekaClientConfig loadConfig() {
+        return com.netflix.discovery.DiscoveryManager.getInstance().getEurekaClientConfig();
```
spencergibb: Why use static refs here and not DI friendly?

tcellucci: that's existing code, relocated to an out-of-the-way method that preserves backward compatibility.  For DI, use the two-arg constructor that requires neither DiscoveryManager nor EurekaClientConfig.

build.gradle
tcellucci: this release will get tagged as a minor version, moving from 1.7.x to 1.8.x

tcellucci:
merging to create a candidate release, v1.8.0-rc.1

eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
```diff
@@ -1135,24 +1135,6 @@ private void reconcileAndLogDifference(Applications delta, String reconcileHashC
             return;
         }
 
-        if (logger.isDebugEnabled()) {
```
qiangdavidliu: 👍

eureka-client/src/main/java/com/netflix/discovery/shared/Applications.java
```diff
@@ -186,25 +175,18 @@ public Application getRegisteredApplications(String appName) {
...
      */
     public List<InstanceInfo> getInstancesBySecureVirtualHostName(String secureVirtualHostName) {
-        AtomicReference<List<InstanceInfo>> ref = this.shuffledSecureVirtualHostNameMap
-                .get(secureVirtualHostName.toUpperCase(Locale.ROOT));
-        if (ref == null || ref.get() == null) {
-            return new ArrayList<InstanceInfo>();
-        } else {
-            return ref.get();
-        }
+        return Optional.ofNullable(this.secureVirtualHostNameAppMap.get(secureVirtualHostName.toUpperCase(Locale.ROOT)))
```
qiangdavidliu: You can merge this and the insecure version into both calling an internal function with the two different maps.

eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
```diff
@@ -768,7 +768,7 @@ public InstanceInfo getNextServerFromEureka(String virtualHostname, boolean secu
                     + virtualHostname);
         }
         Applications apps = this.localRegionApps.get();
-        int index = (int) (apps.getNextIndex(virtualHostname.toUpperCase(Locale.ROOT),
+        int index = (int) (apps.getNextIndex(virtualHostname,
```
qiangdavidliu: This has lost the `.toUpperCase()`?

eureka-client/src/main/java/com/netflix/discovery/shared/Applications.java
```diff
@@ -252,201 +237,91 @@ public String getReconcileHashCode() {
...
+     * acquired.
+     * 
+     * @param instanceCountMap
+     *            the instance count map to use for generating the hash
      * @return the hash code for this instance
      */
-    public static String getReconcileHashCode(TreeMap<String, AtomicInteger> instanceCountMap) {
-        String reconcileHashCode = "";
+    public static String getReconcileHashCode(Map<String, AtomicInteger> instanceCountMap) {
+        StringBuilder reconcileHashCode = new StringBuilder(75);
```
qiangdavidliu: why 75?

eureka-client/src/main/java/com/netflix/discovery/shared/Applications.java
```diff
@@ -461,49 +336,34 @@ private void shuffleInstances(boolean filterUpInstances, boolean indexByRemoteRe
...
-    private void shuffleAndFilterInstances(
-            Map<String, AbstractQueue<InstanceInfo>> srcMap,
-            Map<String, AtomicReference<List<InstanceInfo>>> destMap,
-            Map<String, AtomicLong> vipIndexMap, boolean filterUpInstances) {
-        for (Map.Entry<String, AbstractQueue<InstanceInfo>> entries : srcMap.entrySet()) {
-            AbstractQueue<InstanceInfo> instanceInfoQueue = entries.getValue();
-            List<InstanceInfo> l = new ArrayList<InstanceInfo>(instanceInfoQueue);
+    private void shuffleAndFilterInstances(Map<String, VipIndexSupport> srcMap, boolean filterUpInstances) {
+
+        Random shuffleRandom = new Random();
```
qiangdavidliu: Use a ThreadLocal random?

eureka-client/src/test/java/com/netflix/discovery/converters/EurekaJacksonCodecIntegrationTest.java
```diff
@@ -0,0 +1,117 @@
...
+
+    /**
+     * parse discovery response in a long-running loop with a delay
+     * 
+     * @throws Exception
+     */
+    @Test
+    public void testRealDecode() throws Exception {
+        Applications applications;
+        File localDiscovery = new File("/var/folders/6j/qy6n1npj11x5j2j_9ng2wzmw0000gp/T/discovery-data-6054758555577530004.json"); //downloadRegistration(System.getProperty("discovery.url"));
```
qiangdavidliu: undo the commented out code?

eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
tcellucci: yes, it's been moved into Applications.getNextIndex for consistency (previously Applications class applies 'toUpperCase' to some of its other method parameters but not this one)

eureka-client/src/main/java/com/netflix/discovery/shared/Applications.java
tcellucci: some maths :). 
If you sum the lengths of each member in InstanceStatus (35), add in 2 * InstanceStatus.length for the separators that the hashing implementation adds (10), and figure an max size of 6 characters for the stringified 'count' associated with each status (25), then you get 76.  75 was subjectively used instead because fives.

eureka-client/src/main/java/com/netflix/discovery/shared/Applications.java
tcellucci: these methods part of public api, can't consolidate without breaking compatibility

eureka-client/src/main/java/com/netflix/appinfo/InstanceInfo.java
```diff
@@ -368,21 +377,26 @@ public boolean equals(Object obj) {
 
         private String namespace;
 
-        private Builder(InstanceInfo result, VipAddressResolver vipAddressResolver) {
+        private Builder(InstanceInfo result, VipAddressResolver vipAddressResolver, Function<String,String> intern) {
             this.vipAddressResolver = vipAddressResolver;
             this.result = result;
+            this.intern = intern != null ? intern : StringCache::intern;
```
tcellucci: there are three other pre-existing usages (constructor + 2 'newBuilder' factory methods) that call this new private constructor.  Each of those usages would have to specify StringCache::intern as the interning function, instead of having the default established in a single place.

eureka-client/src/main/java/com/netflix/discovery/shared/Applications.java
qiangdavidliu: Sorry I should have been more clear. I was thinking of something along the lines of 
```
private List<InstanceInfo> static internalGetInstancesByVirtualHostName(Map<> vipAppMapping, String vipName) {
    return Optional.ofNullable(vipAppMapping.get(vipName.toUpperCase(Locale.ROOT)))
    .map(VipIndexSupport::getVipList)
     .map(AtomicReference::get)
     .orElseGet(Collections::emptyList); 
}
```

eureka-client/src/main/java/com/netflix/discovery/shared/Applications.java
qiangdavidliu: 👍 Be nice to have that in a comment

eureka-client/src/main/java/com/netflix/appinfo/InstanceInfo.java
qiangdavidliu: All three use cases are still internal to this class, so it seems reasonable to explicitly specify at each place. The benefit we'd get is that the if check would no longer need to be performed at each builder creation.

eureka-client/build.gradle
```diff
@@ -18,6 +18,8 @@ dependencies {
     compile "org.apache.httpcomponents:httpclient:${apacheHttpClientVersion}"
     compile "com.google.inject:guice:${guiceVersion}"
 
+    compile "com.github.vlsi.compactmap:compactmap:1.2.1"
```
brharrington: It looks like this library is licensed with LGPL. I think there was general advice to keep everything Apache 2 if possible so the set of licenses consumers need to worry about is minimal.

eureka-client/src/main/java/com/netflix/appinfo/InstanceInfo.java
```diff
@@ -308,9 +314,12 @@ public InstanceInfo(InstanceInfo ii) {
...
         public static InstanceStatus toEnum(String s) {
-            for (InstanceStatus e : InstanceStatus.values()) {
-                if (e.name().equalsIgnoreCase(s)) {
-                    return e;
+            if (s != null) {
+                try {
+                    return InstanceStatus.valueOf(s.toUpperCase());
+                } catch (IllegalArgumentException e) {
+                    // ignore and fall through to unknown
+                    if (logger.isDebugEnabled()) logger.debug("illegal argument supplied to InstanceStatus.valueOf: {}, defaulting to {}", s, UNKNOWN);
```
brharrington: In this particular case couldn't you just remove the condition? It is already using placeholders to avoid substituting in the string if debug is not enabled and the logging library will be doing the same check internally.

Issue #973, opened 2017-08-04
InstanceInfoReplicator register service after unregistered when shutting down


jingege:
```
20:36:28.856 [Thread-8] INFO   c.netflix.discovery.DiscoveryClient.shutdown 852 - Shutting down DiscoveryClient ...
20:36:28.857 [Thread-8] INFO   c.netflix.discovery.DiscoveryClient.unregister 884 - Unregistering ...
20:36:28.861 [Thread-8] INFO   c.netflix.discovery.DiscoveryClient.unregister 886 - DiscoveryClient_SHOWOFF/10.242.36.20:showoff:19001:0.1.0 - deregister  status: 200
20:36:28.862 [DiscoveryClient-InstanceInfoReplicator-0] INFO   c.netflix.discovery.DiscoveryClient.register 795 - DiscoveryClient_SHOWOFF/10.242.36.20:showoff:19001:0.1.0: registering service...
20:36:28.865 [DiscoveryClient-InstanceInfoReplicator-0] INFO   c.netflix.discovery.DiscoveryClient.register 804 - DiscoveryClient_SHOWOFF/10.242.36.20:showoff:19001:0.1.0 - registration status: 204
20:36:28.867 [Thread-8] INFO   c.netflix.discovery.DiscoveryClient.shutdown 873 - Completed shut down of DiscoveryClient
```

Thread "Thread-8" is shutting down discovery client, but thread "DiscoveryClient-InstanceInfoReplicator-0" is still scheduling.

See the code in com.netflix.discovery.InstanceInfoReplicator.java

```
    public void stop() {
        scheduler.shutdownNow();
        started.set(false);
    }
```

in JDK 8, shutdownNow says "This method does not wait for actively executing tasks to
     * terminate.  Use {@link #awaitTermination awaitTermination} to
     * do that."

Should we change shutdownNow to awaitTermination to avoid this situation?

jingege:
PR #974

PR #974, opened 2017-08-04
Avoid scheduling when DiscoveryClient.shutdown called

jingege:
#973

### DIFF
--- a/eureka-client/src/main/java/com/netflix/discovery/InstanceInfoReplicator.java
+++ b/eureka-client/src/main/java/com/netflix/discovery/InstanceInfoReplicator.java
@@ -68,7 +68,11 @@ class InstanceInfoReplicator implements Runnable {
     }
 
     public void stop() {
-        scheduler.shutdownNow();
+        try {
+            scheduler.awaitTermination(3, TimeUnit.SECONDS);
+        } catch (InterruptedException e) {
+            logger.warn("Scheduler stop interrupted");
+        }
         started.set(false);
     }
 
--- a/eureka-client/src/main/java/com/netflix/discovery/InstanceInfoReplicator.java
+++ b/eureka-client/src/main/java/com/netflix/discovery/InstanceInfoReplicator.java
@@ -68,12 +68,25 @@ class InstanceInfoReplicator implements Runnable {
     }
 
     public void stop() {
-        try {
-            scheduler.awaitTermination(3, TimeUnit.SECONDS);
-        } catch (InterruptedException e) {
-            logger.warn("Scheduler stop interrupted");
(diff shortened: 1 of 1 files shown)


qiangdavidliu:
@jingege thanks for the PR, that makes sense. I believe you still need to call scheduler.shutdown() before blocking on an awaitTermination however. See the last example at the [ExecutorService javadoc](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ExecutorService.html). The phased shutdown example in the javadoc seems to be the right way to execute the shutdown, would you mind updating your PR to reflect that?

Appreciate the PR, thanks.

jingege:
@qiangdavidliu Thx for your advice. I updated the PR.

eureka-client/src/main/java/com/netflix/discovery/InstanceInfoReplicator.java
```diff
@@ -68,10 +69,27 @@ public void start(int initialDelayMs) {
     }
 
     public void stop() {
-        scheduler.shutdownNow();
+        Future latestPeriodic = scheduledPeriodicRef.get();
+        if (latestPeriodic != null && !latestPeriodic.isDone()) {
+            logger.info("Canceling the latest scheduled before stop");
+            boolean result = latestPeriodic.cancel(false);
+            System.out.println("Result:" + result);
```
rs017991: If keeping this, should use logger and reformat the message to give more context.

eureka-client/src/main/java/com/netflix/discovery/InstanceInfoReplicator.java
```diff
@@ -68,10 +69,27 @@ public void start(int initialDelayMs) {
     }
 
     public void stop() {
-        scheduler.shutdownNow();
+        Future latestPeriodic = scheduledPeriodicRef.get();
+        if (latestPeriodic != null && !latestPeriodic.isDone()) {
+            logger.info("Canceling the latest scheduled before stop");
+            boolean result = latestPeriodic.cancel(false);
```
qiangdavidliu: You shouldn't need to cancel the future here, shutting down the scheduler should handle that as the future is scheduled on that.

qiangdavidliu:
@jingege apologies for the late review, I have been away. Some minor comments.

abracadv8:
@qiangdavidliu - This user has not been active since this pull request.  I created a second pull request based on their work #1023 - Can you review that?

jingege:
@qiangdavidliu Very sorry for not responding this issue so long. I was so busy these days and missed the notification. @abracadv8 Thank you very much for finishing this PR.

PR #984, opened 2017-09-12
Fix Override Status Regression

qiangdavidliu:
Setting and deleting the override status on an eureka server has a regression due to the `lastDirtyTimestamp` being updated by the server in response to the set and delete, which can in turn lead to client update/registers potentially being ignored due to the timestamp being ahead on the server compared to client local.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -12,7 +12,7 @@ plugins {
 
 idea {
     project {
-        languageLevel = '1.7'
+        languageLevel = '1.8'
     }
 }
 
--- a/eureka-client/src/main/java/com/netflix/appinfo/InstanceInfo.java
+++ b/eureka-client/src/main/java/com/netflix/appinfo/InstanceInfo.java
@@ -1237,6 +1237,16 @@ public class InstanceInfo {
         lastDirtyTimestamp = System.currentTimeMillis();
     }
 
+    /**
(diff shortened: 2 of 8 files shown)


spencergibb:
What version does this affect (where was the regression)?

qiangdavidliu:
Unfortunately it was a while ago at release 1.4.11 (triggered due to PR #822, but not because of this PR). This issue mainly affects server side status override removals that does not specify a suggested status. The behaviour then will be:
1. server side status is updated to UNKNOWN (with an updated dirty timestamp) and will force a re-registration from the client
2. the client re-registers with an InstanceInfo that potentially has an older dirty timestamp.
3. the server will disregard the re-registration and keep on using its local copy, leading to an unchanged status of UNKNOWN (until the client refreshes at some point with an even later timestamp)

spencergibb:
OK. Wondering if you could patch the Java 7 version as well as master.

qiangdavidliu:
Sure thing. Once we've validated this fix, I'll cut a patch based on release 1.7.0 (last pre java8 release)

qiangdavidliu:
@spencergibb here's the diff (https://github.com/Netflix/eureka/compare/v1.7.0...v1.7.x). I'll cut 1.7.1 release this afternoon.

qiangdavidliu:
v1.7.1 is released with this fix on the 1.7.x branch.

Issue #816, opened 2016-07-12
Add support for custom HostnameVerifier


bitsofinfo:
There seems to be no way to specify a custom HostnameVerifier for the DiscoveryClient.

Can this support be easily added?

See #812

qiangdavidliu:
Hi @bitsofinfo that sounds like a useful addition for eureka. We are a bit swamped with other work at this moment, would you like to try a PR for this feature? Thanks.

mattnelson:
This has been released with 1.8.5

https://github.com/Netflix/eureka/releases/tag/v1.8.5

bitsofinfo:
should this be closed?

PR #1001, opened 2017-10-19
Support configuration of custom SSLContext/HostnameVerifier

mattnelson:
Implementation of #816 

Supports configuration of custom SSLContext/HostnameVerifier for jersey1/jersey2 implementations.

### DIFF
--- a/eureka-client-jersey2/src/main/java/com/netflix/discovery/shared/transport/jersey2/Jersey2ApplicationClientFactory.java
+++ b/eureka-client-jersey2/src/main/java/com/netflix/discovery/shared/transport/jersey2/Jersey2ApplicationClientFactory.java
@@ -16,6 +16,7 @@
 
 package com.netflix.discovery.shared.transport.jersey2;
 
+import javax.net.ssl.HostnameVerifier;
 import javax.net.ssl.SSLContext;
 import javax.ws.rs.client.Client;
 import javax.ws.rs.client.ClientBuilder;
@@ -31,6 +32,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
+import java.util.Optional;
 
 import com.netflix.appinfo.AbstractEurekaIdentity;
 import com.netflix.appinfo.EurekaAccept;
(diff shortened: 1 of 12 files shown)


mattnelson:
@qiangdavidliu @twicksell Could I get a review on this?

qiangdavidliu:
@mattnelson we'll take a look. Thanks for the PR.

eureka-core/src/main/java/com/netflix/eureka/registry/AbstractInstanceRegistry.java
```diff
@@ -959,14 +954,12 @@ public Applications getApplicationDeltasFromMultipleRegions(String[] remoteRegio
...
+            logger.debug("The number of elements in the delta queue is :{}", this.recentlyChangedQueue.size());
             while (iter.hasNext()) {
                 Lease<InstanceInfo> lease = iter.next().getLeaseInfo();
                 InstanceInfo instanceInfo = lease.getHolder();
-                Object[] args = {instanceInfo.getId(),
-                        instanceInfo.getStatus().name(),
-                        instanceInfo.getActionType().name()};
-                logger.debug("The instance id %s is found with status %s and actiontype %s", args);
+                logger.debug("The instance id {} is found with status {} and actiontype {}",
+                        instanceInfo.getStatus().name(), instanceInfo.getActionType().name());
```
qiangdavidliu: You missed an `instanceInfo.getId()` here.

eureka-client/src/main/java/com/netflix/discovery/shared/transport/jersey/JerseyEurekaHttpClientFactory.java
```diff
@@ -112,13 +116,18 @@ public void shutdown() {
     public static JerseyEurekaHttpClientFactory create(EurekaClientConfig clientConfig,
                                                        Collection<ClientFilter> additionalFilters,
                                                        InstanceInfo myInstanceInfo,
-                                                       AbstractEurekaIdentity clientIdentity) {
```
qiangdavidliu: I know this is very unlikely to be used outside of internal classes, but give that the method has a public signature, do you mind creating a new overload with the optionals and delegate over (rather than changing) from the original? Thanks.

eureka-client-jersey2/src/main/java/com/netflix/discovery/shared/transport/jersey2/Jersey2ApplicationClientFactory.java
```diff
@@ -78,14 +80,19 @@ public void shutdown() {
     public static Jersey2ApplicationClientFactory create(EurekaClientConfig clientConfig,
             Collection<ClientRequestFilter> additionalFilters,
             InstanceInfo myInstanceInfo,
-            AbstractEurekaIdentity clientIdentity) {
```
qiangdavidliu: I know this is very unlikely to be used outside of internal classes, but give that the method has a public signature, do you mind creating a new overload with the optionals and delegate over (rather than changing) from the original? Thanks.

PR #1003, opened 2017-10-24
Prevent "null" url server url prefix

encyphered:
When set `shouldUseDns` as true with null `eurekaServer.context`, `EndpointUtils#getServiceUrlsFromDNS` appends "null" string to service url postfix.

e.g. http://ec2-10-0-0-1.ap-northeast-2.compute.amazonaws.com/null/

Fixed this.

### DIFF
--- a/eureka-client/src/main/java/com/netflix/discovery/endpoint/EndpointUtils.java
+++ b/eureka-client/src/main/java/com/netflix/discovery/endpoint/EndpointUtils.java
@@ -145,10 +145,18 @@ public class EndpointUtils {
                     randomizer.randomize(ec2Urls);
                 }
                 for (String ec2Url : ec2Urls) {
-                    String serviceUrl = "http://" + ec2Url + ":"
-                            + clientConfig.getEurekaServerPort()
-                            + "/" + clientConfig.getEurekaServerURLContext()
-                            + "/";
+                    StringBuilder sb = new StringBuilder()
+                            .append("http://")
+                            .append(ec2Url)
+                            .append(":")
+                            .append(clientConfig.getEurekaServerPort());
+                    if (clientConfig.getEurekaServerURLContext() != null) {
+                        if (!clientConfig.getEurekaServerURLContext().startsWith("/")) {
+                            sb.append("/");
+                        }
+                        sb.append(clientConfig.getEurekaServerURLContext());
+                    }
+                    String serviceUrl = sb.toString();
                     logger.debug("The EC2 url is {}", serviceUrl);
                     serviceUrls.add(serviceUrl);
                 }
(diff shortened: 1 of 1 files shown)


eureka-client/src/main/java/com/netflix/discovery/endpoint/EndpointUtils.java
```diff
@@ -145,10 +145,18 @@ public void randomize(List<String> urlList) {
...
+                            .append(ec2Url)
+                            .append(":")
+                            .append(clientConfig.getEurekaServerPort());
+                    if (clientConfig.getEurekaServerURLContext() != null) {
+                        if (!clientConfig.getEurekaServerURLContext().startsWith("/")) {
+                            sb.append("/");
+                        }
+                        sb.append(clientConfig.getEurekaServerURLContext());
+                    }
+                    String serviceUrl = sb.toString();
```
qiangdavidliu: Hi. To replicate the existing behaviour, when the context is not null, this should append an additional `/`.

encyphered: You are right. Added commit to append slash.

eureka-client/src/main/java/com/netflix/discovery/endpoint/EndpointUtils.java
```diff
@@ -145,10 +145,20 @@ public void randomize(List<String> urlList) {
...
-                            + clientConfig.getEurekaServerPort()
-                            + "/" + clientConfig.getEurekaServerURLContext()
-                            + "/";
+                    StringBuilder sb = new StringBuilder()
+                            .append("http://")
+                            .append(ec2Url)
+                            .append(":")
+                            .append(clientConfig.getEurekaServerPort());
+                    if (clientConfig.getEurekaServerURLContext() != null) {
+                        if (!clientConfig.getEurekaServerURLContext().startsWith("/")) {
```
qiangdavidliu: This should be `endsWith("/")` here

eureka-client/src/main/java/com/netflix/discovery/endpoint/EndpointUtils.java
```diff
@@ -145,10 +145,20 @@ public void randomize(List<String> urlList) {
...
+                            sb.append("/");
+                        }
+                        sb.append(clientConfig.getEurekaServerURLContext());
```
qiangdavidliu: there should still be a final check here that 
```
if (!clientConfig.getEurekaServerURLContext().endsWith("/") {
   sb.append("/");
}
```

qiangdavidliu:
Merging this and addressing the PR comment in a separate PR.

PR #1010, opened 2017-11-07
catch potential exceptions around EurekaEventListener notify

qiangdavidliu:


eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
```diff
@@ -1605,7 +1605,11 @@ protected void onCacheRefreshed() {
      */
     protected void fireEvent(final EurekaEvent event) {
         for (EurekaEventListener listener : eventListeners) {
-            listener.onEvent(event);
+            try {
+                listener.onEvent(event);
+            } catch (Exception e) {
+                logger.info("Event {} throw an exception for listener {}", event, listener, e.getMessage());
```
jebeaudet: @qiangdavidliu you should log the whole exception and not the message here, the `e.getMessage()` will not log nothing here as there is no `{}` attached to it.

qiangdavidliu: Good catch, thanks. We'll fix it in a future release.

Issue #1012, opened 2017-11-14
retryableClientQuarantineRefreshPercentage parameter should increase the verification


holy12345:
I was studying eureka client registered to eureka server, that there is a place in the source code can be improved。
An eureka client looks for an eureka server by setting the eureka.client.service-url.defaultZone band,In my case, my properties file reads as follows:

```groovy
server.port=8081
eureka.client.service-url.defaultZone=http://localhost:8082,http://localhost:8083,http://localhost:8084,http://localhost:8085
eureka.client.transport.retryableClientQuarantineRefreshPercentage=10
```

I know by learning the source, when eureka client starts, it will register its own information to eureka server,Specific source code is as follows RetryableEurekaHttpClient.java
```java
@Override
protected <R> EurekaHttpResponse<R> execute(RequestExecutor<R> requestExecutor) {
    List<EurekaEndpoint> candidateHosts = null;
    int endpointIdx = 0;
    for (int retry = 0; retry < numberOfRetries; retry++) {
        EurekaHttpClient currentHttpClient = delegate.get();
        EurekaEndpoint currentEndpoint = null;
        if (currentHttpClient == null) {
            if (candidateHosts == null) {
                candidateHosts = getHostCandidates();
                if (candidateHosts.isEmpty()) {
                    throw new TransportException("There is no known eureka server; cluster server list is empty");
                }
            }
            if (endpointIdx >= candidateHosts.size()) {
                throw new TransportException("Cannot execute request on any known server");
            }

            currentEndpoint = candidateHosts.get(endpointIdx++);
            currentHttpClient = clientFactory.newClient(currentEndpoint);
        }

        try {
            EurekaHttpResponse<R> response = requestExecutor.execute(currentHttpClient);
            if (serverStatusEvaluator.accept(response.getStatusCode(), requestExecutor.getRequestType())) {
                delegate.set(currentHttpClient);
                if (retry > 0) {
                    logger.info("Request execution succeeded on retry #{}", retry);
                }
                return response;
            }
            logger.warn("Request execution failure with status code {}; retrying on another server if available", response.getStatusCode());
        } catch (Exception e) {
            logger.warn("Request execution failed with message: {}", e.getMessage());  // just log message as the underlying client should log the stacktrace
        }

        // Connection error or 5xx from the server that must be retried on another server
        delegate.compareAndSet(currentHttpClient, null);
        if (currentEndpoint != null) {
            quarantineSet.add(currentEndpoint);
        }
    }
[...]

PR #1013, opened 2017-11-15
The getHostCandidates method improves

holy12345:
Prevent retryableClientQuarantineRefreshPercentage from being greater than 1, causing Eureka Client to not register with Eureka Server。

Corresponding issues for https://github.com/Netflix/eureka/issues/1012


holy12345:
The getHostCandidates method improves，
Prevent retryableClientQuarantineRefreshPercentage from being greater than 1, causing Eureka Client to not register with Eureka Server。

Corresponding issues for #1012

qiangdavidliu:
Thanks @holy12345 .

holy12345:
@qiangdavidliu Thank you, may I ask which version of eureka can see this fix. Best wishes！

qiangdavidliu:
Hi @holy12345 we are probably going to look at some of the other outstanding PRs before a new release. This PR only adds a guard against misconfiguration (i.e. when the config value is specified incorrectly as > 1), and can be addressed for now by appropriately setting the config. Is there an urgent need for a release? Thanks.

holy12345:
@qiangdavidliu Thank you very much for your reply : )
I submitted two PR, please check, if there is no problem, hope to merge it. best wishes:)

PR #1017, opened 2017-11-25
Increase the catch exception code

holy12345:
Increase the catch exception code to ensure that when the exception occurs can be friendly to return to the client side

### DIFF
--- a/eureka-core/src/main/java/com/netflix/eureka/resources/InstanceResource.java
+++ b/eureka-core/src/main/java/com/netflix/eureka/resources/InstanceResource.java
@@ -277,16 +277,22 @@ public class InstanceResource {
     @DELETE
     public Response cancelLease(
             @HeaderParam(PeerEurekaNode.HEADER_REPLICATION) String isReplication) {
-        boolean isSuccess = registry.cancel(app.getName(), id,
+        try {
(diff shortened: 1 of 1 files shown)


holy12345:
@qiangdavidliu Hello, hope to pass this pr, the best wishes

qiangdavidliu:
Thanks @holy12345

eureka-core/src/main/java/com/netflix/eureka/resources/InstanceResource.java
```diff
@@ -277,16 +277,22 @@ public Response updateMetadata(@Context UriInfo uriInfo) {
...
-            logger.info("Not Found (Cancel): {} - {}", app.getName(), id);
-            return Response.status(Status.NOT_FOUND).build();
+            if (isSuccess) {
+                logger.debug("Found (Cancel): {} - {}", app.getName(), id);
+                return Response.ok().build();
+            } else {
+                logger.info("Not Found (Cancel): {} - {}", app.getName(), id);
+                return Response.status(Status.NOT_FOUND).build();
+            }
+        } catch (Throwable e) {
```
mattnelson: I know this is already merged, but it isn't released yet.

This should be catching `Exception`, there are very few scenarios in which `Throwable` should be caught and even worse here, swallowed.

PR #1020, opened 2017-11-30
Modify updateRenewalThreshold method

holy12345:
Hello, According to my understanding eureka deletes the client that did not send the heartbeat  by `AbstractInstanceRegistry#evict () `method.However, `numberOfRenewsPerMinThreshold` and `expectedNumberOfRenewsPerMin` are not updated immediately. Update these two parameters is carried out in a regular task. The specific code of this task is as follows：
` PeerAwareInstanceRegistryImpl#updateRenewalThreshold()` 

```
    private void updateRenewalThreshold() {
        try {
            Applications apps = eurekaClient.getApplications();
            int count = 0;
            for (Application app : apps.getRegisteredApplications()) {
                for (InstanceInfo instance : app.getInstances()) {
                    if (this.isRegisterable(instance)) {
                        ++count;
                    }
                }
            }
            synchronized (lock) {
                // Update threshold only if the threshold is greater than the
                // current expected threshold of if the self preservation is disabled.
                if ((count * 2) > (serverConfig.getRenewalPercentThreshold() * numberOfRenewsPerMinThreshold)
                        || (!this.isSelfPreservationModeEnabled())) {
                    this.expectedNumberOfRenewsPerMin = count * 2;
                    this.numberOfRenewsPerMinThreshold = (int) ((count * 2) * serverConfig.getRenewalPercentThreshold());
                }
            }
            logger.info("Current renewal threshold is : {}", numberOfRenewsPerMinThreshold);
        } catch (Throwable e) {
            logger.error("Cannot update renewal threshold", e);
        }
    }
```

According to my understanding, the logic of this method is this:
First, get the number of instances registered in Eureka Server. 
Then make a decision to decide whether to update `numberOfRenewsPerMinThreshold` and `expectedNumberOfRenewsPerMin`.
Then the problem is coming:
The first if expression` (count * 2)> (serverConfig.getRenewalPercentThreshold () * numberOfRenewsPerMinThreshold)` I think it is not accurate。
Because I understand every 15 minutes will recalculate the threshold, if at this time the number of registered instances did not reduce to 85% of the last time registered instance, it will recalculate。because `numberOfRenewsPerMinThreshold = expectedNumberOfRenewsPerMin * serverConfig.getRenewalPercentThreshold ()` so the above judgment is equivalent to ` (count * 2)> (serverConfig.getRenewalPercentThreshold () * serverConfig.getRenewalPercentThreshold () * expectedNumberOfRenewsPerMin)`this is not right。
should be `serverConfig.getRenewalPercentThreshold () * expectedNumberOfRenewsPerMin`，so I submitted a PR, hoping to solve this problem. Maybe my understanding is wrong, I hope you can answer me. Thank you very much, best wishes

### DIFF
--- a/eureka-core/src/main/java/com/netflix/eureka/registry/PeerAwareInstanceRegistryImpl.java
+++ b/eureka-core/src/main/java/com/netflix/eureka/registry/PeerAwareInstanceRegistryImpl.java
@@ -86,6 +86,7 @@ import javax.inject.Singleton;
  * </p>
  *
  * @author Karthik Ranganathan, Greg Kim
+ * @author Gang Li
  *
  */
 @Singleton
@@ -532,7 +533,7 @@ public class PeerAwareInstanceRegistryImpl extends AbstractInstanceRegistry impl
             synchronized (lock) {
                 // Update threshold only if the threshold is greater than the
                 // current expected threshold of if the self preservation is disabled.
-                if ((count * 2) > (serverConfig.getRenewalPercentThreshold() * numberOfRenewsPerMinThreshold)
+                if ((count * 2) > (serverConfig.getRenewalPercentThreshold() * expectedNumberOfRenewsPerMin)
                         || (!this.isSelfPreservationModeEnabled())) {
                     this.expectedNumberOfRenewsPerMin = count * 2;
                     this.numberOfRenewsPerMinThreshold = (int) ((count * 2) * serverConfig.getRenewalPercentThreshold());


holy12345:
@qiangdavidliu Hello, look forward to your reply, thank you very much, best wishes！

holy12345:
@qiangdavidliu Hello, can you review this PR? Thank you, best wishes ：)

holy12345:
@qiangdavidliu  ping...

qiangdavidliu:
@holy12345 thanks for the PR, you logic seems sound and there does seem to be an error in the code. Before I merge in your fix however, I'd appreciate it if you added a unit test that assert and validate the expected behaviour? Thanks.

holy12345:
Hello @qiangdavidliu 
Thank you for your answer, probably because my English is not very good and did not express clearly. Please forgive me if you have caused any confusion.
Now I re-answer the following(Reference eureka official document sentence):
`If any time, the renewals falls below the percent configured for that value (below 85% within 15 mins), the server stops expiring instances to protect the current instance registry information.`
This sentence has two main points:

- Whether to open the self-preservation mechanism (if it is not updated)

- Currently, the number of Eureka Clients with normal heartbeat accounts for more than 85% of Eureka Server's total storage

The code to accomplish the above two key points is as follows:
```
         if ((count * 2) > (serverConfig.getRenewalPercentThreshold() * numberOfRenewsPerMinThreshold)
                    || (!this.isSelfPreservationModeEnabled())) {
                this.expectedNumberOfRenewsPerMin = count * 2;
                this.numberOfRenewsPerMinThreshold = (int) ((count * 2) * serverConfig.getRenewalPercentThreshold());
        }
```
My understanding is as follows:
If you turn off self-preservation, these two parameters are updated under any circumstances.
If we open the self-preservation, then only if  the number of Eureka Clients with normal heartbeat accounts for more than 85% of Eureka Server's total storage will be updated.
What I think is the problem is the formula in the code `(count * 2) > (serverConfig.getRenewalPercentThreshold() * numberOfRenewsPerMinThreshold)`
In my opinion it should be  `(count * 2) > (serverConfig.getRenewalPercentThreshold() * expectedNumberOfRenewsPerMin)`

Maybe my understanding is not correct, I hope you can help me. Any one of your answers are a great encouragement to me, thanks again

qiangdavidliu:
Hi @holy12345 I do think your logic sounds correct. I would however appreciate if you added some unit tests in your PR to explicitly check for and verify the appropriate behaviour with the change. Thanks.

holy12345:
Hi @qiangdavidliu 
Thank you for your reply, thank you very much.
I will try to validate my logic by adding test code because the update is done in a thread and I currently have no idea how to test it(These days I have been looking for ways  but unfortunately did not find) .but do not worry I'm trying to figure out how to solve this problem

holy12345:
Hi @qiangdavidliu 
First of all, I'm sorry. I've been very busy before and I haven't updated this question.I think I'll close it f and when I think of a solution, I'll reopen it.
Thanks again : )

qiangdavidliu:
@holy12345 no worries, thanks.

PR #1021, opened 2017-12-01
Use a more elegant way to judge

holy12345:
Use `list.isEmpty ()` instead of `list.size>0` to make it more elegant.


qiangdavidliu:
Hi, we'd like to move away from using author tags as these can get confusing w.r.t. maintenance down the track. Otherwise, the PR looks good. Thanks.

holy12345:
@qiangdavidliu Okay, I removed it as you suggested. I apologize for the troubles caused to you. Thank you

qiangdavidliu:
No problem, thanks for the PR.

Issue #1025, opened 2017-12-19
com.netflix.discovery.InstanceInfoReplicator.onDemandUpdate() thow RejectedExecutionException


riyueshiwang:
hi, i am using `com.netflix.eureka:eureka-client:1.6.2.jar` within spring cloud.  when the method `org.springframework.cloud.context.scope.refresh.RefreshScope.refreshAll()` was invoked, the following exception occurred.
 i guessed that the method `com.netflix.discovery.InstanceInfoReplicator.start(int initialDelayMs)` had not been invoked before the method `com.netflix.discovery.InstanceInfoReplicator.onDemandUpdate()` did.
was it all right？

`2017-12-19 16:08:42.090 ERROR [lego-config,,,] 16808 --- [Apollo-Config-1] c.c.f.apollo.internals.AbstractConfig    : Failed to invoke config change listener com.iqiyi.solar.config.client.CloudConfig$1

java.util.concurrent.RejectedExecutionException: Task java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask@2bc7b30e rejected from java.util.concurrent.ScheduledThreadPoolExecutor@31c8bd8b[Terminated, pool size = 0, active threads = 0, queued tasks = 0, completed tasks = 4]
	at java.util.concurrent.ThreadPoolExecutor$AbortPolicy.rejectedExecution(ThreadPoolExecutor.java:2063)
	at java.util.concurrent.ThreadPoolExecutor.reject(ThreadPoolExecutor.java:830)
	at java.util.concurrent.ScheduledThreadPoolExecutor.delayedExecute(ScheduledThreadPoolExecutor.java:326)
	at java.util.concurrent.ScheduledThreadPoolExecutor.schedule(ScheduledThreadPoolExecutor.java:533)
	at java.util.concurrent.ScheduledThreadPoolExecutor.submit(ScheduledThreadPoolExecutor.java:632)
	at com.netflix.discovery.InstanceInfoReplicator.onDemandUpdate(InstanceInfoReplicator.java:77)
	at com.netflix.discovery.DiscoveryClient.registerHealthCheck(DiscoveryClient.java:618)
	at com.netflix.discovery.DiscoveryClient$$FastClassBySpringCGLIB$$a84c8cb4.invoke(<generated>)
	at org.springframework.cglib.proxy.MethodProxy.invoke(MethodProxy.java:204)
[...]

PR #1029, opened 2018-01-22
Verify if scheduler is active before submit new task

tjuchniewicz:
Fixes gh-1025.


qiangdavidliu:
Looks good, thanks for the contribution @tjuchniewicz

PR #1030, opened 2018-02-07
Add an EurekaInstanceInfoFactory abstraction to the creation of the initial internal InstanceInfo

qiangdavidliu:


eureka-client/src/main/java/com/netflix/appinfo/providers/EurekaConfigBasedInstanceInfoProvider.java
```diff
@@ -39,7 +40,14 @@
 
     @Inject
     public EurekaConfigBasedInstanceInfoProvider(EurekaInstanceConfig config) {
+        this(config, null);
+    }
+
+    public EurekaConfigBasedInstanceInfoProvider(EurekaInstanceConfig config, VipAddressResolver vipAddressResolver) {
         this.config = config;
+        if (vipAddressResolver != null) {
```
elandau: This null check is not necessary

mattnelson:
Would you be willing to add support for the non-DI `initComponent` flows as well?

https://github.com/Netflix/eureka/blob/v1.8.6/eureka-client/src/main/java/com/netflix/discovery/DiscoveryManager.java#L77

qiangdavidliu:
@mattnelson good catch, let me do that.

qiangdavidliu:
@mattnelson this work was for some internal refactoring that we needed. However after doing some testing the PR as it stands does not quite work for us and I will probably close this. Is this something you are interested in, or something you need?

mattnelson:
@qiangdavidliu I don't need it, only wanted to make sure that the other init flows were covered.

PR #1031, opened 2018-02-08
prevent replication flood caused by socket read time out leads to some node CPU 100%

skyesx:
when the instance become more then 3k,the batch replicated operation become easier to timeout.
because the replication request sender will retry the replication soon,it leads to a flood.
When it's happend,it appears to some eureka node ocupied 100% CPU and stucked.
This commit will help reduce the sympton,and write a clear log message to help debug.

By the way , i suggest to allow user to config the Congestion and TransientError retry delay time

### DIFF
--- a/eureka-core/src/main/java/com/netflix/eureka/cluster/ReplicationTaskProcessor.java
+++ b/eureka-core/src/main/java/com/netflix/eureka/cluster/ReplicationTaskProcessor.java
@@ -2,6 +2,8 @@ package com.netflix.eureka.cluster;
 
 import java.io.IOException;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


qiangdavidliu:
@skyesx thanks for the PR, this looks quite reasonable. Out of curiosity is this something that you have ran into in a deployed setup?

skyesx:
It didn't happen in our product environment.But it had  happen in our stress and develop environment.It's going to happen when 2K or more instances are registed in stress environment.In our develop environment , it happened once,  with nearly 1k instances, it last for a period of time, and then it recovered automatically .

qiangdavidliu:
Thanks @skyesx

Issue #1051, opened 2018-04-09
overriddenstatus changed to overriddenStatus breaks json parsing


ryanjbaxter:
In Spring Cloud we give the option for Eureka clients to not use Jersey and instead use a Spring `RestTemplate` as their HTTP client.  We noticed that when this change is made and we use Eureka Server 1.8.7 that the client can no longer parse the JSON response into an `InstanceInfo` class.  The reason seems to be a change in the case of the `overriddenstatus` property to `overriddenStatus`.  The variable in `InstanceInfo` is still `overriddenstatus`.  Looks like the change was made in this commit https://github.com/Netflix/eureka/commit/939957124a8f055c7d343d67d0842ae06cf59530#diff-1e0de94c9faa44a518abe30d94744178L63

ryanjbaxter:
The same problem does not show up when you look at the XML response

qiangdavidliu:
@ryanjbaxter version 1.8.8 released with the fix.

ryanjbaxter:
Awesome thanks, I will bump the version in Spring Cloud Netflix.

dosdebug:
@ryanjbaxter Is this added in `Finchley.RC1`, because I am still facing issue with mentioned release.

dosdebug:
Nevermind, `Finchley.RC1` still uses 1.8.8

ryanjbaxter:
@qiangdavidliu We are using Eureka 1.9 but I can still reproduce this

spencergibb:
Our non-jersey implementation doesn't use the Jackson codec do we still see the error

PR #1052, opened 2018-04-09
fix for issue #1051

qiangdavidliu:


spencergibb:
I guess I didn't pay enough attention to this since this doesn't actually fix our issue since the case we have is when we don't use jersey and therefore the Eureka codec.

spencergibb:
https://github.com/spring-cloud/spring-cloud-netflix/issues/2756

Issue #1056, opened 2018-04-21
Eureka proxy setting is not working


LuanLouis:
I want to intercept http messages between eureka clients and eureka servers,  and I add corresponding proxy settings in application.properties file like this:
```
eureka.client.proxy-host=127.0.0.1
eureka.client.proxy-port=8001
```
BUT the proxy settings value have't  be set to EurekaJerseyClientBuilder instance. I failed to intercept the http messges between eureka instance.
so I dive into source code , found that proxy settings is missing  in JerseyEurekaHttpClientFactory.buildLegacy() method:

```
@Override
        public JerseyEurekaHttpClientFactory build() {
            Map<String, String> additionalHeaders = new HashMap<>();
            if (allowRedirect) {
                additionalHeaders.put(HTTP_X_DISCOVERY_ALLOW_REDIRECT, "true");
            }
            if (EurekaAccept.compact == eurekaAccept) {
                additionalHeaders.put(EurekaAccept.HTTP_X_EUREKA_ACCEPT, eurekaAccept.name());
            }

            if (experimental) {
                return buildExperimental(additionalHeaders);
            }
            return buildLegacy(additionalHeaders, systemSSL);
        }

        private JerseyEurekaHttpClientFactory buildLegacy(Map<String, String> additionalHeaders, boolean systemSSL) {
            EurekaJerseyClientBuilder clientBuilder = new EurekaJerseyClientBuilder()
                    .withClientName(clientName)
                    .withUserAgent("Java-EurekaClient")
                    .withConnectionTimeout(connectionTimeout)
                    .withReadTimeout(readTimeout)
                    .withMaxConnectionsPerHost(maxConnectionsPerHost)
                    .withMaxTotalConnections(maxTotalConnections)
                    .withConnectionIdleTimeout((int) connectionIdleTimeout)
                    .withEncoderWrapper(encoderWrapper)
                    .withDecoderWrapper(decoderWrapper);
                    **_// proxy settings is lost here_**

            if (systemSSL) {
                clientBuilder.withSystemSSLConfiguration();
            } else if (sslContext != null) {
                clientBuilder.withCustomSSL(sslContext);
            }
            
            if (hostnameVerifier != null) {
                clientBuilder.withHostnameVerifier(hostnameVerifier);
            }

            EurekaJerseyClient jerseyClient = clientBuilder.build();
            ApacheHttpClient4 discoveryApacheClient = jerseyClient.getClient();
            addFilters(discoveryApacheClient);

[...]

PR #1057, opened 2018-04-21
bugfixing: Eureka Proxy Setting is not working

LuanLouis:
Eureka proxy settings is not working because of EurekaJerseyClientBuilder instance have no chance to set proxy settings during it's lifecycle.

related issue: #1056


qiangdavidliu:
@LuanLouis thanks for the fix.

PR #1079, opened 2018-05-31
change overriddenstatus field name in pojo to allow for better serialization compatibility

qiangdavidliu:
further relax deser behaviour on overriddenstatus field to allow for both lower and camel case

### DIFF
--- a/eureka-client/src/main/java/com/netflix/appinfo/InstanceInfo.java
+++ b/eureka-client/src/main/java/com/netflix/appinfo/InstanceInfo.java
@@ -137,7 +137,7 @@ public class InstanceInfo {
     private volatile DataCenterInfo dataCenterInfo;
     private volatile String hostName;
     private volatile InstanceStatus status = InstanceStatus.UP;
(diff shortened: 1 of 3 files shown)


ryanjbaxter:
@spencergibb I think this will fix https://github.com/spring-cloud/spring-cloud-netflix/issues/2941, do you agree?

spencergibb:
Yup

spencergibb:
This didn't fix things for us, since ` @JsonProperty("overriddenstatus")` is still the lowercase version :-(

qiangdavidliu:
I was wondering about that :(. Let me give it one last shot.

spencergibb:
Changing that to camel case breaks tests in `EurekaCodecCompatibilityTest`

spencergibb:
`Unrecognized field "overriddenstatus"`. The opposite of what we have. When you have something in a branch/fork somewhere, let me know, I'll build and try so we don't end up releasing it broken again.

PR #1088, opened 2018-06-21
Concurrent register of custom healthcheck handler in DiscoveryClient

tdanylchuk:
Some clients try to register their health checks and fail since there is possibility to rewrite healthcheck handler by default one.

Root cause:
after not thread safe null check DiscoveryClient will register default healthcheck handler even if during creation of that, custom one will be provided.
Resolution summary:
wrap healthcheck handler in AtomicReference and set default implementation only if reference still contain null.

Example is spring-cloud-netflix project and https://github.com/spring-cloud/spring-cloud-netflix/issues/2490 issue.


qiangdavidliu:
@tdanylchuk looks good, thanks for the contribution.

tdanylchuk:
My pleasure, thank you for supporting the project.

One question regrading this, when I can expect nearest release if there is observable data for this?

PR #1093, opened 2018-07-02
Add possibility to configure expected interval between clients' renew…

mgtriffid:
…s and not break self-preservation

Currently, if one sets renewal interval different from 30 seconds on clients, self-preservation doesn't work correctly. For example, see https://github.com/spring-cloud/spring-cloud-netflix/issues/373 . This PR adds possibility to configure expected renewal interval on Eureka server, so that if clients send renews every, say, 20 seconds, then SP threshold is `0.85 * 3 * count`, not `0.85 * 2 * count`.

### DIFF
--- a/eureka-core/src/main/java/com/netflix/eureka/registry/AbstractInstanceRegistry.java
+++ b/eureka-core/src/main/java/com/netflix/eureka/registry/AbstractInstanceRegistry.java
@@ -101,7 +101,7 @@ public abstract class AbstractInstanceRegistry implements InstanceRegistry {
 
     protected String[] allKnownRemoteRegions = EMPTY_STR_ARRAY;
     protected volatile int numberOfRenewsPerMinThreshold;
-    protected volatile int expectedNumberOfRenewsPerMin;
+    protected volatile int expectedNumberOfClientsSendingRenews;
 
     protected final EurekaServerConfig serverConfig;
     protected final EurekaClientConfig clientConfig;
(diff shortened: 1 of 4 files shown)


holy12345:
@mgtriffid Thanks your PR : )
Tiny things.
Could you please add some validate about `serverConfig.getExpectedClientRenewalIntervalSeconds()` is not zero.

neoremind:
I think there might be a more aggressive solution to make heartbeats interval configurable. Because there are too many hard codes the `/ 2` in `AbstractInstanceRegistry` and `PeerAwareInstanceRegistryImpl`, so that there is no need to add this parameter.

eureka-core/src/main/java/com/netflix/eureka/registry/AbstractInstanceRegistry.java
```diff
@@ -218,13 +218,13 @@ public void register(InstanceInfo registrant, int leaseDuration, boolean isRepli
...
-                        // Since the client wants to cancel it, reduce the threshold
-                        // (1
-                        // for 30 seconds, 2 for a minute)
-                        this.expectedNumberOfRenewsPerMin = this.expectedNumberOfRenewsPerMin + 2;
+                    if (this.expectedNumberOfClientsSendingRenews > 0) {
+                        // Since the client wants to register it, increase the number of clients sending renews
+                        this.expectedNumberOfClientsSendingRenews = this.expectedNumberOfClientsSendingRenews + 1;
                         this.numberOfRenewsPerMinThreshold =
-                                (int) (this.expectedNumberOfRenewsPerMin * serverConfig.getRenewalPercentThreshold());
+                                (int) (this.expectedNumberOfClientsSendingRenews
```
qiangdavidliu: This computation is repeated three times and might be better moved to a central method.

qiangdavidliu:
@mgtriffid no worries, that was a minor issue. Thank you for the PR.

PR #1096, opened 2018-07-10
Cache capacity configurable

holy12345:
Hi @qiangdavidliu 

When `ResponseCacheImpl#readWriteCacheMap` init, it use hardcode to make initial capacity, so I add a parameter which can let customer decide the initial capacity.

Please check it.

Thanks : )


qiangdavidliu:
Looks good, thanks @holy12345 .

PR #1100, opened 2018-07-11
Fix 1098

### LINKED ISSUE #1098: DiscoveryClient.localRegionApps not clear application without any instances
Eureka client still holds those applications who has no any instances in `DiscoveryClient.localRegionApps`. The only way to delete them I found is setting `eureka.client.log-delta-diff` as true. But this approach will add one more full fetch request after delta request. 

In method `updateDelta()`, the logic of handling instance deleted is confused. Even not apps exists, it still adds app back, and does removing op again. 

```java
if (ActionType.DELETED.equals(instance.getActionType())) {
    Application existingApp = applications.getRegisteredApplications(instance.getAppName());
    if (existingApp == null) {
        applications.addApplication(app);
    }
    logger.debug("Deleted instance {} to the existing apps ", instance.getId());
    applications.getRegisteredApplications(instance.getAppName()).removeInstance(instance);
}
```

I think the correct logic is if no app exists, do nothing. If exists, remove instance from it and **check if it has any instance up. If no, remove app.**

holy12345:
Hi @qiangdavidliu 

This PR fix #1098 please check it.

Thanks : )

### DIFF
--- a/eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
+++ b/eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
@@ -1222,8 +1222,8 @@ public class DiscoveryClient implements EurekaClient {
                         logger.debug("Deleted instance {} to the existing apps ", instance.getId());
                         existingApp.removeInstance(instance);
                         /*
-                         *We find all instance list from application(Instance status are not only the status is UP but also other status)
-                         *if instance list is empty, we remove the application.
+                         * We find all instance list from application(The status of instance status is not only the status is UP but also other status)
+                         * if instance list is empty, we remove the application.
                          */
                         if (existingApp.getInstancesAsIsFromEureka().isEmpty()) {
                             applications.removeApplication(existingApp);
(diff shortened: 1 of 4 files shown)


qiangdavidliu:
@holy12345 thanks for the PR. This looks good in theory, though I would like to get some additional verification on whether this is safe to do as removal of apps is not something that expected in current client behaviour. Can you please add some testing to verify? `DiscoveryClientRegistryTest` would be a good starting point for this. Thanks.

holy12345:
Hi @qiangdavidliu  First of all thank you!

I add some test case for this PR, please you check it .

You say 
> removal of apps is not something that expected in current client behaviour

So I think in some case there have many application has no instanceInfo when merge the delta information(the action type is DELETE)

According to your thoughts which is best ?

- When application has no instanceInfo, we remove application(this PR logic)

- Optimize the original logic, the code like this

```
if (ActionType.DELETED.equals(instance.getActionType())) {
    Application existingApp = applications.getRegisteredApplications(instance.getAppName());
    if (existingApp != null) {
        logger.debug("Deleted instance {} to the existing apps ", instance.getId());
        applications.getRegisteredApplications(instance.getAppName()).removeInstance(instance);
    }
}
```

Thanks again : )

qiangdavidliu:
Looks good, thanks @holy12345 .

PR #1101, opened 2018-07-13
EUREKA-1087 Add one more check for "is this my url", so that even whe…

mgtriffid:
…n each node is behind ELB, heartbeats are not replicated to the node they originally were sent to

### DIFF
--- a/eureka-core-jersey2/src/main/java/com/netflix/eureka/transport/Jersey2ReplicationClient.java
+++ b/eureka-core-jersey2/src/main/java/com/netflix/eureka/transport/Jersey2ReplicationClient.java
@@ -127,6 +127,26 @@ public class Jersey2ReplicationClient extends AbstractJersey2EurekaHttpClient im
     }
 
     @Override
+    public EurekaHttpResponse<String> fetchPeerId() {
+        Response response = null;
+        try {
+            response = jerseyClient.target(serviceUrl)
+                    .path(PeerEurekaNode.PEER_IDENTIFIER_PATH)
+                    .request(MediaType.TEXT_PLAIN)
+                    .get();
+            if (!isSuccess(response.getStatus())) {
+                return anEurekaHttpResponse(response.getStatus(), String.class).build();
+            }
+            String peerId = response.readEntity(String.class);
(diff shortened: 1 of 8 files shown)


qiangdavidliu:
@mgtriffid thanks for the PR. We will take a look at this.

mattnelson:
Can the existing instance id be used for self identification? It is already a unique identifier.

Introducing a new endpoint complicates deployment considerations with rollout of this feature and adds more complexity. Additionally, eureka is already pretty chatty between the peers and with heartbeats. I would like to explore other options that don't add more requests.

mgtriffid:
@mattnelson , probably, yes, existing instance id should work.
Speaking of not introducing new requests. When Eureka resolves addresses of peers through DNS, hosts in txt records are the only information it has. Sorry, I don't understand, how can Eureka find corresponding instance ids if not trying these hosts?
Maybe there already exists some request from which we could understand if caller == callee? If so, could you please suggest, where should I look?

Speaking of peers being chatty - please note that these requests are executed only for some short time after deployment, until Eureka finally calls itself and understands "yeah, this peer is actually me". After that, these requests don't happen.

Speaking of complicated deployment - yes, adding new resource to web server is required. But gradual rollout shouldn't be a problem: at some point Eureka will finally call itself (and if Eureka has this check - it means it has endpoint too), will figure out that called peer is actually the same node, and will stop doing these requests.

Overall, I certainly understand that this looks like a kludge, would be happy to find better solution.

qiangdavidliu:
@mgtriffid I think @mattnelson raise some good points here about additional complexity that would be involved with this change. Does your deployment model support node specific configs? If so, the below strategy may be an easier way of achieving the same goal:
1. introduce a config variable that can be set to override the isThisMe url that is checked at https://github.com/Netflix/eureka/blob/master/eureka-core/src/main/java/com/netflix/eureka/cluster/PeerEurekaNodes.java#L246. Doing this eureka nodes can be configured to match on custom names that does not have to be the local hostname.
2. in your deployment, set this config to be the ELB hostname that correspond to each node.

mgtriffid:
@qiangdavidliu , @mattnelson , thanks for ideas! Yes, this option with new configuration parameter is certainly much easier. Submitted new PR https://github.com/Netflix/eureka/pull/1107, declining this one.

PR #1108, opened 2018-08-06
Add an interface for AwsAsgUtil

PropertyOfTuckerSystems:
- Provide a replaceable abstraction over AWS ASG polling logic
- Add a setting to disable AWS ASG polling


eureka-core/src/main/java/com/netflix/eureka/aws/AwsAsgUtil.java
```diff
@@ -371,6 +371,11 @@ private TimerTask getASGUpdateTask() {
             @Override
             public void run() {
                 try {
+                    if (!serverConfig.shouldUseAwsAsgApi()) {
```
qiangdavidliu: You'll also need to handle the async refresh on line 70.

PropertyOfTuckerSystems: Resolved

PR #1114, opened 2018-08-15
Add possibility to register Eureka Server metrics to Spectator registry

mgtriffid:
Hi, this adds possibility to register `EurekaMonitors` to some Spectator registry. Helps if Eureka is deployed not as Tomcat web app. Users may provide own registry and get metrics reported to Atlas, for example.

However, if you think this is not enough and would be nice to add some Atlas configuration, so that even Tomcat web app could post metrics to Atlas, please let me know. However, I don't know how to pick proper `commonTags`, and we would need to provide own implementation of `AtlasConfig`, etc, etc. I bet if people are using Eureka and considering necessity of Atlas metrics, they already have some company-specific solutions for this, which can easily be applied with my PR.

### DIFF
--- a/eureka-core/src/main/java/com/netflix/eureka/util/EurekaMonitors.java
+++ b/eureka-core/src/main/java/com/netflix/eureka/util/EurekaMonitors.java
@@ -26,6 +26,9 @@ import com.netflix.appinfo.DataCenterInfo.Name;
 import com.netflix.servo.DefaultMonitorRegistry;
 import com.netflix.servo.annotations.DataSourceType;
 import com.netflix.servo.monitor.Monitors;
+import com.netflix.spectator.api.Counter;
+import com.netflix.spectator.api.Id;
+import com.netflix.spectator.api.Registry;
 
 /**
  * The enum that encapsulates all statistics monitored by Eureka.
@@ -89,9 +92,11 @@ public enum EurekaMonitors {
 
     @com.netflix.servo.annotations.Monitor(name = "count", type = DataSourceType.COUNTER)
     private final AtomicLong counter = new AtomicLong();
+    private Counter spectatorCounter;
 
     @com.netflix.servo.annotations.Monitor(name = "count-minus-replication", type = DataSourceType.COUNTER)
     private final AtomicLong myZoneCounter = new AtomicLong();
+    private Counter spectatorMyZoneCounter;
 
     /**
      * Increment the counter for the given statistic.
@@ -109,13 +114,26 @@ public enum EurekaMonitors {
      *            true if this a replication, false otherwise.
      */
     public void increment(boolean isReplication) {
-        counter.incrementAndGet();
+        incrementCounter();
 
         if (!isReplication) {
-            myZoneCounter.incrementAndGet();
+            incrementMyZoneCounter();
         }
     }
 
+    private void incrementCounter() {
+        counter.incrementAndGet();
+        if (spectatorCounter != null) {
+            spectatorCounter.increment();
+        }
+    }
+
+    private void incrementMyZoneCounter() {
+        myZoneCounter.incrementAndGet();
+        if (spectatorMyZoneCounter != null) {
+            spectatorMyZoneCounter.increment();
+        }
+    }
     /**
      * Gets the statistic name of this monitor.
      *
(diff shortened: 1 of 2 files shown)


qiangdavidliu:
Thanks for the PR @mgtriffid ! @brharrington do you have any thoughts to this?

eureka-core/src/main/java/com/netflix/eureka/util/EurekaMonitors.java
```diff
@@ -174,6 +192,26 @@ public static void registerAllStats() {
         }
     }
 
+    /**
+     * Registers metrics to given Spectator registry
+     * @param registry Spectator registry to register counters to
+     */
+
+    public static void registerAllStatsSpectator(Registry registry) {
```
brharrington: It looks like these are just here to allow access to the registry from a static context. Spectator has a notion of a [global registry](http://netflix.github.io/spectator/en/latest/intro/registry/#global-registry) for that purpose. This would have two main benefits:

1. When configuring the registry the user doesn't need to now about these specific methods scattered around various classes.
2. You could get rid of the null checks for the use. It will default to a noop implementation if no registries are added.

eureka-core/src/main/java/com/netflix/eureka/util/EurekaMonitors.java
```diff
@@ -109,13 +114,26 @@ public void increment() {
      *            true if this a replication, false otherwise.
      */
     public void increment(boolean isReplication) {
-        counter.incrementAndGet();
+        incrementCounter();
 
         if (!isReplication) {
-            myZoneCounter.incrementAndGet();
+            incrementMyZoneCounter();
+        }
```
brharrington: Typically we recommend modeling it so you can drill into the data easier. I would suggest making whether or not it is for replication a dimension, for example:

```java
private final Counter replicationCounter =
    Spectator.globalRegistry().counter("eureka." + name, "isReplication", "true");
private final Counter myZoneCounter =
    Spectator.globalRegistry().counter("eureka." + name, "isReplication", "false");

public void increment(boolean isReplication) {
  Counter c = isReplication ? replicationCounter : myZoneCounter;
  c.increment();
}
```

When querying the old `count` version would just be just restrict to a given name and it will automatically sum across the other dimensions. You could further restrict based on `isReplication` to get the part where that is true or false. This also makes it easy to group by `isReplication` to see both or visualize a percentage.

eureka-core/src/main/java/com/netflix/eureka/util/EurekaMonitors.java
```diff
@@ -89,9 +92,11 @@ private EurekaMonitors(String name, String description) {
 
     @com.netflix.servo.annotations.Monitor(name = "count", type = DataSourceType.COUNTER)
```
brharrington: Are the servo stats still needed? Internally this will cause duplication with both getting sent into Atlas.

brharrington:
> Helps if Eureka is deployed not as Tomcat web app. Users may provide own registry and get metrics reported to Atlas, for example.

A user can provide their own registry even if running as a Tomcat webapp, but they would probably need to pull it in as a library and build their own war file with additional config settings.

As a side note, you can also make Servo report into Atlas:

https://github.com/Netflix/servo/tree/master/servo-atlas

Recent versions of Servo also have a SpectatorContext which can be used to have all servo monitors report in via a Spectator registry. This is something we are using internally as we have to keep backwards compatibility for Servo instrumented apps for quite a while.

mgtriffid:
> Recent versions of Servo also have a SpectatorContext which can be used to have all servo monitors report in via a Spectator registry. This is what we are using internally now as we have to keep backwards compatibility for Servo instrumented apps for quite a while.

Wow, that's actually great, I think for our needs we may just increase version of Servo dependency to 0.12.25 and simply use `SpectatorContext`. Thanks for sharing this!

Given that you guys already have some solution for reporting this to Atlas (@brharrington mentioned duplication), and newer Servo provides a solution for this case, I think we can simply decline this PR - with `SpectatorContext` anybody can add Spectator and Atlas without any difficulties.

qiangdavidliu:
Thanks for that @mgtriffid . Closing the PR per the comment updates.

PR #1119, opened 2018-08-30
defaultZone separator support container space

ileler:


qiangdavidliu:
@ileler thank you for the PR. This looks fine, but would you mind adding some unit tests to verify both the existing and the future behaviour? Thanks.

ileler:
It's my pleasure.

PR #1122, opened 2018-09-17
Add User-Agent when requesting information from AWS Metadata service

willbengtson:
Add User-Agent when requesting information from AWS Metadata service


eureka-client/src/main/java/com/netflix/discovery/internal/util/AmazonInfoUtils.java
```diff
@@ -19,6 +19,7 @@ public static String readEc2MetadataUrl(MetaDataKey metaDataKey, URL url, int co
         HttpURLConnection uc = (HttpURLConnection) url.openConnection();
         uc.setConnectTimeout(connectionTimeoutMs);
         uc.setReadTimeout(readTimeoutMs);
+        uc.setRequestProperty("User-Agent", "eureka-java-client");
```
PropertyOfTuckerSystems: nit: this seems to contain an extra leading whitespace.

PropertyOfTuckerSystems: Seems to be an issue on latest Chrome for Linux, don't see it on Firefox, closing.

PR #1126, opened 2018-10-01
EUREKA-1120 Move write.lock higher in stack...

mgtriffid:
so that payload serialization is performed while locked

### DIFF
--- a/eureka-core/src/main/java/com/netflix/eureka/registry/AbstractInstanceRegistry.java
+++ b/eureka-core/src/main/java/com/netflix/eureka/registry/AbstractInstanceRegistry.java
@@ -885,7 +885,7 @@ public abstract class AbstractInstanceRegistry implements InstanceRegistry {
                     applicationInstancesMap.put(instanceInfo.getAppName(), app);
                     apps.addApplication(app);
                 }
-                app.addInstance(decorateInstanceInfo(lease));
+                app.addInstance(new InstanceInfo(decorateInstanceInfo(lease)));
             }
 
             boolean disableTransparentFallback = serverConfig.disableTransparentFallbackToOtherRegion();
@@ -962,7 +962,7 @@ public abstract class AbstractInstanceRegistry implements InstanceRegistry {
                     applicationInstancesMap.put(instanceInfo.getAppName(), app);
                     apps.addApplication(app);
                 }
-                app.addInstance(decorateInstanceInfo(lease));
+                app.addInstance(new InstanceInfo(decorateInstanceInfo(lease)));
             }
 
             if (includeRemoteRegion) {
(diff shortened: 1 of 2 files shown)


mattnelson:
Maybe I'm interpreting this wrong, but are the locks inverted? Could that be the root cause of this issue?

Mutative(write) actions are acquiring a read lock

https://github.com/Netflix/eureka/blob/ec2649e3a0a7eae5d3daabd2dfff7f8c2a91cc49/eureka-core/src/main/java/com/netflix/eureka/registry/AbstractInstanceRegistry.java#L191-L193

https://github.com/Netflix/eureka/blob/ec2649e3a0a7eae5d3daabd2dfff7f8c2a91cc49/eureka-core/src/main/java/com/netflix/eureka/registry/AbstractInstanceRegistry.java#L455-L459

Read operations are acquiring a write lock

https://github.com/Netflix/eureka/blob/ec2649e3a0a7eae5d3daabd2dfff7f8c2a91cc49/eureka-core/src/main/java/com/netflix/eureka/registry/AbstractInstanceRegistry.java#L865-L871

mgtriffid:
I don't think so, @mattnelson . RRWL's read lock can be entered by multiple threads, but if write lock is not locked, then other threads cannot enter even read lock. I think the intention here was to allow many threads modify different registered instances (statuses, metadata), but if response cache needs to be loaded - we take a pause and prepare response. But let's better wait for Eureka developers to respond to your concern.

qiangdavidliu:
@mgtriffid thanks for the PR. I checked some statistics and the data serialization (in extreme p99.9+ cases) can be pretty heavy, beyond even a second. With a write lock around this, this means instance registration and updates will be unable to proceed for that duration, which can be problematic.

Seems like the root issue here is the mutable InstanceInfo objects to be serialized. It might be worth doing a quick experiment to see which is cheaper: locking around the serialization or ensuring copies of the data is returned by the `getApplicationDelta*` methods.

mgtriffid:
Thanks for pointing this out, I actually was concerned about performance too, for some reason thought serialization is quick. Good that you have some stats from production of your scale.
[Naive benchmark|https://gist.github.com/mgtriffid/831b0154adaad305642c0881c4cd0639] shows that on average calling copying constructors is 11 times faster:
```
Encoding: 84819
Copying constructors: 6561
Now actual test:
Encoding: 81785
Copying constructors: 6803
Now without metadata
Encoding: 82318
Copying constructors: 7140
```
Will update PR.

mgtriffid:
Moved locks to original place, added copying constructors.

qiangdavidliu:
@pparth we run with a readonly tier internally (same REST API -> cache -> EurekaClient ---> write servers), which due to its nature mitigated the original issue referenced in this PR.

qiangdavidliu:
The PR looks good, thank you @mgtriffid for the fix.

PR #1132, opened 2018-10-12
eureka-1131: handle spot instance termination

krutsko:


eureka-client/src/main/java/com/netflix/appinfo/AmazonInfo.java
```diff
@@ -63,6 +63,8 @@
         availabilityZone("availability-zone", "placement/"),
         publicHostname("public-hostname"),
         publicIpv4("public-ipv4"),
+        spotInstanceTerminationTime("termination-time", "spot/"),
```
qiangdavidliu: very minor nit: mind calling this `spotTerminationTime`? That aligns more with how the rest of the enums are named.

eureka-client/src/main/java/com/netflix/appinfo/CloudInstanceConfig.java
```diff
@@ -50,7 +50,7 @@
     };
 
     private final AmazonInfoConfig amazonInfoConfig;
-    private final RefreshableAmazonInfoProvider amazonInfoHolder;
+    /* visible for testing */ RefreshableAmazonInfoProvider amazonInfoHolder;
```
qiangdavidliu: I would appreciate this be kept final to better ensure it is always correctly instantiated. An alternative for your test can be to add a new constructor that can be used to set `RefreshableAmazonInfoProvider`, with test visibility?

eureka-client/src/main/java/com/netflix/appinfo/ApplicationInfoManager.java
```diff
@@ -219,6 +219,21 @@ public void refreshDataCenterInfoIfRequired() {
             builder.setHostName(newAddress).setIPAddr(newIp).setDataCenterInfo(config.getDataCenterInfo());
             instanceInfo.setIsDirty();
         }
+
+        String existingSpotInstanceAction  = instanceInfo.getMetadata().get(AmazonInfo.MetaDataKey.spotInstanceAction.name);
+        DataCenterInfo dataCenterInfo = config.getDataCenterInfo();
+        if (dataCenterInfo instanceof AmazonInfo) {
+            AmazonInfo amazonInfo = (AmazonInfo) dataCenterInfo;
+            String newSpotInstanceAction = amazonInfo.getMetadata().get(AmazonInfo.MetaDataKey.spotInstanceAction.name);
```
qiangdavidliu: It is unfortunate that the API for `AmazonInfo` is a bit leaky, but the preferred API to extract data is
https://github.com/Netflix/eureka/blob/f8439886921e3655a301adb149a13eaa34268074/eureka-client/src/main/java/com/netflix/appinfo/AmazonInfo.java#L299, where the enum can be used directly.

eureka-client/src/main/java/com/netflix/appinfo/ApplicationInfoManager.java
```diff
@@ -219,6 +219,21 @@ public void refreshDataCenterInfoIfRequired() {
...
+            if (newSpotInstanceAction != null && !newSpotInstanceAction.equals(existingSpotInstanceAction)) {
+                logger.warn(String.format("The spot instance termination action changed from: %s => %s",
+                        existingSpotInstanceAction,
+                        newSpotInstanceAction));
+                InstanceInfo.Builder builder = new InstanceInfo.Builder(instanceInfo);
+                builder.setDataCenterInfo(config.getDataCenterInfo());
```
qiangdavidliu: There is duplication of code here with code on line 218/219/220, mind consolidating them? Thanks.

eureka-client/src/main/java/com/netflix/appinfo/ApplicationInfoManager.java
```diff
@@ -219,6 +219,21 @@ public void refreshDataCenterInfoIfRequired() {
             builder.setHostName(newAddress).setIPAddr(newIp).setDataCenterInfo(config.getDataCenterInfo());
             instanceInfo.setIsDirty();
         }
+
+        String existingSpotInstanceAction  = instanceInfo.getMetadata().get(AmazonInfo.MetaDataKey.spotInstanceAction.name);
```
qiangdavidliu: (mostly due to our bad naming) I think this is the wrong metadata map to get here, you want the metadata from the instanceInfo.getDataCenterInfo(), if it is an instance of AmazonInfo.

eureka-client/src/main/java/com/netflix/appinfo/ApplicationInfoManager.java
```diff
@@ -219,6 +219,21 @@ public void refreshDataCenterInfoIfRequired() {
...
+        DataCenterInfo dataCenterInfo = config.getDataCenterInfo();
+        if (dataCenterInfo instanceof AmazonInfo) {
+            AmazonInfo amazonInfo = (AmazonInfo) dataCenterInfo;
+            String newSpotInstanceAction = amazonInfo.getMetadata().get(AmazonInfo.MetaDataKey.spotInstanceAction.name);
+            if (newSpotInstanceAction != null && !newSpotInstanceAction.equals(existingSpotInstanceAction)) {
+                logger.warn(String.format("The spot instance termination action changed from: %s => %s",
```
qiangdavidliu: This is more you use case, is this change worth a WARN level log? Assuming spot termination actions can change often, this'll get noisy.

eureka-client/src/main/java/com/netflix/appinfo/CloudInstanceConfig.java
krutsko: Fixed.

eureka-client/src/main/java/com/netflix/appinfo/ApplicationInfoManager.java
krutsko: Fixed.

eureka-client/src/main/java/com/netflix/appinfo/ApplicationInfoManager.java
krutsko: I have changed to INFO. If you think that DEBUG is even better here, let me know. In our case, we don't have so much spot-instances termination.

eureka-client/src/main/java/com/netflix/appinfo/ApplicationInfoManager.java
krutsko: Got it. Fixed.

eureka-client/src/main/java/com/netflix/appinfo/ApplicationInfoManager.java
krutsko: Right. Fixed.

eureka-client/src/main/java/com/netflix/appinfo/AmazonInfo.java
krutsko: Sure. Changed to `spotTerminationTime`.

krutsko:
@qiangdavidliu thanks for looking into this. I have fixed all comments.

qiangdavidliu:
Thanks @krutsko

PR #1151, opened 2018-11-08
EUREKA-1149 Allow custom randomizer

mgtriffid:
so that services on same host could talk to different Eureka peers

### DIFF
--- a/eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
+++ b/eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
@@ -157,7 +157,6 @@ public class DiscoveryClient implements EurekaClient {
     private final AtomicReference<String[]> remoteRegionsRef;
     private final InstanceRegionChecker instanceRegionChecker;
 
-    private final EndpointUtils.ServiceUrlRandomizer urlRandomizer;
     private final Provider<BackupRegistry> backupRegistryProvider;
     private final EurekaTransport eurekaTransport;
 
@@ -326,7 +325,6 @@ public class DiscoveryClient implements EurekaClient {
 
         this.backupRegistryProvider = backupRegistryProvider;
 
-        this.urlRandomizer = new EndpointUtils.InstanceInfoBasedUrlRandomizer(instanceInfo);
         localRegionApps.set(new Applications());
 
         fetchRegistryGeneration = new AtomicLong(0);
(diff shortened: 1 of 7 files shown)


eureka-client/src/main/java/com/netflix/discovery/shared/resolver/ResolverUtils.java
```diff
@@ -41,13 +41,40 @@
 
     private static final Logger logger = LoggerFactory.getLogger(ResolverUtils.class);
 
-    private static final String LOCAL_IPV4_ADDRESS = SystemUtil.getServerIPv4();
-
     private static final Pattern ZONE_RE = Pattern.compile("(txt\\.)?([^.]+).*");
 
+    private static Randomizer RANDOMIZER = new Randomizer() {
```
elandau: ResolverUtils is a utility class and should have no state.  The static setter also makes it impossible to test different randomizers in parallel without forking the JVM.  While that may be an unlikely scenario we prefer to stick to this philosophy throughout the codebase.

eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
```diff
@@ -326,7 +325,6 @@ public synchronized BackupRegistry get() {
 
         this.backupRegistryProvider = backupRegistryProvider;
 
-        this.urlRandomizer = new EndpointUtils.InstanceInfoBasedUrlRandomizer(instanceInfo);
```
elandau: Let's keep the urlRandomizer as a configurable aspect of the DiscoveryClient and not static state of ResolverUtils. 
 This can be made customizable either via Guice bindings or configuration.  I'm also thinking this should be done as a ServiceUrlRandomizerFactory that takes an InstanceInfo argument.

eureka-client/src/main/java/com/netflix/discovery/shared/resolver/ResolverUtils.java
```diff
@@ -41,13 +41,40 @@
...
+
+        private final String LOCAL_IPV4_ADDRESS = SystemUtil.getServerIPv4();
```
elandau: Currently the local address is derived from the local InstanceInfo, which makes testing easier and keeps the source of truth for such information in once place.  For consistency I think it's best to continue using the local InstanceInfo.

mgtriffid:
Declining in favor of #1194.

PR #1153, opened 2018-11-13
Tweak Staleness metrics

elandau:
Tweak the staleness guages to return 0 for when the client is shut down or when refreshes are not enabled.  Otherwise the metric can be misleading by incorrectly making it appear that refreshes are not working.


eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
```diff
@@ -1686,6 +1690,10 @@ private long getLastSuccessfulHeartbeatTimePeriodInternal() {
     @com.netflix.servo.annotations.Monitor(name = METRIC_REGISTRY_PREFIX + "lastSuccessfulRegistryFetchTimePeriod",
             description = "How much time has passed from last successful local registry update", type = DataSourceType.GAUGE)
     private long getLastSuccessfulRegistryFetchTimePeriodInternal() {
+        if (!clientConfig.shouldRegisterWithEureka() || isShutdown.get()) {
```
mgtriffid: Not sure if I understand the first part of this condition, but maybe your intention was to check `shouldFetchRegistry` here and `shouldRegisterWithEureka` above?

elandau: Yes, these two were reversed.

Issue #1154, opened 2018-11-13
[Bug] If instance is not found when trying to update metadata via REST API, server error is thrown


mgtriffid:
It seems that Eureka server should return 404 in such case. Just like it does if instance is not found when trying to update status or DELETE instance from registry.

mgtriffid:
Related PR is merged, so closing this.

PR #1155, opened 2018-11-13
EUREKA-1154 Return 404 if instance is not found when updating metadata

mgtriffid:


mgtriffid:
Fixes #1154.

eureka-core/src/main/java/com/netflix/eureka/resources/InstanceResource.java
```diff
@@ -118,7 +118,7 @@ public Response renewLease(
         }
         // Check if we need to sync based on dirty time stamp, the client
         // instance might have changed some value
-        Response response = null;
```
PropertyOfTuckerSystems: This change is unrelated and I'd suggest to submit a separate PR for style only fixes if you're interested.

eureka-core/src/main/java/com/netflix/eureka/resources/InstanceResource.java
```diff
@@ -239,15 +239,15 @@ public Response updateMetadata(@Context UriInfo uriInfo) {
...
-                return Response.serverError().build();
+                logger.warn("Cannot find instance while updating metadata for instance {}/{}", app.getName(), id);
+                return Response.status(Status.NOT_FOUND).build();
             }
             MultivaluedMap<String, String> queryParams = uriInfo.getQueryParameters();
             Set<Entry<String, List<String>>> entrySet = queryParams.entrySet();
             Map<String, String> metadataMap = instanceInfo.getMetadata();
             // Metadata map is empty - create a new map
             if (Collections.emptyMap().getClass().equals(metadataMap.getClass())) {
-                metadataMap = new ConcurrentHashMap<String, String>();
```
PropertyOfTuckerSystems: Again would suggest to separate, but I understand these are minor so up to you.

PR #1156, opened 2018-11-14
Fix-1154

holy12345:
Change response code when instance not found we return 404 not 500

Please see https://github.com/Netflix/eureka/issues/1154

Thanks


PropertyOfTuckerSystems:
Duplicates #1155.

Issue #1174, opened 2019-01-16
Instance status keep STARTING in eureka server even it is UP on Local


addozhang:
At first, it is very difficult to reproduce. Even, this case is very rare but fatal. Currently, we met 3 times of 60+ services and 150+ instances in past two months.

> eureka-client: 1.7.2
spring-boot: 1.5.12.RELEASE
spring-cloud: Edgware.SR3

### Instance info on remote

![image](https://user-images.githubusercontent.com/2224492/51249988-16ec7d00-19d0-11e9-8f38-c9242d3228ff.png)

### Status info on local

heap dump

![image](https://user-images.githubusercontent.com/2224492/51250008-1f44b800-19d0-11e9-9588-15bfc354a4c5.png)

### Instance info in Heartbeat request

tcp dump

![image](https://user-images.githubusercontent.com/2224492/51250018-266bc600-19d0-11e9-84b9-8e12aa0cfa5d.png)

| Scope | Status | lastDirtyTimestamp | lastUpdatedTimestamp  | registrationTimestamp |
| --- | --- | --- | --- | --- |
| Request | UP | 1545039481813 |  |  |
| Local | UP | 1545039481813 | 1545039472888 |  |
| Remote | STARTING | 1545039481813 | 1545039481899 | 1545039481898 |

### InstanceInfo initialization

The initialization is via `InstanceInfoFactory#create()` and status is `STARTING`

### Instance registration

The real logic of registration is in `DiscoveryClient#register()`. But this method has two invokers. 

#### I. Main flow registration

`EurekaAutoServiceRegistration` implements `SmartLifecycle` interface, its instance is initialized in `EurekaClientAutoConfiguration#eurekaAutoServiceRegistration()`.

`EurekaAutoServiceRegistration#start()` method is invoker after all spring bean initialized and arranges `EurekaServiceRegistry` to register `EurekaRegistration`:

```java
public void start() {
    ...
    if (!this.running.get() && this.registration.getNonSecurePort() > 0) {
    	this.serviceRegistry.register(this.registration);
    	this.context.publishEvent(
    			new InstanceRegisteredEvent<>(this, this.registration.getInstanceConfig()));
    	this.running.set(true);
    }
}
```

`EurekaRegistration#register()`: it set the instance status to initial stats `UP` which can be configured by `eureka.instance.initial-status`. During status updating, stats changed and timestamp marked down in `lastDirtyTimestamp`. Once status changed, `StatusChangeListener#notify()` is triggerred 

[...]

PR #1175, opened 2019-01-17
move gain of lastDirtyTimestamp before gain of status to avoid race condition

addozhang:
refer to solution of #1174


addozhang:
delete forked repo by mistake. New one #1198 created. Close this

PR #1191, opened 2019-03-22
Fix: DiscoveryClient NullPointException

markGbug:
there possible null pointer dereference of applications in com.netflix.discovery.DiscoveryClient.fetchRegistry . when applications is null, the log param will cause null point.


troshko111:
The contract seems to be the `getApplications()` is never `null` and is not allowed to be, the code throughout the `DiscoveryClient` assumes it's not and never sets it to `null`. It's accessed without a null check in many places including the `ctor`, in fact even with this added check it makes no functional difference as it's immediately accessed without any check later in the `ctor`.

i.e.

```
    @Override
    public Applications getApplications() {
        return localRegionApps.get();
    }
```

and all `localRegionApps.set` calls check for `null` and never set it to `null`.

If you have a clear repro how this can happen without blowing up in all other places, feel free to reopen.

PR #1194, opened 2019-04-17
EUREKA-1149 Add injectable customizable EndpointRandomizer

### LINKED ISSUE #1149: ServiceUrlRandomizer based on hostname may lead to uneven distribution of clients
Back in 2013 (here 3767d56efbabc8cf5e26fc442c7b008c841bbf19) randomizer for list of available Eureka nodes was introduced. And in this implementation reordering service URLs in fact depends on host name.
This implementation is still the only one present in Eureka client. However, if number of services per host is big, but number of hosts is not, it leads to uneven distribution of load on Eureka. For example, if there are some big hosts running numerous Docker containers.

Hence, 3 questions:
1) Why does this implementation depends on host name, why not just use random seed calculated once?
2) If reason is hard to find since the implementation is 5yo, would you mind if we support random seed?
3) And if you don't mind that we support random seed, do you want us to leave hostname-dependent implementation as default one and add config parameter to enable random-based, or we may just replace hostname-dependent one with random-based?

Thanks!

mgtriffid:
This is another attempt to fix #1149. This time tried to take comments from #1151 into account.

Difficult thing here is that there are actually two randomizers across Eureka codebase. And the one that is used by DiscoveryClient ends up being just a call to static method. So I decided to not touch existing randomizer that is a field of DiscoveryClient (because I have no idea where can it be used) and introduced new field. This EndpointRandomizer is passed to ZoneAffinityClusterResolver through some chain of calls. I have left behavior unchanged, just made it customizeable.

### DIFF
--- a/eureka-client-archaius2/src/main/java/com/netflix/discovery/guice/InternalEurekaClientModule.java
+++ b/eureka-client-archaius2/src/main/java/com/netflix/discovery/guice/InternalEurekaClientModule.java
@@ -20,6 +20,8 @@ import com.netflix.discovery.DiscoveryClient;
 import com.netflix.discovery.EurekaArchaius2ClientConfig;
 import com.netflix.discovery.EurekaClient;
 import com.netflix.discovery.EurekaClientConfig;
+import com.netflix.discovery.shared.resolver.EndpointRandomizer;
+import com.netflix.discovery.shared.resolver.ResolverUtils;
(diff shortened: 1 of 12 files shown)


pparth:
Anyone taking a look at this PR ?

troshko111:
Thanks for the PR @mgtriffid!

PR #1210, opened 2019-06-04
Enhance getHostCandidates method to avoid quarantineSet never be cleared

### LINKED ISSUE #1208: RetryableEurekaHttpClient candidateHosts size may improve to avoid quarantineSet not be cleared
Assume we have a same url in two zone(some reasons, although it's not reasonable), simply the configuration file as flows:

```
eureka.client.region=region1
eureka.client.availability-zones.region1 = zone1,zone2
eureka.client.serviceUrl.zone1=http://localhost:8080/eureka
eureka.client.serviceUrl.zone2=http://localhost:8080/eureka
eureka.client.transport.retryableClientQuarantineRefreshPercentage=1
```
If eureka server is not available, then the url will be added to quarantineSet. Client will not be able to re-register to eureka server when eureka server restarted (Always throw throw exception "There is no known eureka server; cluster server list is empty").

I analyse the code of RetryableEurekaHttpClient in the eureka-client:
```
  private final Set<EurekaEndpoint> quarantineSet = new ConcurrentSkipListSet<>();

  ...

	 if (currentEndpoint != null) {
                quarantineSet.add(currentEndpoint);
         }
  ...

  private List<EurekaEndpoint> getHostCandidates() {
        List<EurekaEndpoint> candidateHosts = clusterResolver.getClusterEndpoints();
        quarantineSet.retainAll(candidateHosts);

        // If enough hosts are bad, we have no choice but start over again
        int threshold = (int) (candidateHosts.size() * transportConfig.getRetryableClientQuarantineRefreshPercentage());
        //Prevent threshold is too large
        if (threshold > candidateHosts.size()) {
            threshold = candidateHosts.size();
        }
        if (quarantineSet.isEmpty()) {
            // no-op
        } else if (quarantineSet.size() >= threshold) {
            logger.debug("Clearing quarantined list of size {}", quarantineSet.size());
            quarantineSet.clear();
        } else {
            List<EurekaEndpoint> remainingHosts = new ArrayList<>(candidateHosts.size());
            for (EurekaEndpoint endpoint : candidateHosts) {
                if (!quarantineSet.contains(endpoint)) {
                    remainingHosts.add(endpoint);
                }
            }
            candidateHosts = remainingHosts;
        }

        return candidateHosts;
  }
```

The problem is: threshold use ```candidateHosts.size()``` (Use ```List```), but only ```quarantineSet.size() >= threshold``` the quarantineSet can be cleared.
Although two zones is diferent ```AwsEndpoint```, they are the same ```EurekaEndpoint```. At this time quarantineSet size is 1, but threshold is 2, quarantineSet.size() will always smaller than threshold. QuarantineSet will never be cleared, so client will not be able to re-register to eureka server when eureka server restarted.

My solution is translate candidateHosts list to set, and then use the candidateHostsSetSize:
```
    private List<EurekaEndpoint> getHostCandidates() {
        List<EurekaEndpoint> candidateHosts = clusterResolver.getClusterEndpoints();
        quarantineSet.retainAll(candidateHosts);

	 // Use Set to remove the same EurekaEndpoint in the List to prevent threshold always larger than quarantineSet size
        int candidateHostsSetSize = (new ConcurrentSkipListSet<>(candidateHosts)).size();
        // If enough hosts are bad, we have no choice but start over again
        int threshold = (int) (candidateHostsSetSize * transportConfig.getRetryableClientQuarantineRefreshPercentage());
        //Prevent threshold is too large
        if (threshold > candidateHostsSetSize) {
            threshold = candidateHostsSetSize;
        }
	if (quarantineSet.isEmpty()) {
            // no-op
        } else if (quarantineSet.size() >= threshold) {
            logger.debug("Clearing quarantined list of size {}", quarantineSet.size());
            quarantineSet.clear();
        } 
   ...
  }
```

izdt:
Fix #1208 
Use ```ConcurrentSkipListSet```  to remove  duplicates  in the ```List<EurekaEndpoint>``` candidateHosts to avoid threshold always larger than quarantineSet size.

### DIFF
--- a/eureka-client/src/main/java/com/netflix/discovery/shared/transport/decorator/RetryableEurekaHttpClient.java
+++ b/eureka-client/src/main/java/com/netflix/discovery/shared/transport/decorator/RetryableEurekaHttpClient.java
@@ -53,6 +53,7 @@ import static com.netflix.discovery.EurekaClientNames.METRIC_TRANSPORT_PREFIX;
  *
  * @author Tomasz Bak
  * @author Li gang
+ * @author Dongtiao Zhao
  */
 public class RetryableEurekaHttpClient extends EurekaHttpClientDecorator {
 
@@ -161,12 +162,13 @@ public class RetryableEurekaHttpClient extends EurekaHttpClientDecorator {
     private List<EurekaEndpoint> getHostCandidates() {
         List<EurekaEndpoint> candidateHosts = clusterResolver.getClusterEndpoints();
         quarantineSet.retainAll(candidateHosts);
-
+        // Use Set to remove the same EurekaEndpoint in the List to prevent threshold always larger than quarantineSet size
+        int candidateHostsSetSize = (new ConcurrentSkipListSet<>(candidateHosts)).size();
         // If enough hosts are bad, we have no choice but start over again
-        int threshold = (int) (candidateHosts.size() * transportConfig.getRetryableClientQuarantineRefreshPercentage());
+        int threshold = (int) (candidateHostsSetSize * transportConfig.getRetryableClientQuarantineRefreshPercentage());
         //Prevent threshold is too large
-        if (threshold > candidateHosts.size()) {
-            threshold = candidateHosts.size();
+        if (threshold > candidateHostsSetSize) {
+            threshold = candidateHostsSetSize;
         }
         if (quarantineSet.isEmpty()) {
             // no-op


troshko111:
See the comment in the linked issue, this does not make sense from the configuration standpoint, better validation is a maybe but actual support is not an option (and I doubt it would not go hay wire in other places) anyway.

Issue #1211, opened 2019-06-06
No way to add additional HttpHeader to all peer replication calls


LittleBaiBai:
Currently, all `PeerEurekaNode` are created by a protected method on singleton `PeerEurekaNodes.createPeerEurekaNode()`, which calls a static method on `JerseyReplicationClient.createReplicationClient()` to create a `HttpReplicationClient`. I don't see an easy way to add additional ClientFilter to the client.

I see two ways to solve this issue based:
1. Add an `addReplicationClientFilter` method to the `HttpReplicationClient` interface. This way I can just override the `createPeerEurekaNode` with my additional filter when creating my new `PeerEurekaNodes`. It's not the cleanest solution, but will be a very small PR.
2. Similar to how singleton `DiscoveryClient` takes in `DiscoveryClientOptionalArgs`, singleton `PeerEurekaNodes` can take in `ReplicationClientOptionalArgs`, and pass it down to create `HttpReplicationClient`. This is definitely a cleaner and more extensive solution than the first one, but requires more testing.

Any inputs here would be very much helpful.

PR #1212, opened 2019-06-06
add ClientFilter to HttpReplicationClient

LittleBaiBai:
Connects to #1211

### DIFF
--- a/eureka-core/src/main/java/com/netflix/eureka/cluster/HttpReplicationClient.java
+++ b/eureka-core/src/main/java/com/netflix/eureka/cluster/HttpReplicationClient.java
@@ -5,7 +5,6 @@ import com.netflix.discovery.shared.transport.EurekaHttpResponse;
 import com.netflix.eureka.cluster.protocol.ReplicationList;
 import com.netflix.eureka.cluster.protocol.ReplicationListResponse;
 import com.netflix.eureka.resources.ASGResource.ASGStatus;
-import com.sun.jersey.api.client.filter.ClientFilter;
 
 /**
  * @author Tomasz Bak
@@ -15,6 +14,4 @@ public interface HttpReplicationClient extends EurekaHttpClient {
     EurekaHttpResponse<Void> statusUpdate(String asgName, ASGStatus newStatus);
 
     EurekaHttpResponse<ReplicationListResponse> submitBatchUpdates(ReplicationList replicationList);
-
-    default void addReplicationClientFilter(ClientFilter clientFilter) {};
 }
--- a/eureka-core/src/main/java/com/netflix/eureka/cluster/HttpReplicationClient.java
+++ b/eureka-core/src/main/java/com/netflix/eureka/cluster/HttpReplicationClient.java
@@ -5,6 +5,7 @@ import com.netflix.discovery.shared.transport.EurekaHttpResponse;
 import com.netflix.eureka.cluster.protocol.ReplicationList;
 import com.netflix.eureka.cluster.protocol.ReplicationListResponse;
 import com.netflix.eureka.resources.ASGResource.ASGStatus;
+import com.sun.jersey.api.client.filter.ClientFilter;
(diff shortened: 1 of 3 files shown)


eureka-core/src/main/java/com/netflix/eureka/cluster/HttpReplicationClient.java
```diff
@@ -15,4 +16,5 @@
 
     EurekaHttpResponse<ReplicationListResponse> submitBatchUpdates(ReplicationList replicationList);
 
+    default void addReplicationClientFilter(ClientFilter clientFilter) {};
```
elandau: Adding this method here would force a Jersey dependency on HttpReplicationClient.  You should be able to provide your own PeerEurekaNodes implementation similar to Jersey2PeerEurekaNodes that would add filters to the JerseyReplicationClient.

spencergibb: So no need for this change, but the one below is still valid, correct?

elandau: Given the static nature of these API's I'd say yes.  Ideally I'd rather have a proper factory and only allow adding filters at creation time.  Not sure about the thread safety guarantees when adding a filter to an existing client.  Not sure it's worth the effort to refactor that now.

LittleBaiBai: A new `PeerEurekaNodes` implementation itself isn't enough to add filters to `JerseyReplicationClient` because `JerseyReplicationClient.createReplicationClient()` returns `HttpReplicationClient`, that's why I went ahead and changed the interface. But I agree that it's not great to force a Jersey dependency on that client. 

So maybe I can try implementing the second option proposed in the issue, which is passing in additional filters on the static creation method, or use `ReplicationClientOptionalArgs` if we follow the pattern of DiscoveryClient using `DiscoveryClientOptionalArgs`.

elandau: JerseyReplicationClient.createReplicationClient() returns a JerseyReplicationClient and not HttpReplicationClient.
 https://github.com/Netflix/eureka/blob/07876a70a547ea52822c286b49c001971a1e6880/eureka-core/src/main/java/com/netflix/eureka/transport/JerseyReplicationClient.java#L135

spencergibb: We don't interact directly with that. We have a bean that extends `PeerEurekaNodes`. I guess we'd override `createPeerEurekaNode()`?

spencergibb: RE: concurrency, we would only add a client filter before eureka server starts up during the spring lifecycle.

LittleBaiBai:
@elandau Based on your comment of "only allow adding filters at creation time", I did this quick spike on how that could look like: https://github.com/Netflix/eureka/compare/master...LittleBaiBai:add-additional-filters-on-creation.

I haven't tested this yet. Do you think this is on the right (or better) direction?

eureka-core/src/main/java/com/netflix/eureka/cluster/HttpReplicationClient.java
LittleBaiBai: @elandau I updated this PR removing the `addReplicationClientFilter` method from the interface.

elandau:
@LittleBaiBai I think the current implementation (this PR) is fine.  I actually view the existing pattern of passing in an object of optional dependencies as poor api design.  That pattern came about as a hack around Guice's poor support for optional injection.  If we ever decide to refactor this code I'd prefer to deprecate all constructors and switch to the builder pattern.

PR #1224, opened 2019-07-14
make AmazonInfo deserializer more robust

brharrington:
Prior to this change, it was easy to trigger an endless loop in
`StringInterningAmazonInfoBuilder` if the `dataCenterInfo` payload
wasn't exactly right. For example, it could be done with:

```
"dataCenterInfo": {
  "@class": "com.netflix.appinfo.AmazonInfo",
  "metadata": {
    "instance-id": "i-12345"
  }
}
```

If there wasn't a field before `metadata` it would skip over the
`metadata` field and then end up in an endless loop because it
didn't check for the end of input.


troshko111:
Thanks @brharrington !

PR #1229, opened 2019-08-07
move gain of lastDirtyTimestamp before gain of status to avoid race condition

addozhang:
solution for #1174


tjuchniewicz:
Not so deep in the code but how serialization order may change anything?

tjuchniewicz:
OK. Now I get your explanation in #1174.
Make sense for me. I vote for this PR.

PR #1235, opened 2019-08-29
Uses HashMap if CompactHashMap is not available on classpath.

spencergibb:
This allows eureka users to exclude compactmap library if they so choose.


mattnelson:
Should the dependency be scoped as optional?

https://github.com/Netflix/eureka/blob/fb355976fe989b41d126d1801a0cc159951b3b38/eureka-client/build.gradle#L21

spencergibb:
@mattnelson I don't think so. I'm shooting for the defaults to stay the same. If it gets excluded, it still works.

PR #1237, opened 2019-09-06
Optimized CircularQueue by extending LinkedList with constant time size implementation

### LINKED ISSUE #1236: AbstractInstanceRegistry$CircularQueue is inneficient
I have noticed that `AbstractInstanceRegistry$CircularQueue` implementation is quite inefficient due to `ConcurrentLinkedQueue#size`, which is linear time complexity of storage size, effectively making add element O(n).

gavlyukovskiy:
Since `AbstractInstanceRegistry` already have synchronization on `recentRegisteredQueue` and `recentCanceledQueue` it's safe to change `CircularQueue` implementation to extend `LinkedList` with constant time `size()`.

Benchmark:
```
import org.openjdk.jmh.annotations.Benchmark;
import org.openjdk.jmh.annotations.BenchmarkMode;
import org.openjdk.jmh.annotations.Fork;
import org.openjdk.jmh.annotations.Level;
import org.openjdk.jmh.annotations.Measurement;
import org.openjdk.jmh.annotations.Mode;
import org.openjdk.jmh.annotations.OutputTimeUnit;
import org.openjdk.jmh.annotations.Scope;
import org.openjdk.jmh.annotations.Setup;
import org.openjdk.jmh.annotations.State;
import org.openjdk.jmh.annotations.Warmup;
import org.openjdk.jmh.runner.Runner;
import org.openjdk.jmh.runner.RunnerException;
import org.openjdk.jmh.runner.options.Options;
import org.openjdk.jmh.runner.options.OptionsBuilder;

import java.util.concurrent.TimeUnit;

@BenchmarkMode(Mode.Throughput)
@OutputTimeUnit(TimeUnit.SECONDS)
@State(Scope.Benchmark)
@Fork(value = 2)
@Warmup(iterations = 3, batchSize = 10000)
@Measurement(iterations = 3, batchSize = 10000)
public class CircularQueueBenchmark {

    private final EurekaCircularQueue<String> eureka = new EurekaCircularQueue<>(10000);
    private final EurekaCircularQueueLinkedList<String> eurekaLinkedList = new EurekaCircularQueueLinkedList<>(10000);

    @Setup(Level.Iteration)
    public void setup(){
        eureka.clear();
        eurekaLinkedList.clear();
    }

    @Benchmark
    public EurekaCircularQueue<String> eureka() {
        synchronized (eureka) {
            eureka.add("something");
        }
        return eureka;
    }

    @Benchmark
    public EurekaCircularQueueLinkedList<String> eurekaLinkedList() {
        synchronized (eurekaLinkedList) {
            eurekaLinkedList.add("something");
        }
        return eurekaLinkedList;
    }

    public static void main(String[] args) throws RunnerException {
        Options opt = new OptionsBuilder()
                .include(CircularQueueBenchmark.class.getSimpleName())
                .build();

        new Runner(opt).run();
    }
}
```
Results:
```
Benchmark                                 Mode  Cnt     Score     Error  Units
CircularQueueBenchmark.eureka            thrpt    6     3.297 ±   0.005  ops/s
CircularQueueBenchmark.eurekaLinkedList  thrpt    6  3893.782 ± 110.320  ops/s
```

fixes #1236

### DIFF
--- a/eureka-core/src/main/java/com/netflix/eureka/registry/AbstractInstanceRegistry.java
+++ b/eureka-core/src/main/java/com/netflix/eureka/registry/AbstractInstanceRegistry.java
@@ -21,18 +21,17 @@ import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.AbstractQueue;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
-import java.util.NoSuchElementException;
 import java.util.Random;
 import java.util.Set;
 import java.util.Timer;
 import java.util.TimerTask;
+import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.ConcurrentMap;
@@ -174,12 +173,8 @@ public abstract class AbstractInstanceRegistry implements InstanceRegistry {
     @Override
     public void clearRegistry() {
         overriddenInstanceStatusMap.clear();
-        synchronized (recentCanceledQueue) {
-            recentCanceledQueue.clear();
-        }
-        synchronized (recentRegisteredQueue) {
-            recentRegisteredQueue.clear();
-        }
+        recentCanceledQueue.clear();
+        recentRegisteredQueue.clear();
         recentlyChangedQueue.clear();
         registry.clear();
     }
@@ -238,11 +233,9 @@ public abstract class AbstractInstanceRegistry implements InstanceRegistry {
                 lease.setServiceUpTimestamp(existingLease.getServiceUpTimestamp());
             }
             gMap.put(registrant.getId(), lease);
-            synchronized (recentRegisteredQueue) {
-                recentRegisteredQueue.add(new Pair<Long, String>(
-                        System.currentTimeMillis(),
-                        registrant.getAppName() + "(" + registrant.getId() + ")"));
-            }
+            recentRegisteredQueue.add(new Pair<Long, String>(
+                    System.currentTimeMillis(),
+                    registrant.getAppName() + "(" + registrant.getId() + ")"));
             // This is where the initial state transfer of overridden status happens
             if (!InstanceStatus.UNKNOWN.equals(registrant.getOverriddenStatus())) {
                 logger.debug("Found overridden status {} for instance {}. Checking to see if needs to be add to the "
@@ -310,9 +303,7 @@ public abstract class AbstractInstanceRegistry implements InstanceRegistry {
             if (gMap != null) {
                 leaseToCancel = gMap.remove(id);
             }
-            synchronized (recentCanceledQueue) {
-                recentCanceledQueue.add(new Pair<Long, String>(System.currentTimeMillis(), appName + "(" + id + ")"));
-            }
+            recentCanceledQueue.add(new Pair<Long, String>(System.currentTimeMillis(), appName + "(" + id + ")"));
             InstanceStatus instanceStatus = overriddenInstanceStatusMap.remove(id);
             if (instanceStatus != null) {
                 logger.debug("Removed instance id {} from the overridden map which has value {}", id, instanceStatus.name());
@@ -1163,10 +1154,7 @@ public abstract class AbstractInstanceRegistry implements InstanceRegistry {
      */
     @Override
     public List<Pair<Long, String>> getLastNRegisteredInstances() {
-        List<Pair<Long, String>> list;
-        synchronized (recentRegisteredQueue) {
-            list = new ArrayList<Pair<Long, String>>(recentRegisteredQueue);
-        }
+        List<Pair<Long, String>> list = new ArrayList<Pair<Long, String>>(recentRegisteredQueue);
         Collections.reverse(list);
         return list;
     }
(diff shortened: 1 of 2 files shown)


eureka-core/src/main/java/com/netflix/eureka/registry/AbstractInstanceRegistry.java
```diff
@@ -1289,19 +1290,13 @@ public CircularQueue(int size) {
...
-
         }
 
         private void makeSpaceIfNotAvailable() {
             if (this.size() == size) {
                 this.remove();
             }
         }
-
-        public boolean offer(E e) {
```
troshko111: I don't think it's a good idea to remove this even though it's unused (for now), it breaks the type invariant as now one can use the base method which is unbounded. This either is a circular buffer or it is not, if it is - it must ensure all write operations are size-bounded.

gavlyukovskiy: I agree that the implementation should be consistent and cover all insert operations, main reason removing `offer` override was that `LinkedList#offer(E)` delegates to `add`, which is already overridden. I would say it's unlikely that this behavior will change.

troshko111: > `LinkedList#offer(E)` delegates to `add`.

This is an implementation detail which may change in the future JVM versions, the doc does not guarantee this [contract](https://docs.oracle.com/javase/7/docs/api/java/util/LinkedList.html#offer(E)).

Even if it did, maybe`LinkedList#offer(E)` does delegate the call, but if a year or two from now someone changes it to a non-`LinkedList` backed type which does not, this becomes a bug, I just don't see a reason here to create an opportunity for a bug.

gavlyukovskiy:
> Also ideally this should be an array-based circular buffer as both the old and the new implementations are not cache-friendly and are slow to iterate which is one of their primary uses, but it does not matter in this context much (if at all), I'm ok with this change as it makes perfect sense to me as long as you address the issues.

I was also thinking about it, and I even have working implementation, but I thought it would be too risky for you to accept it considering low performance difference (I measured only `add`) between array and linked list backed queues :)

I added commit with additional synchronization and array backed implementation. I also noticed that iteration just moves elements to array list, which is much faster when using collection as a source, especially with efficient `toArray` implementation. I made some measurements and it looks pretty impressive:
```
CircularQueueIterateBenchmark.arrayCircularQueue                             thrpt    3   14177.309 ±  291.964  ops/s
CircularQueueIterateBenchmark.arrayCircularQueueConstructor                  thrpt    3   45361.993 ± 2427.956  ops/s
CircularQueueIterateBenchmark.arrayCircularQueueConstructorOptimizedToArray  thrpt    3  264552.864 ± 5394.090  ops/s
CircularQueueIterateBenchmark.eurekaLinkedList                               thrpt    3   11591.028 ±  819.217  ops/s
```
I don't know how to create a benchmark that has proper memory layout to show higher cache hit rate for array backed implementation, but even difference between loop copy - `arrayCircularQueue` and constructor with efficient `toArray` - `arrayCircularQueueConstructorOptimizedToArray` is impressive!

eureka-core/src/main/java/com/netflix/eureka/registry/AbstractInstanceRegistry.java
gavlyukovskiy: Basically to do it 100% right, we would need to override like 10 more insert methods such as `offerLast`, `offerFirst`, `addAll`, etc. But anyway I implemented queue from `AbstractQueue`, so all insert methods are implemented and it should be ok now.

troshko111:
Sorry for the delay on this one, I'm giving this full attention now.

eureka-core/src/main/java/com/netflix/eureka/registry/AbstractInstanceRegistry.java
troshko111: Good point.

eureka-core/src/main/java/com/netflix/eureka/registry/AbstractInstanceRegistry.java
```diff
@@ -1278,29 +1280,113 @@ long getCurrentTimeNano() {  // for testing
 
     }
 
-    private class CircularQueue<E> extends ConcurrentLinkedQueue<E> {
+    /* visible for testing */  static class CircularQueue<E> extends AbstractQueue<E> {
```
troshko111: This looks correct to me but considering you're rolling out a data structure, the amount of unit tests is expected to be much larger, I don't see test cases for [0, 1, even #, odd #] capacity buffers, tests for ops when under capacity, same tests for at capacity, etc. The risk/benefit ratio is just too high.

Based on your benchmarks and iterations so far, it seems like the best way forward is to
- Changed to LinkedList-based as in [iteration 1](https://github.com/Netflix/eureka/pull/1237/commits/e08ddfa7ba2f3b1e296707c033629dbd4b760f24)
- Fix `clearRegistry` to be thread safe as you did [here](https://github.com/Netflix/eureka/pull/1237/commits/3e7a8246827a1abea80e55c9da1a7f4aa3eaa479#diff-f80bacaf7859f75ae25a290673be6b19R174)
- Pre-allocate result lists in `getLastN ` methods similar to https://github.com/Netflix/eureka/pull/1237/commits/3e7a8246827a1abea80e55c9da1a7f4aa3eaa479#diff-f80bacaf7859f75ae25a290673be6b19R174

This would be correct, about the same performance, get rid of unnecessary synchronization, require no additional testing.

Let me know what you think.

gavlyukovskiy:
@brharrington I agree with you, I checked the code and implementation is the same as mine, except locking. And I more believe Doug Lea than myself to implement and test data structures :)

Also it looks like with this approach we can remove locking on insertion as this code will be sufficient even under race:
```  
@Override
  public boolean offer(T item) {
    while (!underlying.offer(item)) {
      underlying.poll();
    }
    return true;
  }
```

@troshko111 what do you think on this approach?

troshko111:
This collection is thread-safe and based on the usage in the instance registry, all of the sync code can be removed, i.e. clear, element addition, toArray.

gavlyukovskiy:
@troshko111 I updated the PR with suggestions from @brharrington. Please review

PR #1243, opened 2019-09-19
Fix DeserializerStringCache if value transformer is used (#1242)

### LINKED ISSUE #1242: Regression in Eureka server 1.8.0 (and above)
I've just updated from Eureka 1.7.2 to the latest Eureka version and ended up with a regression due to the PR #970 
More specifcally the InstanceInfo now contains a lowercase app name even though it shouldn't.
Reason: One of my apps registering at Eureka uses a lowercase app name. But normally Eureka itself should do an "uppercase" call to have this fixed.
This is expected to happen for the deserialization at https://github.com/Netflix/eureka/blob/9d07ac4aaba90876d89a691521c128c9c59bcce8/eureka-client/src/main/java/com/netflix/discovery/converters/EurekaJacksonCodec.java#L516-L526
But the thing is: The cache already contains the string in the lowercase variant and the value transformer is not even called, resulting in the lowercase variant to be used.
My idea would be to overwrite the source of the underlying ArrayCharBuffer at https://github.com/Netflix/eureka/blob/9d07ac4aaba90876d89a691521c128c9c59bcce8/eureka-client/src/main/java/com/netflix/discovery/util/DeserializerStringCache.java#L320
by calling the value transformer to make sure that the source already has the right value (offset and length need to be set accordingly). No need to keep the value transformer anymore.
@troshko111 what do you think?

GFriedrich:
Resolves #1242

### DIFF
--- a/eureka-client/src/main/java/com/netflix/discovery/util/DeserializerStringCache.java
+++ b/eureka-client/src/main/java/com/netflix/discovery/util/DeserializerStringCache.java
@@ -285,7 +285,7 @@ public class DeserializerStringCache implements Function<String, String> {
     private interface CharBuffer {
 
         public static CharBuffer wrap(JsonParser source, Supplier<String> stringSource) throws IOException {
-            return new ArrayCharBuffer(source, stringSource);
+            return stringSource == null ? wrap(source) : wrap(stringSource.get());
         }
 
         public static CharBuffer wrap(JsonParser source) throws IOException {
@@ -303,21 +303,15 @@ public class DeserializerStringCache implements Function<String, String> {
         OfInt chars();
 
         static class ArrayCharBuffer implements CharBuffer {
-            private char[] source;
-            private int offset;
+            private final char[] source;
+            private final int offset;
             private final int length;
             private int hash;
-            private Supplier<String> valueTransform;
 
             ArrayCharBuffer(JsonParser source) throws IOException {
                 this.source = source.getTextCharacters();
                 this.offset = source.getTextOffset();
                 this.length = source.getTextLength();
-            }
-
-            ArrayCharBuffer(JsonParser source, Supplier<String> valueTransform) throws IOException {
-                this(source);
-                this.valueTransform = valueTransform;
             }
 
             @Override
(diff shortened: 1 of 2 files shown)


GFriedrich:
@tcellucci I'm aware of the issues that you've mentioned, but I think this is better than a broken cache implementation.

I've already tried to explain the bug in detail in #1242 and it's not the one that you've described but I can try again to make it more clear:
If you have a cache that looks like this:
Charbuffer{"value"} -> String{"value"}
and you get to the point where the application name is deserialized from a response that is written in lowercase and contains also the name "value", you will lookup the cache by using a Charbuffer{"value"}. The cache in turn will return the String{"value"} even though your value transformer would have created a String{"VALUE"} instead. Due to that Eureka will create a weird exception underneath complaining about the lowercase application name that was returned from the cache.

Of course you could change the value transformer to work with some input from the cache and do the transformation after the lookup, but this in turn would create a new String on each and every lookup. And as I'm not aware were this String is used later on (maybe it is kept at some different place too), I've decided against it, to not have several instances of String{"VALUE"} in the heap.

Instead I've decided to do the transformation upfront, so that the cache is able to hold a value for "value" and "VALUE". By that the String{"VALUE"} is always reused and the just created string to lookup the cache is thrown away directly, resulting in only one element more in the cache and the rest can be GC'ed right afterwards.

If you've a better solution, feel free to change the PR, but I think it's far better to have a correct and working implementation than a buggy cache that pays more focus on possible GC workload than on correctness.

tcellucci:
I think a unit test demonstrating the problem is in order, rather than more exposition.

tcellucci:
reading through your comments, it sounds as if your code is accessing the cache in different ways for the same key (this implied by the presence of a lower-case cache value in the cache, despite an uppercasing cache value transform).  More plainly, for the same key, I am inferring that sometimes a value transformer is used and sometimes it is not.  The observed behavior is that the cache always returns the initially cached value, rather than transformed value you're expecting.

Can you confirm my understanding of this issue, and if correct, explain how this situation comes about with sometimes-specified valueTransform?

GFriedrich:
@tcellucci there you go - I've created a test for this case. It fails with the old implementation but works with the new one.
This case is happening if the cache was used already by e.g. some other field of the InstanceInfo containing the value as lowercase and then trying to deserialize the app name which should be converted to uppercase but isn't.

troshko111:
@GFriedrich thanks for bringing this up! I've merged the linked PR, closing this one.

Issue #1221, opened 2019-06-22
eureka client not register  it self ofter a long time network connection problem


chenjie901:
1. after a long time network connection problem
2. client instance is removed from eureka server
3. renew thread try to send heart beat but eureka server return 404
4. sendHeartBeat thow TransportException with 404 status code
5. in the code below, try to re-register client but cause step 4 throw exception, so the re-register logic not woked
```
```
boolean renew() {
        EurekaHttpResponse<InstanceInfo> httpResponse;
        try {
            httpResponse = eurekaTransport.registrationClient.sendHeartBeat(instanceInfo.getAppName(), instanceInfo.getId(), instanceInfo, null);
            logger.debug(PREFIX + "{} - Heartbeat status: {}", appPathIdentifier, httpResponse.getStatusCode());
            if (httpResponse.getStatusCode() == Status.NOT_FOUND.getStatusCode()) {
                REREGISTER_COUNTER.increment();
                logger.info(PREFIX + "{} - Re-registering apps/{}", appPathIdentifier, instanceInfo.getAppName());
                long timestamp = instanceInfo.setIsDirtyWithTime();
                boolean success = register();
                if (success) {
                    instanceInfo.unsetIsDirty(timestamp);
                }
                return success;
            }
            return httpResponse.getStatusCode() == Status.OK.getStatusCode();
        } catch (Throwable e) {
            logger.error(PREFIX + "{} - was unable to send heartbeat!", appPathIdentifier, e);
            return false;
        }
    }
```
6. Eureka client always retry to send heartbeat with no re-register

troshko111:
It's pretty ridiculous a `404` response from the server would be a `TransportException` as the transport is perfectly fine and this is an application protocol code but if this is the case then it's a bug as the code you included clearly assumes it will not result in an exception. Open to PRs.

PR #1249, opened 2019-10-08
fix eureka client not register it self ofter a long time network conn…

skyguard1:
In this issue (https://github.com/Netflix/eureka/issues/1221), eureka client try to re - register but throw TransportException, so the re - register logic not executed, in my opinion, throw ThranportException show that server-side processing logic errors, Therefore, I personally think that DiscoveryClient should re-register when receiving TransportException. Please review the pull request and make Suggestions. I will track this problem, thank you


troshko111:
General:

- We can only re-register if we received a 404, no other error as this explicitly indicates the instance has been evicted, see the linked issue for an example. You can't just re-register on any error.

- Needs tests to demonstrate the issue and the fix.

- Exception handling looks incorrect as both renew and doRegister swallow exceptions.

PR #1271, opened 2020-01-20
decrement expectedNumberOfClientsSendingRenews for expired leases

### LINKED ISSUE #1266: Where to update the "expectedNumberOfClientsSendingRenews" when we evict a instance?
When I look at eureka has been a source of confusion in "selfPreservationMode", I saw the sorce code with mater branch,
when we see register or cancel method, there will be update "expectedNumberOfClientsSendingRenews", just like the code:

```java
// The lease does not exist and hence it is a new registration
synchronized (lock) {
    if (this.expectedNumberOfClientsSendingRenews > 0) {
        // Since the client wants to register it, increase the number of clients sending renews
        this.expectedNumberOfClientsSendingRenews = this.expectedNumberOfClientsSendingRenews + 1;
        updateRenewsPerMinThreshold();
    }
}
```

i have read the code with "evict" method in "AbstractInstanceRegistry" class, but i didn't find where to update the "expectedNumberOfClientsSendingRenews" when we evict a instance according to a timed task. 

can you tell me where to update the "expectedNumberOfClientsSendingRenews" when we evict a instance ?

snowp:
Moves the logic that decrements the expected number of clients into
`internalCancel` - this should ensure that we're calling it both for
explicit cancels as well as well as when the instance lease expires.

Fixes #1266

### DIFF
--- a/eureka-core/src/main/java/com/netflix/eureka/registry/AbstractInstanceRegistry.java
+++ b/eureka-core/src/main/java/com/netflix/eureka/registry/AbstractInstanceRegistry.java
@@ -326,11 +326,20 @@ public abstract class AbstractInstanceRegistry implements InstanceRegistry {
                 }
                 invalidateCache(appName, vip, svip);
                 logger.info("Cancelled instance {}/{} (replication={})", appName, id, isReplication);
-                return true;
             }
         } finally {
             read.unlock();
         }
+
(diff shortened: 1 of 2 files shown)


snowp:
I couldn't find any existing test coverage for this, happy to add some if ya'll have any pointers to where to start

PR #1275, opened 2020-01-31
Broadcast public ipv4 addresses on override

argha-c:
This is a quick and dirty PR to get Eureka to return public ipv4 addresses when overriden by a new config flag `broadcastPublicIpv4`.

However, it doesn't fundamentally alter the behavior of `resolveDefaultAddress` as part of this change. 

/cc @drobertduke

### DIFF
--- a/eureka-client/src/main/java/com/netflix/appinfo/CloudInstanceConfig.java
+++ b/eureka-client/src/main/java/com/netflix/appinfo/CloudInstanceConfig.java
@@ -128,8 +128,17 @@ public class CloudInstanceConfig extends PropertiesInstanceConfig implements Ref
 
     @Override
     public String getIpAddress() {
-        String ipAddr = amazonInfoHolder.get().get(MetaDataKey.localIpv4);
-        return ipAddr == null ? super.getIpAddress() : ipAddr;
+        String publicIpv4Addr = amazonInfoHolder.get().get(MetaDataKey.publicIpv4);
+        return this.shouldBroadcastPublicIpv4Addr() ?  getPublicIpv4Addr() : getPrivateIpv4Addr();
+    }
+
+    private String getPrivateIpv4Addr() {
+        String privateIpv4Addr = amazonInfoHolder.get().get(MetaDataKey.localIpv4);
+        return privateIpv4Addr == null ? super.getIpAddress() : privateIpv4Addr;
+    }
+    private String getPublicIpv4Addr() {
+        String publicIpv4Addr = amazonInfoHolder.get().get(MetaDataKey.publicIpv4);
+        return publicIpv4Addr == null ? super.getIpAddress() : publicIpv4Addr;
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


eureka-client/src/main/java/com/netflix/appinfo/EurekaInstanceConfig.java
```diff
@@ -221,6 +221,13 @@
      */
     String getIpAddress();
 
+
+    /**
+     * Indicates if the public ipv4 address of the instance should be advertised.
+     * @return true if the public ipv4 address of the instance should be advertised, false otherwise .
+     */
+    boolean shouldBroadcastPublicIpv4Addr();
```
spencergibb: Does this need to go in a public api? Looks like it could be in PropertiesInstanceConfig and used by CloudInstanceConfig without being in the interface.

argha-c: Yes, I was planning to throw in a default implementation as I realized it breaks the interface, but I could keep it to `PropertiesInstanceConfig` which was my initial plan. LMK what you prefer.

argha-c: @spencergibb PTAL

spencergibb: What does ptal mean?

argha-c: Please Take A(nother) Look

eureka-client/src/main/java/com/netflix/appinfo/CloudInstanceConfig.java
```diff
@@ -128,8 +128,17 @@ public String getHostName(boolean refresh) {
 
     @Override
     public String getIpAddress() {
-        String ipAddr = amazonInfoHolder.get().get(MetaDataKey.localIpv4);
-        return ipAddr == null ? super.getIpAddress() : ipAddr;
+        String publicIpv4Addr = amazonInfoHolder.get().get(MetaDataKey.publicIpv4);
```
troshko111: Dead code.

eureka-client/src/main/java/com/netflix/appinfo/PropertiesInstanceConfig.java
```diff
@@ -294,6 +294,14 @@ public String getSecureHealthCheckUrl() {
         return result == null ? new String[0] : result.split(",");
     }
 
+    /**
+     * Indicates if the public ipv4 address of the instance should be advertised.
+     * @return true if the public ipv4 address of the instance should be advertised, false otherwise .
+     */
+    public boolean shouldBroadcastPublicIpv4Addr() {
+        return configInstance.getBooleanProperty(namespace + ADVERTISE_PUBLIC_IPV4_ADDR, super.shouldBroadcastPublicIpv4Addr()).get();
```
troshko111: _KEY to stay consistent.

PR #1282, opened 2020-03-31
Add stats for DiscoveryClient

yobennett:
We introduce a nested class for `DiscoveryClient` to make it easier for users to read helpful attributes that can assist in log analysis and debugging. Most of these attributes are captured by existing metrics:

* `initTimestampMs`: timestamp when the client was initialized
* `localRegistrySize`: number of instances for all applications
* `lastSuccessfulRegistryFetchTimestampMs`: timestamp for last successful fetch
* `lastSuccessfulHeartbeatTimestampMs`: timestamp for last successful heartbeat

We add a new attribute called `initLocalRegistrySize`, which is the number of instances for all applications read when the client was initialized. We include a helper method that uses this attribute so that users can determine whether the client's initial fetch of registry information succeeded or failed. Note that other accessors are suffixed with `Ms` (for milliseconds) to denote the time unit.

Additionally, we update the existing `registrySize` with the value from `initLocalRegistrySize` during initialization instead of waiting until the first local registry refresh.

### DIFF
--- a/eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
+++ b/eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
@@ -189,6 +189,9 @@ public class DiscoveryClient implements EurekaClient {
     protected final EurekaTransportConfig transportConfig;
 
     private final long initTimestampMs;
+    private final int initRegistrySize;
+
+    private final Stats stats = new Stats();
 
     private static final class EurekaTransport {
         private ClosableResolver bootstrapResolver;
@@ -381,8 +384,10 @@ public class DiscoveryClient implements EurekaClient {
             DiscoveryManager.getInstance().setEurekaClientConfig(config);
 
             initTimestampMs = System.currentTimeMillis();
+            initRegistrySize = this.getApplications().size();
+            registrySize = initRegistrySize;
             logger.info("Discovery Client initialized at timestamp {} with initial instances count: {}",
-                    initTimestampMs, this.getApplications().size());
+                    initTimestampMs, initRegistrySize);
 
             return;  // no need to setup up an network tasks and we are done
         }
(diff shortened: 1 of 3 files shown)


eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
```diff
@@ -1743,4 +1750,53 @@ private long computeStalenessMonitorDelay(long delay) {
...
+     */
+    public class Stats {
+
+        private Stats() {}
+
+        public int initLocalRegistrySize() {
+            return initRegistrySize;
+        }
+
+        public long initTimestamp() {
```
troshko111, line 1774: nit: initTimestampMs() to disambiguate the unit.

eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
```diff
@@ -1743,4 +1750,53 @@ private long computeStalenessMonitorDelay(long delay) {
...
+            return lastSuccessfulHeartbeatTimestamp;
+        }
+
+        /**
+         * Used to determine if the Discovery client's first attempt to fetch from the service registry succeeded with
+         * non-empty results.
+         *
+         * @return true if succeeded, failed otherwise
+         */
+        public boolean isInitSucceeded() {
```
troshko111, line 1796: nit: drop `is`, just `initSucceeded()`.

eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
yobennett: This would make it consistent with the private field. Do you think we should also use `lastSuccessfulRegistryFetchTimestampMs` and `lastSuccessfulHeartbeatTimestampMs` to be consistent? We could do that with stats and leave the private fields as-is.

troshko111: Yes good point, agree with you on both accounts, rename the stat ones to include the unit, leave the fields (or you can rename those as well if you prefer so).

yobennett: Done. Left the private fields as-is.

PR #1283, opened 2020-04-04
Update unit tests

melochin:
update ``InstanceInfoReplicatorTest`` and ``RateLimiterTest``
1. remove staled mock code in ``InstanceInfoReplicatorTest.setup()``
2. change the loop condition in ``RateLimiterTest.testEvenLoad()``. Otherwise, the loop body will not be executed.


troshko111:
Thanks!

PR #1284, opened 2020-04-07
upgrade jackson to - 2.9.10

RakefetEP:
(for some security fixes)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -37,7 +37,7 @@ allprojects {
         servoVersion = '0.12.21'
         governatorVersion = '1.17.5'
         archaiusVersion = '0.7.6'
-        jacksonVersion = '2.9.4'
+        jacksonVersion = '2.10.3'
         woodstoxVersion = '5.2.1'
 
         // test deps
--- a/build.gradle
+++ b/build.gradle
@@ -37,7 +37,7 @@ allprojects {
         servoVersion = '0.12.21'
         governatorVersion = '1.17.5'
         archaiusVersion = '0.7.6'
-        jacksonVersion = '2.10.3'
+        jacksonVersion = '2.9.10.3'
         woodstoxVersion = '5.2.1'
 
         // test deps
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -37,7 +37,7 @@ allprojects {
         servoVersion = '0.12.21'
         governatorVersion = '1.17.5'
         archaiusVersion = '0.7.6'
-        jacksonVersion = '2.9.4'
+        jacksonVersion = '2.10.3'
```
troshko111, line 40: Seems like `2.9.10` include the same security fixes, I don't mind jumping further but would probably want it separately to make sure nothing breaks (it's 2 years worth of changes).

RakefetEP: Thanks for the fast reply!

I saw 2 vulnerabilities in 2.9.10:
https://github.com/advisories/GHSA-fmmc-742q-jg75 
https://github.com/advisories/GHSA-mx7p-6679-8g3q

If you prefer to do it in steps, maybe we should upgrade to 2.9.10.1?
(Maybe there are more in later releases but these are the ones found in a quick search)

troshko111: Let's do it in 2 steps if you don't mind, the latest 2.9.x and then try out `2.10.x`?

troshko111: `2.9.10.3` seems like a safe bet?

RakefetEP: Sounds Good :)

I'll upgrade to 2.9.10.3 first.

Thanks! :)

RakefetEP:
I see that jackson-databind has 2.9.10.3 but jackson-core does not have it and there is only one variable for Jackson libraries.

I will change to 2.9.10 for now.

RakefetEP:
Thanks, @troshko111 !

Have a lovely day :)

troshko111:
Thanks for the PR! I'll release this now.

RakefetEP:
Great! Thanks a lot!

RakefetEP:
Hi @troshko111 :) Regarding the upgrade to latest Jackson 2.10.3, are there any special steps I should do before creating the PR?

Thanks again!

troshko111:
Hey, I went over the changelog and don't really see anything there which could break compatibility. I think it'd be a good idea to compare the `JSON` representation of various resources (instance info, app list, etc.) pre- and post-change to ensure there are no serialization differences.

PR #1287, opened 2020-04-26
resolve endpoint zone from metadata

MI-cool:
endpoint zone can be get from metadata

### DIFF
--- a/eureka-client/src/main/java/com/netflix/discovery/shared/resolver/ResolverUtils.java
+++ b/eureka-client/src/main/java/com/netflix/discovery/shared/resolver/ResolverUtils.java
@@ -120,6 +120,8 @@ public final class ResolverUtils {
         DataCenterInfo dataCenterInfo = instanceInfo.getDataCenterInfo();
         if (dataCenterInfo instanceof AmazonInfo) {
             zone = ((AmazonInfo) dataCenterInfo).get(AmazonInfo.MetaDataKey.availabilityZone);
+        } else {
+            zone = instanceInfo.getMetadata().get("zone");
         }
 
         String networkAddress;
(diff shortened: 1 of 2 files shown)


troshko111:
The change LGTM, but can you elaborate on the bug and add some tests showing how it fixes it?

MI-cool:
Sorry.
I use the following configuration to make the client dynamically aware of the server nodes. 

eureka.client.transport.bootstrapResolverStrategy=composite
eureka.client.transport.readClusterVip=eureka-read-server
eureka.client.transport.writeClusterVip=eureka-write-server

During the test, it was found that the client lost the server zone information.

MI-cool:
I will try to add some tests.

MI-cool:
Test was done.

troshko111:
Sorry for the delay, this fell of my attention, I promise to have a short turnaround now.
I looked around the code and seems like this is Aws-specific path (see the return type) so it's expected to not to have the zone info on `MyDatacenterInfo`, there's no contract saying there should be any zone there in the meta. What issues it this causing?

MI-cool:
My eureka server was build with spring-cloud-starter-netflix-eureka-server. On server side, the framework build instances use [dataCenterInfo](https://github.com/spring-cloud/spring-cloud-netflix/blob/188b9c112c2bcad9a4474e3e4a49191bb1087195/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/EurekaInstanceConfigBean.java#L159) type of `MyDataCenterInfo` as default.
My eureka client was build with spring-cloud-starter-netflix-eureka-client. I use `eureka.client.transport.bootstrapResolverStrategy=composite` to make the client dynamically aware of the server nodes. Here comes the problem, since [newBootstrapResolver](https://github.com/Netflix/eureka/blob/2596a054e348def518799b98a32db61c4f3eab48/eureka-client/src/main/java/com/netflix/discovery/shared/transport/EurekaHttpClients.java#L106) lost `zone` of server instances, eureka clients can't be zone affinity to server instances.

troshko111:
Yeah I think this is reasonable, the current AWS/non-AWS separation is weak so I don't see a better place to do this tbh. Thanks and sorry for the delay.

MI-cool:
No worries.

PR #1288, opened 2020-04-30
Checks for existence of ConfigurationManager.

spencergibb:
This is useful if a project has excluded Archaius 1 such as spring-cloud-netflix. This is a simple workaround.


spencergibb:
My current workaround https://github.com/spring-cloud/spring-cloud-netflix/pull/3746

PR #1293, opened 2020-05-08
Remove uses of 'provided' in the eureka client

drobertduke:
After updating gradle, we can stop using the deprecated 'provided' configuration.
This fixes an issue where the jackson-dataformat-xml and woodstox-core dependencies are affecting projects consuming eureka-client.

### DIFF
--- a/eureka-client/build.gradle
+++ b/eureka-client/build.gradle
@@ -1,5 +1,4 @@
 apply plugin: 'nebula.test-jar'
-apply plugin: 'nebula.provided-base'
 
 configurations.all {
     // jersey2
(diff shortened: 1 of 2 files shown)


troshko111:
What about the `providedCompile` (2 hits), should those go to?

kilink:
Nitpick, but maybe the wording of this comment should be updated as well: https://github.com/Netflix/eureka/blob/18e9f4a43f7156a5635edecc4cedba6f81346fa5/eureka-core/build.gradle#L13

drobertduke:
> What about the `providedCompile` (2 hits), should those go to?

providedCompile seems to be still used to exclude dependencies from War files https://github.com/gradle/gradle/issues/1171

PR #1297, opened 2020-05-28
Add ability to gate InstanceStatus updates

yobennett:
We add a way for users to define gates when updating their InstanceStatus that is irrespective of the healthcheck handler response. For example, if the healthcheck handler returns an InstanceStatus of `UP` but a gate disallows that status then the healthcheck status will be overridden and the new status will be set to `DOWN`.

If there are no gates defined, the default behavior of the discovery client is unchanged.

### DIFF
--- a/eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
+++ b/eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
@@ -193,6 +193,8 @@ public class DiscoveryClient implements EurekaClient {
 
     private final Stats stats = new Stats();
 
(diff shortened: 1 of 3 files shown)


qiangdavidliu:
I think instead of doing this in the client, you may just want to implement a custom InstanceStatusMapper. Ref: https://github.com/Netflix/eureka/blob/master/eureka-client/src/test/java/com/netflix/appinfo/ApplicationInfoManagerTest.java#L82

yobennett:
@qiangdavidliu Thanks. I hadn't considered using an `InstanceStatusMapper` to address these use cases. Functionally this seems to be comparable to the proposed gates. I'm going to review the differences to weigh the benefits of each approach.

yobennett:
Closing in favor of #1307

PR #1299, opened 2020-06-02
Modify iteration Map Mode

qq1036916439:
Modify iteration Map Mode。I think the performance through entrySet Iterate Map is better.


qq1036916439:
@troshko111 This is my first time participating in an open source project. It is a pity that I did not use the email bound to github when I submitted, but used the email of the company by mistake. I don't know if there is a remedy. I want to see my name in the Contributions.

troshko111:
I have reverted it, please re-submit using the correct git user info.

PR #1301, opened 2020-06-02
Modify iteration Map function

qq1036916439:
Modify iteration Map Mode.I think the performance through entrySet Iterate Map is better.
Thank you!


qq1036916439:
I re-submit.please re-merge.thanks.

PR #1304, opened 2020-06-10
Travis CI: Trusty is EOL

cclauss:


troshko111:
Seems to be failing, also unsure why use JDK9?

cclauss:
Used JDK9 because JDK8 is no longer available.

brharrington:
You should be able to switch to `openjdk8`, it is just the Oracle distribution that is no longer available.

troshko111:
Thanks!

PR #1307, opened 2020-06-16
Add shouldEnforceFetchRegistryAtInit config

yobennett:
We add a `shouldEnforceFetchRegistryAtInit` configuration to support users that want the Eureka client to fail fast during initialization. If `shouldEnforceFetchRegistryAtInit` is set to `true` then the client will throw an `IllegalStateException` during instance construction if its initial attempt to fetch of information from the remote Eureka servers is unsuccessful. This setting does not have an effect if `shouldFetchRegistry` is set to `false`.


eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
```diff
@@ -435,8 +435,19 @@ public synchronized BackupRegistry get() {
             throw new RuntimeException("Failed to initialize DiscoveryClient!", e);
         }
 
-        if (clientConfig.shouldFetchRegistry() && !fetchRegistry(false)) {
-            fetchRegistryFromBackup();
+        if (clientConfig.shouldFetchRegistry()) {
```
troshko111, line 438: I think the logic here is off, you want to
if (shouldFetch) {
  fetch
  if (unsuccessful) {
    log at info that the normal fetch failed
    fetch from backup
    if still unsuccessful, log a detailed msg and throw
  }
}

Right now you're fetching and then fetching again from backup on success.

yobennett: Good catch. I revisited this logic and tried to clean it up. The changes should now require that fetching from both the primary and backup registry fail before applying the `shouldEnforceFetchRegistryAtInit` configuration.

PR #1312, opened 2020-06-27
remove null check code

Hanope:
I think it is unnecessary to check the null of `instanceStatus`.
Because it doesn't matter if `instanceStatus` is null when calling `overrideStatus.equals`


troshko111:
You're correct from the logical perspective and the change will be equivalent, but it looks to me the intent of the check was readability(?), it reads "if no override or unequal override", without the null check it makes you think harder what the intent is.

PR #1314, opened 2020-06-27
fix typos

Hanope:


eureka-core/src/main/java/com/netflix/eureka/registry/InstanceRegistry.java
```diff
@@ -80,7 +80,7 @@ boolean deleteStatusOverride(String appName, String id, InstanceStatus newStatus
 
     int getNumOfRenewsPerMinThreshold();
 
-    int isBelowRenewThresold();
+    int isBelowRenewThreshold();
```
troshko111, line 83: This cannot be renamed as it's a part of a public contract (an interface method of a public type).

spencergibb:
Maybe a new method that defaults to calling the misspelling?

troshko111:
> Maybe a new method that defaults to calling the misspelling?

I normally see these corrected in the doc string / comment only, but either way works.

Hanope:
> LGTM, but please revert the public interface method rename.

Okay, I revert files.

PR #1317, opened 2020-06-29
Unbreak EurekaClientConfig interface compat

troshko111:
EurekaClientConfig is treated as stable backward-compat API as of couple of years ago, update it to preserve backward-compat.


troshko111:
This un-does breaking contract change introduced by https://github.com/Netflix/eureka/commit/6438fba6a42a6d2a1dd2e4cb48eab66c34bd0a21

tauparticle:
LGTM

osi:
thanks for the quick fix!

PR #1319, opened 2020-06-30
async anonymous class change to implements Runnable interface class.

qq1036916439:
better read


troshko111:
This looks like a readability and API regression to me:

- Introduces a public type which is not a contract for no apparent reason
- Requires to jump to a different file to read the logic

qq1036916439:
Can I use lambda or private inner classes?

troshko111:
Yes, don't see why not, it's a negligible change though, but legit.

qq1036916439:
I think that inner class should not call outer class in this way.
    `InstanceInfoReplicator.this.run(); `
 I have never seen such a way before.
I am looking at the code of Eureka and hope to provide some help for Eureka. Although only trivial.

PR #1325, opened 2020-08-19
Add Muse configuration file

nescohen:
This PR adds a configuration file for the Muse platform to run on Eureka. I was able to fix a few issues Muse found and my changes were recently accepted. I thought that the Eureka project would benefit from Muse analysis.

In addition to this configuration, a maintainer should install this app (free forever for open source): https://github.com/marketplace/muse-dev

This will enable Muse to scan incoming pull requests and report potentially introduced bugs into the diff. Please feel free to reach out if there is anything else I can help with 😄

### DIFF
--- a/.muse.toml
+++ b/.muse.toml
@@ -1,0 +1,2 @@
+
+tools = [ "infer", "errorprone", "findsecbugs" ]


troshko111:
Talked offline, we won't be enabling Muse at this very point, but looking forward to experimenting with it more in the future.

PR #1327, opened 2020-08-24
Downgrade log level of noisy log messages

jamesmulcahy:


spencergibb:
Care to explain why you think they should be info and not error? They're almost all in a catch block.

eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
```diff
@@ -856,9 +856,9 @@ public Applications getApplications(String serviceUrl) {
                 logger.debug(PREFIX + "{} -  refresh status: {}", appPathIdentifier, response.getStatusCode());
                 return response.getEntity();
             }
-            logger.error(PREFIX + "{} - was unable to refresh its cache! status = {}", appPathIdentifier, response.getStatusCode());
+            logger.info(PREFIX + "{} - was unable to refresh its cache! status = {}", appPathIdentifier, response.getStatusCode());
```
troshko111, line 859: I'd also improve this log msg, e.g. it'd be useful if it stated that this is a periodic background process and that it will be retried in X seconds. Additionally, it may be valuable to preserve the exception details in some form.

eureka-client/src/main/java/com/netflix/discovery/shared/transport/decorator/RedirectingEurekaHttpClient.java
```diff
@@ -80,15 +80,15 @@ public void shutdown() {
                 TransportUtils.shutdown(delegateRef.getAndSet(currentEurekaClientRef.get()));
                 return response;
             } catch (Exception e) {
-                logger.error("Request execution error. endpoint={}", serviceEndpoint, e);
+                logger.info("Request execution error. endpoint={}", serviceEndpoint);
```
troshko111, line 83: Same, exception details and we can elaborate on the details.

spencergibb:
I'll go ahead and hide my comment in shame :grimacing:

troshko111:
@spencergibb in the app refresh case the error is handled and hence is not an error any more by definition, it's simply noise. There's nothing the users can do or should do, it's non-actionable and frankly unimportant if this happens infrequently. If it's permanent, it will manifest in actual issues (stale/empty registry) which must be detected in other ways anyway.

The redirecting client is similar, I struggle to see what a developer may do when seeing this?

What are your thoughts?

spencergibb:
Sounds good to me.

troshko111:
@spencergibb curious if you have received / seen feedback on this from users as well, we can open a separate issue / PR if we need to. The logging game in this client can use some love imo.

eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
jamesmulcahy: @troshko111 Can you elaborate on preserve the exception details? I enquired about capturing the stack track, but we agreed that we wouldn't.  The log already captures the exception's detail message string (oddly labelled as "status=", but it is there).  

Good idea on the refresh interval; I've added that to my local branch and will push once I've got clarification on the above.

eureka-client/src/main/java/com/netflix/discovery/shared/transport/decorator/RedirectingEurekaHttpClient.java
jamesmulcahy: I realize in this case, we *are* losing the exception message, so I'll ensure I add it here.

eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
troshko111: I'm inclined to believe I was wrong, I'm trying to find some sample history to have something to look at, but iirc the actual reason would be couple frames down so printing the msg only may be insufficient and having more stack info would be best.

jamesmulcahy:
@troshko111 Updated to address all comments, PTAL when convenient.

jamesmulcahy:
@troshko111 Ack - I wrapped the lines and re-pushed

PR #1328, opened 2020-08-25
Log status transitions to DOWN at error level, instead of warn

jamesmulcahy:
I also dropped the transitions out of down to info, rather than warn, since
there isn't really anything to be concerned about on such a transition

### DIFF
--- a/eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
+++ b/eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
@@ -1349,10 +1349,8 @@ public class DiscoveryClient implements EurekaClient {
 
                 @Override
                 public void notify(StatusChangeEvent statusChangeEvent) {
-                    if (InstanceStatus.DOWN == statusChangeEvent.getStatus() ||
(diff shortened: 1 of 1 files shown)


eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
```diff
@@ -1349,10 +1349,8 @@ public String getId() {
 
                 @Override
                 public void notify(StatusChangeEvent statusChangeEvent) {
-                    if (InstanceStatus.DOWN == statusChangeEvent.getStatus() ||
-                            InstanceStatus.DOWN == statusChangeEvent.getPreviousStatus()) {
-                        // log at warn level if DOWN was involved
-                        logger.warn("Saw local status change event {}", statusChangeEvent);
+                    if (InstanceStatus.DOWN == statusChangeEvent.getStatus()) {
```
troshko111, line 1352: Do you mind also changing the this Yoda condition?

jamesmulcahy: Done

PR #1337, opened 2020-09-18
Upgrade xstream to fix illegal reflection access reported by tomcat

jamesmulcahy:


jamesmulcahy:
While running locally, I noticed tomcat complain about an illegal reflective access.  Upgrading xstream led to that error going away.

troshko111 (approved):

PR #1339, opened 2020-09-18
Add metrics for instances, tagged with VIP, Status and Expiry

jamesmulcahy:


jamesmulcahy:
Sharing for visibility. I haven't tested this thoroughly yet, but I think it should work (TM)

PR #1343, opened 2020-09-28
Remote fetch metrics

jamesmulcahy:


eureka-core/src/main/java/com/netflix/eureka/registry/PeerAwareInstanceRegistryImpl.java
```diff
@@ -478,6 +486,11 @@ public boolean isLeaseExpirationEnabled() {
         return numberOfRenewsPerMinThreshold > 0 && getNumOfRenewsInLastMin() > numberOfRenewsPerMinThreshold;
     }
 
+    @com.netflix.servo.annotations.Monitor(name = METRIC_REGISTRY_PREFIX + "isLeaseExpirationEnabled", type = DataSourceType.GAUGE)
```
spencergibb, line 489: Is servo a required dependency for eureka client?

jamesmulcahy: Yeah, it has been for some time.

```
$ git blame ./eureka-client/build.gradle | grep servo
f284e3e01 (David Liu          2015-07-24 13:33:32 -0700 13)     compile "com.netflix.servo:servo-core:${servoVersion}"
```

eureka-core/src/main/java/com/netflix/eureka/registry/RemoteRegionRegistry.java
```diff
@@ -388,6 +392,7 @@ private Applications fetchRemoteRegistry(boolean delta) {
                 int httpStatus = httpResponse.getStatusCode();
                 if (httpStatus >= 200 && httpStatus < 300) {
                     logger.debug("Got the data successfully : {}", httpStatus);
+                    timeOfLastSuccessfulRemoteFetch = System.currentTimeMillis();
```
troshko111, line 395: You also want to update it on line 175 or inside of `private boolean fetchRegistry()` which gets periodically executed.

troshko111 (approved):
See the comment, otherwise LGTM!

eureka-core/src/main/java/com/netflix/eureka/registry/RemoteRegionRegistry.java
troshko111: Sorry let me load this in VSCode, this GH PR view is really poor, I misunderstood the place due to collapsed file.

jamesmulcahy: OK! I just added the code you suggested, but I'm realizing it ends up in the same place.  Perhaps it makes sense to do it once at the higher level though, rather than at the lower level where I did it originally.

troshko111: OK, viewed this locally, so to me it seems like capturing at the most top level (executor or its core `fetchRegistry`) would be beneficial, as currently this is network-level meaning if the deltas won't match we may update it anyway?

jamesmulcahy: 100% agree -- just pushed a new version; LMK what you think?

troshko111 (approved):

eureka-core/src/main/java/com/netflix/eureka/registry/RemoteRegionRegistry.java
troshko111: Yep, all good 👍

jamesmulcahy: Thanks 👍

PR #1355, opened 2020-10-15
Bump default polling interval

fiibbb:
Bump default polling interval from 30 seconds to 45 seconds.


mattnelson:
This is a non passive change. Can you change it via config instead of changing the default?

fiibbb:
Meant to do https://github.com/Netflix/eureka/pull/1356 instead.

PR #1356, opened 2020-10-16
Bump default connection idle timeout

fiibbb:
Bump default connection idle timeout to 45 seconds.


spencergibb:
Again, why change the default when you can change it via configuration?

troshko111:
@spencergibb 
> Again, why change the default when you can change it via configuration?
I think it makes sense to change the default in this particular case as the default was improperly chosen. The default polling interval is `30s` and the client configures to close its socket on `30s` of inactivity which basically creates a race and may cause it unnecessary bounce among Eureka servers making keep alive useless. WDYT?

spencergibb:
Sounds good

troshko111 (approved):

mattnelson:
> The default polling interval is `30s` and the client configures to close its socket on `30s` of inactivity which basically creates a race and may cause it unnecessary bounce among Eureka servers making keep alive useless. WDYT?

Should eureka log a message if this configuration is detected?

PR #1361, opened 2020-11-05
Make instance config factory support custom amazon info provider

fiibbb:
This diff adds `CustomAmazonInfoProviderInstanceConfigFactory`. This allow us to inject custom implementation of `<Provider>AmazonInfo`. The intention is to bind this new factory implementation to the `EurekaInstanceConfigFactory` interface, as documented [here](https://github.com/Netflix/eureka/blob/master/eureka-client-archaius2/src/main/java/com/netflix/discovery/guice/EurekaClientModule.java#L22), so that users can control how `AmazonInfo` is provided.

### DIFF
--- a/eureka-client-archaius2/src/main/java/com/netflix/appinfo/Ec2EurekaArchaius2InstanceConfig.java
+++ b/eureka-client-archaius2/src/main/java/com/netflix/appinfo/Ec2EurekaArchaius2InstanceConfig.java
@@ -1,6 +1,7 @@
 package com.netflix.appinfo;
 
 import javax.inject.Inject;
+import javax.inject.Provider;
 import javax.inject.Singleton;
 
 import com.netflix.discovery.CommonConstants;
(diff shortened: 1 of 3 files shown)


troshko111 (approved):

eureka-client-archaius2/src/main/java/com/netflix/appinfo/Ec2EurekaArchaius2InstanceConfig.java
```diff
@@ -40,6 +41,12 @@ public Ec2EurekaArchaius2InstanceConfig(Config configInstance, AmazonInfoConfig
         this(configInstance, new Archaius2AmazonInfoConfig(configInstance), CommonConstants.DEFAULT_CONFIG_NAMESPACE, info, false);
     }
 
+    public Ec2EurekaArchaius2InstanceConfig(Config configInstance, Provider<AmazonInfo> amazonInfoProvider) {
```
troshko111, line 44: Opening brace on the same line.

PR #1363, opened 2020-11-10
Fix GzipEncodingEnforcingFilter getHeaders always with a const value

kebe7jun:
HttpHeaders.ACCEPT_ENCODING only need add to header names, not values.


troshko111:
Thanks, this looks good to me, but can you please add a test demonstrating the issue and the fix?

kebe7jun:
hello @troshko111 , I have added tests for this case.

troshko111 (approved):

troshko111:
Thanks!

PR #1376, opened 2021-01-13
Make eureka instance config namespace configurable

fiibbb:
Make eureka instance config namespace configurable. This allow us to use `CustomAmazonInfoProviderInstanceConfigFactory` the same way as `CompositeInstanceConfigFactory`.

Specifically, when we use `CompositeInstanceConfigFactory`, we can override the config namespace for which `InstaceConfig` reads its data from, using DI. Within Netflix this is overridden from `eureka` to `netflix.appinfo`. In `CustomAmazonInfoProviderInstanceConfigFactory`, however, this config namespace was previously hardcoded, causing test failures.

### DIFF
--- a/eureka-client-archaius2/src/main/java/com/netflix/appinfo/Ec2EurekaArchaius2InstanceConfig.java
+++ b/eureka-client-archaius2/src/main/java/com/netflix/appinfo/Ec2EurekaArchaius2InstanceConfig.java
@@ -47,6 +47,12 @@ public class Ec2EurekaArchaius2InstanceConfig extends EurekaArchaius2InstanceCon
         this.amazonInfoHolder = amazonInfoProvider;
     }
 
+    public Ec2EurekaArchaius2InstanceConfig(Config configInstance, Provider<AmazonInfo> amazonInfoProvider, String namespace) {
+        super(configInstance, namespace);
+        this.amazonInfoConfig = null;
+        this.amazonInfoHolder = amazonInfoProvider;
+    }
+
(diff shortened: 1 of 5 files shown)


eureka-client-archaius2/src/main/java/com/netflix/appinfo/Ec2EurekaArchaius2InstanceConfig.java
```diff
@@ -41,8 +41,8 @@ public Ec2EurekaArchaius2InstanceConfig(Config configInstance, AmazonInfoConfig
         this(configInstance, new Archaius2AmazonInfoConfig(configInstance), CommonConstants.DEFAULT_CONFIG_NAMESPACE, info, false);
     }
 
-    public Ec2EurekaArchaius2InstanceConfig(Config configInstance, Provider<AmazonInfo> amazonInfoProvider) {
-        super(configInstance, CommonConstants.DEFAULT_CONFIG_NAMESPACE);
+    public Ec2EurekaArchaius2InstanceConfig(Config configInstance, Provider<AmazonInfo> amazonInfoProvider, String namespace) {
```
spencergibb, line 44: What about a new constructor while the old one preserves the existing behavior?

fiibbb: From a pure backward compatibility perspective that's totally the right thing to do. I doubt this gets used by anybody outside of Netflix though. So is it really worth making the code harder to follow?

PR #1383, opened 2021-03-23
Update ResolverUtils::randomize

jebeaudet:
The method will now properly randomize a list with a different result on each invocation. Using a constant seed made the same list of `["a","b","c"]` always "randomized" to the same result on each invocation.

Since `SessionedEurekaHttpClient` is supposed to round robin through the different eureka endpoints at a configurable interval, having a randomize method that always returns the same value render this class useless.

Thanks


troshko111:
Thanks!

PR #1404, opened 2021-06-15
Upgrade jacksonVersion and xstream.

AmitAmar:


troshko111:
This is now a no-op after the rebase on top of the origin/master, do you want to close it now?

PR #1405, opened 2021-06-16
refactor: use diamond operator

gregorriegler:
Co-authored-by: Moderne <team@moderne.io>


troshko111 (changes requested):
Thanks for the PR!

One change request: please revert diamond operator usages where the expression no longer has the type specified on the left hand side (i.e. the type is completely erased from the expression).

Good: `Set<String> cnamesSet = new TreeSet<~String~>();`
Since the type appears on the left, the rhs is redundant.

Bad: 

```
heartbeatExecutor = new ThreadPoolExecutor(
                    1, clientConfig.getHeartbeatExecutorThreadPoolSize(), 0, TimeUnit.SECONDS,
                    new SynchronousQueue<>(),
                    new ThreadFactoryBuilder()
                            .setNameFormat("DiscoveryClient-HeartbeatExecutor-%d")
                            .setDaemon(true)
```

```
this.applications = new ConcurrentLinkedQueue<>();
```

Here the type does not appear on the lhs, and erasing from the rhs make it impossible to tell what it is.

gregorriegler:
Ok, I went through the changes end reverted the ones where no type-info was left.

troshko111 (approved):

PR #1409, opened 2021-06-21
fix jackson version - WS.

AmitAmar:


troshko111:
Jackson is a big transitive dependency, I'm inclined to move slower here than 2 branches up. This also needs some validation to make sure nothing breaks. What's your motivation to move to this version?

AmitAmar:
Hi @troshko111 .

I'm inclined to move faster in Jackson versions because there are a lot of vulnerabilies in jackson.

We are using WhiteSource and it seems we need to upgrade to the latest version of Jackson.

In our system we already done this upgrade, we want to upgrade Jackson in Eureka as well.

Let me know if you see any problem with that. you want to wait some time for testing?

BR,

Amit.

spencergibb:
I'd expect a change like this to land in a new minor (1.11.x)

AmitAmar:
This is your decision of course,
I think we should fix it soon as possible but if it risky from your side I'll close the PR.

Thanks,

Amit.

troshko111:
> I'm inclined to move faster in Jackson versions because there are a lot of vulnerabilities in Jackson.

Jackson backports vuln fixes, is there anything in particular?

> In our system we already done this upgrade, we want to upgrade Jackson in Eureka as well.

You can always force a different Jackson version in your built, irrespecutful of what Eureka version is, right?

> I think we should fix it soon as possible but if it risky from your side I'll close the PR.

Again I agree we want vulns patches, but based on my understanding 2.10 is still being security patched:

```
2.10: nominally open branch but unlikely to get new full patches beyond 2.10.5
Micro-patches possible for critical fixes like security patches```

from https://github.com/FasterXML/jackson/wiki/Jackson-Releases

We can move up to `2.11` once security patches stop coming to `2.10`, that's what we did before, makes sure we're both patched and minimally depending.

AmitAmar:
Hi,

After checking it again you are right and it seems that 2.10 is safe to use.
I'll upgrade it again in the future if it will be necessary.

Thanks again and have a nice day,
Amit.

PR #1412, opened 2021-06-26
Refactor: use diamond operator

schrosolver:


troshko111 (approved):

troshko111:
Just FYI that these sort of changes are best to do in one go, I'd like to avoid a large number of PRs with the same logical change as it negatively affects the history. Thanks!

schrosolver:
> Just FYI that these sort of changes are best to do in one go, I'd like to avoid a large number of PRs with the same logical change as it negatively affects the history. Thanks!

Thanks! Will keep this in mind. Checked the last merged PR and saw there were few misses, so thought will add them too.

PR #1415, opened 2021-07-26
refactor: JUnit Jupiter migration from JUnit 4.x

jkschneider:
👋 This PR will get you most of the way there to JUnit 5 from JUnit 4. We'd just need to migrate `DiscoveryClientResource` and `SimpleEurekaHttpServerResource` which don't have direct analogs in JUnit 5. Shouldn't be very difficult to complete from here, but hopefully this takes away most of the grueling work.

Co-authored-by: Moderne <team@moderne.io>

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -51,7 +51,7 @@ allprojects {
 
         // test deps
         jetty_version = '7.2.0.v20101020'
-        junit_version = '4.11'
+        junit_version = '5.7.2'
         mockitoVersion = '1.10.19'
         mockserverVersion = '3.9.2'
     }
--- a/eureka-client-archaius2/src/test/java/com/netflix/appinfo/Ec2EurekaArchaius2InstanceConfigTest.java
+++ b/eureka-client-archaius2/src/test/java/com/netflix/appinfo/Ec2EurekaArchaius2InstanceConfigTest.java
@@ -2,8 +2,8 @@ package com.netflix.appinfo;
 
 import com.netflix.archaius.config.MapConfig;
 import com.netflix.discovery.util.InstanceInfoGenerator;
-import org.junit.Before;
-import org.junit.Test;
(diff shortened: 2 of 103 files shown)


troshko111:
This looks pretty harmless / low risk, but at the same time I wonder what the benefits are?

jkschneider:
Hi @troshko111. Just trying to help so Eureka isn't built on a stale unsupported core testing framework? Feel free to close if you wish.

troshko111:
I went over the PR and it looks good to me, I don't really see any issue with using `JUnit 4` regardless of its support status as it's not part of the running service, so there's no risk of CVEs and such, basically it does not matter how (un)supported it is, it works just fine. 

Given the number of new tests written, and potential benefit of using the new `JUnit` version there, such effort is unlikely to ever pay off, but it's already done so why not. Re-running and merging if all passes.

troshko111:
Oh I see what you meant in the original PR message, this _does require changes before it can land_ (it does not compile as-is due to those two types not being migrated). In this case I can keep it open in case you / anyone else wants to drive it past the finish line, but I don't remember anyone expressing interest before, so if you won't be getting it into a mergeable state, I will probably close.

jkschneider:
Feel free to close. It's a disappointing attitude toward the codebase, to be honest, and adds to the feeling that Eureka may be barely maintained. I thought having done most of the heavy lifting, it would be an easy win.

troshko111:
It's unfortunate you feel this way, but in general we prefer _compiling PRs_ (all checks must pass), if this was a high value change I could see us taking it from here, but per my earlier response the value is debatable plus this does not compile, so closing. In the future, one easy way to avoid this situation is to open an issue first, have a discussion and alignment, and only then proceed with heavy lifting, once it's clear the change/scope is agreed upon.

PR #1417, opened 2021-08-05
Enable setting a EurekaClient instance in DiscoveryManager

asibross:


eureka-client/src/main/java/com/netflix/discovery/DiscoveryManager.java
```diff
@@ -42,6 +42,7 @@
 public class DiscoveryManager {
     private static final Logger logger = LoggerFactory.getLogger(DiscoveryManager.class);
     private DiscoveryClient discoveryClient;
+    private EurekaClient eurekaClient; // Might be a proxy to DiscoveryClient
```
troshko111, line 45: Can you call it something like `clientOverride`?

eureka-client/src/main/java/com/netflix/discovery/DiscoveryManager.java
```diff
@@ -58,6 +59,10 @@ public void setDiscoveryClient(DiscoveryClient discoveryClient) {
         this.discoveryClient = discoveryClient;
     }
 
+    public void setEurekaClient(EurekaClient eurekaClient) {
```
troshko111, line 62: Same here (`setClientOverride`).

troshko111 (approved):

eureka-client/src/main/java/com/netflix/discovery/DiscoveryManager.java
asibross: Done

eureka-client/src/main/java/com/netflix/discovery/DiscoveryManager.java
asibross: Done

PR #1420, opened 2021-08-23
Fix xstream security issue CVE-2021-39139

wcc526:
Fix xstream security issue CVE-2021-39139

https://x-stream.github.io/security.html


troshko111:
Thanks, merged.

Issue #1424, opened 2021-09-10
Why Eureka Client used Synchronized Map for Instance Metadata


lsc1943:
We are using eureka client 1.x, and recently we found some threads were locked by "java.util.Collections$SynchronizedMap" which was happened while getting  data from instanceInfo metadata under a high workload. It seems impact the performance. For the instance metadata, it's returned as a concurrentHashmap from server, but after deserialization, it turned to be a SynchronizedMap.

The below code is under EurekaJacksonCodec class, any reason why we are using synchronizedMap here?
 builder.setMetadata(metadataMap == null ? Collections.emptyMap() : Collections.synchronizedMap(metadataMap));

troshko111:
> We are using eureka client 1.x, and recently we found some threads were locked by "java.util.Collections$SynchronizedMap" which was happened while getting data from instanceInfo metadata under a high workload. It seems impact the performance.

While possible, I do wonder under which scenarios you'd get high contention accessing individual instance metadata in the client? A flame graph would help if you have one handy.

> For the instance metadata, it's returned as a concurrentHashmap from server, but after deserialization, it turned to be a SynchronizedMap.

Well it's returned as a map, that's the only contract. I don't think there's an expectation of any sort of type information to be preserved (beyond primitive, array and object) via JSON/XML across different languages, or even within Java. 

> The below code is under EurekaJacksonCodec class, any reason why we are using synchronizedMap here?
builder.setMetadata(metadataMap == null ? Collections.emptyMap() : Collections.synchronizedMap(metadataMap));

It's hard to say now, but I'd guess for simplicity's sake, I see no harm swapping this for any other thread safe implementation, if we can show a clear benefit.

lsc1943:
@troshko111 , thank you for your comment. Our scenario is a little tricky, we have a gateway and hundreds of instances for an upstream service, once a request came in, the gateway will try to filter the instances by iterate the instanceInfo list and get metatdata for each instanceInfo, and when the concurrent workload is high, we saw the locked instances of "java.util.Collections$SynchronizedMap" from JFR as many concurrent requests may try to get the metadata at the same time for a single instanceInfo.
For the (builder.setMetadata) here, why do we need a thread safe implementation here, under any condition the eureka client will try to build or modify the metadata concurrently?
Thank you!

PR #1426, opened 2021-09-27
Remove redundant metadata synchronization (#1424)

troshko111:


spencergibb:
Interesting, I assume this is a perf issue?

troshko111:
@spencergibb yeah a user reported contention under load, and I honestly can't see any evidence a thread safe map is required (per contract or per actual usage). LMK what your thoughts are on this.

spencergibb:
I haven't either and can't guess any future issues

troshko111:
After a conversation with @tcellucci we figured that it should not be a breaking change barring use cases where users mutate the metadata of instances from the registry (a bad idea in general) without explicitly syncing access (an even worse idea). I will merge this but keep an open mind about potential effects, if they prove to be widespread, it may get reverted or put behind a property (sync map by default). For now I don't think it warrants to be behind a prop, it's below my safety threshold.

Issue #1427, opened 2021-09-29
It is easy to trigger null pointer exception


wcnnkh:
https://github.com/Netflix/eureka/blob/4d36f04849541df6c6b8738622544d1b26dfe8d4/eureka-core/src/main/java/com/netflix/eureka/util/StatusUtil.java#L59-L61

https://github.com/Netflix/eureka/blob/4d36f04849541df6c6b8738622544d1b26dfe8d4/eureka-core/src/main/java/com/netflix/eureka/util/StatusInfo.java#L105-L107

wcnnkh:
If the first step is not executed during construction, the method calling the second step will throw a null pointer exception. Can you change it to:
`public boolean isHealthy() { 
     return  isHeathly == null? false:isHeathly.booleanValue(); 
 } `

troshko111:
SGTM, feel free to send a PR.

wcnnkh:
Push to that branch?Master?

troshko111:
You won't be able to push to origin/master, need to fork and send a PR which I see you did.

PR #1428, opened 2021-09-30
Fix null pointer exception(#1427)

wcnnkh:


troshko111 (approved):

troshko111:
Test failing, does it pass for you locally?

wcnnkh:
> Test failing, does it pass for you locally?

Not tested locally because I don't have a gradle environment. But why is such a simple modification wrong? I haven't seen that the error is related to my modification. I'll test it after I install the environment.

spencergibb:
No need to install Gradle, just run the wrapper at the root of the project

troshko111:
> Not tested locally because I don't have a gradle environment. But why is such a simple modification wrong? I haven't seen that the error is related to my modification. I'll test it after I install the environment.

Well good question, I admit that Gradle does an amazing job burying any useful information in the sea of output it produces, but after wasting some good 5m I managed to locate the failure:

```
com.netflix.eureka.util.StatusUtilTest > testGetStatusInfoUnsetHealth FAILED
    java.lang.AssertionError: Excpected NPE to be thrown when healthy threshold is not set
        at org.junit.Assert.fail(Assert.java:88)
        at com.netflix.eureka.util.StatusUtilTest.testGetStatusInfoUnsetHealth(StatusUtilTest.java:52)
```

Looks like there's a test asserting the existing behavior in `eureka-core/src/test/java/com/netflix/eureka/util/StatusUtilTest.java` method `testGetStatusInfoUnsetHealth`.

And apparently
`eureka-core/src/main/java/com/netflix/eureka/util/StatusUtil.java` does:

```
        // Only set the healthy flag if a threshold has been configured.
        if (peerEurekaNodes.getMinNumberOfAvailablePeers() > -1) {
            builder.isHealthy(upReplicasCount >= peerEurekaNodes.getMinNumberOfAvailablePeers());
        }
```

So there's a dependency on the existing behavior.

I'm inclined to decline the change based on this, unless you provide more information on how this affects any real use case / how to trigger this NPE.

PR #1433, opened 2021-11-08
Improve GRADLE build Performance

shisheng-1:
[Parallel builds](https://docs.gradle.org/current/userguide/multi_project_configuration_and_execution.html#sec:parallel_execution). This project contains multiple modules. Parallel builds can improve the build speed by executing tasks in parallel. We can enable this feature by setting `org.gradle.parallel=true`.

[File system watching](https://blog.gradle.org/introducing-file-system-watching). Since Gradle 6.5, File system watching was introduced which can help to avoid unnecessary I/O. This feature is the default since 7.0. For an older version, we can enable this feature by setting `org.gradle.vfs.watch=true`.

[Configuration on demand](https://docs.gradle.org/current/userguide/multi_project_configuration_and_execution.html#sec:configuration_on_demand). Configuration on demand tells Gradle to configure modules that only are relevant to the requested tasks instead of configuring all of them. We can enable this feature by setting `org.gradle.configureondemand=true`.

[gradle caching](https://docs.gradle.org/current/userguide/build_cache.html). Shared caches can reduce the number of tasks you need to execute by reusing outputs already generated elsewhere. This can significantly decrease build times. We can enable this feature by setting `org.gradle.caching=true`.

=====================
If there are any inappropriate modifications in this PR, please give me a reply and I will change them.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,1 +1,5 @@
 release.scope=patch
+org.gradle.caching = true
+org.gradle.parallel = true
+org.gradle.configureondemand = true
+org.gradle.vfs.watch = true
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,5 +1,4 @@
 release.scope=patch
 org.gradle.caching = true
 org.gradle.parallel = true
-org.gradle.configureondemand = true
 org.gradle.vfs.watch = true


gradle.properties
```diff
@@ -1 +1,5 @@
-release.scope=patch
\ No newline at end of file
+release.scope=patch
+org.gradle.caching = true
+org.gradle.parallel = true
+org.gradle.configureondemand = true
```
troshko111, line 4: This is incubating, so I would not enable just yet, can you please comment it out and explain we nay turn it out once it matures?

troshko111 (changes requested):

gradle.properties
shisheng-1: The configuration on demand feature is incubating. If projects are configured as what it asked, it would be a good configuration for project build.
I fully understand your concern about stability, so we can just try to remove it

troshko111: Thanks.

PR #1437, opened 2021-11-29
V1.7 comment

kismet2399:


troshko111 (changes requested):
1. Some context around the changes, ideally this should have been an issue first where we could discuss the proposed change, but it's too late for that, we can discuss now.
2. Rebase on top of `origin/master`.

PR #1464, opened 2022-08-14
Expose updateStatus and updateMetadata in DiscoveryClient

krutsko:
Hello 

Some background:
We use an external tool to manage Eureka Status and Metadata of the service via REST API (i.e. service registered as STARTING, and then external tool update status to UP/OUT_OF_SERVICE + manage some data in Metadata). 
Now, we are going to migrate to a new architure where the service itself will manager its State.
I tried to use `ApplicationInfoManager.getInstance().setInstanceStatus()` and `ApplicationInfoManager.getInstance().registerAppMetadata(metadata)`.
But it does not work for us (maybe because of old eureka server version) or reincarnation of bug https://github.com/Netflix/eureka/issues/1174
Anyway, it is not reliable to continue with this path, so we deciced to use REST API and make requests to server from instance itself. 

Proposal:
It would be very useful to re-use functionality of DiscoveryClient (cause it's already covers eureka server url discovery, retries, metrics etc.).

Would you consider to expose such functionality in DiscoveryClient?

### DIFF
--- a/eureka-client-jersey2/src/main/java/com/netflix/discovery/shared/transport/jersey2/AbstractJersey2EurekaHttpClient.java
+++ b/eureka-client-jersey2/src/main/java/com/netflix/discovery/shared/transport/jersey2/AbstractJersey2EurekaHttpClient.java
@@ -211,8 +211,7 @@ public abstract class AbstractJersey2EurekaHttpClient implements EurekaHttpClien
         try {
             Builder requestBuilder = jerseyClient.target(serviceUrl)
                     .path(urlPath)
-                    .queryParam("key", key)
-                    .queryParam("value", value)
+                    .queryParam(key, value)
                     .request();
             addExtraProperties(requestBuilder);
             addExtraHeaders(requestBuilder);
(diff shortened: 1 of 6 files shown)


mgtriffid:
@krutsko hi bro, did you maybe check if you actually have the same `InstanceInfo` - or maybe even `ApplicationInfoManager` - object in `DiscoveryClient` and what you get by invoking `getInstance()`? The behavior you describe is sus, looks like a bug. Could you please, just to confirm the described issue, invoke `getApplicationInfoManager()` on your instance of `DiscoveryClient` instead of calling static lookup `ApplicationInfoManager.getInstance()`? Sorry if the comment is irrelevant, just have some gut feeling and can't pass by.

krutsko:
hi @mgtriffid !
 
Yes , it is definitely a bug. The Status of local InstanceInfo (from `ApplicationInfoManager.getInstance()`) is different to what I see on Eureka Server Registry (e.g. when I query remote through `discoveryClient.getApplication().getInstances()`).

So, to sum up the flow: 
- service registered in STARTING
- `ApplicationInfoManager.getInstance().setInstanceStatus(UP)` - OK 
- `ApplicationInfoManager.getInstance().setInstanceStatus(OUT_OF_SERVICE)` - NOT OK (on remote status is UP , on local OUT_OF_SERVICE).

krutsko:
It seems, we misused OUT_OF_SERVICE status.
Found this comment which explains it https://github.com/spring-cloud/spring-cloud-netflix/issues/3941#issuecomment-848311134
So, we will try to use the correct flow (UP <-> DOWN). Closing this ticket.

PR #1469, opened 2022-10-03
Publish 2.x snapshots

spencergibb:


.github/workflows/nebula-snapshot.yml
```diff
@@ -29,7 +30,7 @@ jobs:
             ~/.gradle/wrapper
           key: ${{ runner.os }}-gradlewrapper-${{ hashFiles('gradle/wrapper/*') }}
       - name: Build
-        run: ./gradlew build snapshot
+        run: ./gradlew build snapshot -Prelease.version=2.0.0-SNAPSHOT
```
spencergibb, line 33: Since nebula will go find the latest version, it would end up with 1.10.18, so I force 2.0.0-SNAPSHOT.

howardyuan (approved):
LGTM

PR #1473, opened 2022-10-31
Fixme 2.0 patch1

spencergibb:
Creates new EurekaServerHttpClientFactory

This is basically EurekaServerHttpClients, but as an interface. Various constructors changed from taking EurekaHttpClient to EurekaServerHttpClientFactory.

Adds TransportClientFactories as required parameter to DiscoveryClient.

Removes deprecated DiscoveryClient constructors.


eureka-core/src/main/java/com/netflix/eureka/EurekaBootStrap.java
```diff
@@ -222,7 +223,11 @@ protected void initEurekaServerContext() throws Exception {
         EurekaMonitors.registerAllStats();
     }
 
-    protected abstract EurekaHttpClient getEurekaHttpClient();
+    protected AbstractDiscoveryClientOptionalArgs<?> getDiscoveryClientOptionalArgs() {
```
spencergibb, line 226: We should probably have a discussion about the name of this class `AbstractDiscoveryClientOptionalArgs` (or a refactor) as `TransportClientFactories` is no longer optional.

eureka-core-jersey3/src/main/java/com/netflix/eureka/transport/Jersey3RemoteRegionClientFactory.java
```diff
@@ -0,0 +1,117 @@
...
+import com.netflix.eureka.resources.ServerCodecs;
+import jakarta.inject.Inject;
+import jakarta.ws.rs.client.Client;
+import jakarta.ws.rs.core.Feature;
+import jakarta.ws.rs.core.MultivaluedHashMap;
+import org.glassfish.jersey.message.GZipEncoder;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class Jersey3RemoteRegionClientFactory implements TransportClientFactory {
```
spencergibb, line 25: This implementation came from the now deleted `EurekaServerHttpClients` https://github.com/Netflix/eureka/pull/1473/files#diff-bcd6df8c6a0b1c6ebf2d76ccb55219b364009609e9b57df874f9401f31faa2be

eureka-client-jersey3/src/main/java/com/netflix/discovery/shared/transport/jersey3/EurekaJersey3ClientImpl.java
```diff
@@ -232,7 +232,7 @@ class MyDefaultApacheHttpClient4Config extends ClientConfig {
 
                 if (systemSSL) {
                     cm = createSystemSslCM();
-                } else if (trustStoreFileName != null) {
+                } else if (trustStoreFileName != null && !trustStoreFileName.trim().isEmpty()) {
```
spencergibb, line 235: Unsure why this was needed.

howardyuan: Not sure exactly. But it seems to be preventing the file name to be empty string which probably do any harm.

eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
```diff
@@ -228,34 +232,19 @@ void shutdown() {
...
-     * @deprecated use constructor that takes ApplicationInfoManager instead of InstanceInfo directly
-     */
-    @Deprecated
-    public DiscoveryClient(InstanceInfo myInfo, EurekaClientConfig config, AbstractDiscoveryClientOptionalArgs args) {
-        this(ApplicationInfoManager.getInstance(), config, args);
-    }
-
-    public DiscoveryClient(ApplicationInfoManager applicationInfoManager, EurekaClientConfig config) {
-        this(applicationInfoManager, config, null);
+    @SuppressWarnings("rawtypes")
```
howardyuan, line 235: Now TransportClientFactories is mandatory. This would break anyone who doesn't use SB cloud. Is there a way that we can construct a default TransportClientFactories for users and provide ctor option w/o TransportClientFactories? I know it's a major release, but I'm not sure how users would know about what to do to construct a TransportClientFactories?

eureka-core/src/main/java/com/netflix/eureka/registry/RemoteRegionRegistry.java
```diff
@@ -145,20 +140,22 @@ public RemoteRegionRegistry(EurekaServerConfig serverConfig,
         } catch (UnknownHostException e) {
             logger.warn("Cannot find localhost ip", e);
         }
-        EurekaServerIdentity identity = new EurekaServerIdentity(ip);*/
-        // FIXME 2.0
-        // discoveryApacheClient.addFilter(new EurekaIdentityHeaderFilter(identity));
+        EurekaServerIdentity identity = new EurekaServerIdentity(ip);
```
howardyuan, line 143: Should these lines still be here? There's no more `FIXME 2.0` so I'm not sure.

spencergibb: yeah, those are for the non-experimental client I still need to add back. If you look at the top of the comment block, there is a `FIXME: 2.0` still

eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
spencergibb: Not without merging jersey 3 modules. Spring cloud netflix with auto-configure the right thing. Right now there is no guice support, so until that happens no one can use it without spring cloud netflix anyway. If you do get guice support, then that would do the right thing.

PR #1479, opened 2022-12-01
fix failing test in TimeConsumingInstanceRegistryTest

sullis:


eureka-core/src/test/java/com/netflix/eureka/registry/TimeConsumingInstanceRegistryTest.java
```diff
@@ -139,10 +139,10 @@ public void execute() {
                         "There should be 45 instances in application - MYLOCAPP");
                 }
             }),
-            buildEvent(40, new SingleEvent.Action() {
+            buildEvent(60, new SingleEvent.Action() {
```
troshko111, line 142: Why does this increase? What's causing the delays?

troshko111 (changes requested):

PR #1489, opened 2023-03-06
Call size method on Application to avoid copying instance info

kilink:
DiscoveryClient called getInstancesAsIsFromEureka in a couple locations to get the number of instances for an application, when Application already has a method for this purpose that doesn't also incur the overhead of defensively copying.

Additionally, fix a concurrency bug in Application.size, as it was accessing the underlying instances field without synchronization.


kilink:
Closing this in favor of #1511 which targets 2.x.

PR #1504, opened 2023-06-15
Fix handling null messages during timeout evaluation.

OlgaMaciaszek:
Ported from https://github.com/Netflix/eureka/pull/1503.


OlgaMaciaszek:
Thanks, @howardyuan.

PR #1505, opened 2023-06-16
Add basic auth headers if userinfo present.

OlgaMaciaszek:
In Eureka `1.x` this is being done under the hood by Apache with `RequestTargetAuthentication`, added by the `DefaultHttpClient`, that is created through [this call](https://github.com/Netflix/eureka/blob/efc1cf6017f13f5f2491838bde894630a37dfd75/eureka-client/src/main/java/com/netflix/discovery/shared/transport/jersey/EurekaJerseyClientImpl.java#L52). 

This fixes https://github.com/spring-cloud/spring-cloud-netflix/issues/4161.

Additionally, I've renamed the client variable, as we no longer use the Apache Jersey Client and have removed unused constant.


spencergibb (approved):

OlgaMaciaszek:
Thanks, @howardyuan.

PR #1513, opened 2023-09-14
Add support for EIPs associated with secondary ENIs

brituck:
Currently, Eureka client only supports registering EIPs that are directly associated to an instance and available via the root `public-ipv4` IMDS key. This PR adds support to register an EIP that is attached via a secondary ENI. Due to the complexities of discovering and registering multiple public IPs for a single instance, we only register the first found public IP address. If an instance has both an EIP directly associated and and EIP associated via an ENI, the directly associated EIP will be registered. If the instance only has a EIP(s) associated via a ENI(s), the first found EIP will be registered. Additional use cases (i.e., registering multiple public IPv4 addresses) will not be supported.

### DIFF
--- a/eureka-client/src/main/java/com/netflix/appinfo/AmazonInfo.java
+++ b/eureka-client/src/main/java/com/netflix/appinfo/AmazonInfo.java
@@ -22,6 +22,7 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.net.MalformedURLException;
 import java.net.URL;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.regex.Matcher;
@@ -63,6 +64,13 @@ public class AmazonInfo implements DataCenterInfo, UniqueIdentifier {
         availabilityZone("availability-zone", "placement/"),
         publicHostname("public-hostname"),
         publicIpv4("public-ipv4"),
+        macs("macs", "network/interfaces/"), // macs declared above public-ipv4s so will be found before publicIpv4s (where it is needed)
+        publicIpv4s("public-ipv4s", "network/interfaces/macs/") {
+            @Override
+            public URL getURL(String prepend, String mac) throws MalformedURLException {
+                return new URL(AWS_METADATA_URL + this.path + mac + "/" + this.name);
+            }
+        },
         ipv6("ipv6"),
         spotTerminationTime("termination-time", "spot/"),
         spotInstanceAction("instance-action", "spot/"),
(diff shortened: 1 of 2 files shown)


error0702:
这是来自QQ邮箱的假期自动回复邮件。你好，我最近正在休假中，无法亲自回复你的邮件。我将在假期结束后，尽快给你回复。

eureka-client/src/test/java/com/netflix/discovery/InstanceInfoReplicatorTest.java
```diff
@@ -76,7 +76,7 @@ public void testOnDemandUpdateRateLimiting() throws Throwable {
         assertTrue(replicator.onDemandUpdate());
         Thread.sleep(10);  // give some time for execution
         assertFalse(replicator.onDemandUpdate());
-        Thread.sleep(10);
+        Thread.sleep(1000);
```
drobertduke, line 79: Was this flakey?

eureka-core/src/test/java/com/netflix/eureka/GzipEncodingEnforcingFilterTest.java
```diff
@@ -89,7 +89,6 @@ public void testForceGzip() throws Exception {
     @Test
     public void testForceGzipOtherHeader() throws Exception {
         noneGzipRequest();
-        when(request.getHeader("Test")).thenReturn("ok");
```
drobertduke, line 92: What changed here?

eureka-test-utils/src/main/java/com/netflix/discovery/shared/transport/ClusterSampleData.java
drobertduke, line 1: Why did the ClusterSampleData change?

eureka-client/src/test/java/com/netflix/discovery/InstanceInfoReplicatorTest.java
brituck: Yeah. I'm guessing both those `onDemandUpdate()` calls didn't have time to finish so the rate limiter counts never triggered. Increasing the sleep give them time to finish and the counter to increment.

eureka-core/src/test/java/com/netflix/eureka/GzipEncodingEnforcingFilterTest.java
brituck: Updating the Mockito version to 3.4.0 changes the strictness of mock checks. That call `request.GetHeader("Test")` was never called during the test so it fails with `unnecessarystubbingexception`. It's not necessary for the test so I removed it.

eureka-test-utils/src/main/java/com/netflix/discovery/shared/transport/ClusterSampleData.java
brituck: In the update to Mockito 3.4.0 the ArgumentMatcher for `anyString()` changed from accepting "any string or null" to "any string". In the `testStatusUpdate()` and `testDeleteStatusOverride()` tests in [PeerReplicationResourceTest](https://github.com/Netflix/eureka/blob/master/eureka-core/src/test/java/com/netflix/eureka/resources/PeerReplicationResourceTest.java#L107), the provided parameter for `status` in the downstream `instanceResource.statusUpdate()` call is `null`, which results in the Mock not matching and our response returning null.

There are two solutions to this:
1. Refactor the fields in the returned `ReplicationInstance` (current change in this PR) to correctly capture the intent of the `statusUpdate()` and `deleteStatusUpdate()` calls.
2. Change the mock expression for `statusUpdate()` and `deleteStatusUpdate()` to `isNull()` for the parameters that return `null`.

I prefer the first option as I believe it's the more correct change, but I'm happy to go with 2) if we want to minimize the changes in this PR.

PR #1523, opened 2023-11-29
replace servo with spectator in the client

joshgord:

PR #1526, opened 2024-01-05
replace servo with spectator

joshgord:
This is a port of this PR to the 2.x branch https://github.com/Netflix/eureka/pull/1523

In general when trivial I changed direct usages of guava with their replacements.  I did leave guava in the eureka-core project because we'll have to reimplement some caches.


eureka-core/src/main/java/com/netflix/eureka/util/EurekaMonitors.java
```diff
@@ -87,11 +86,11 @@ private EurekaMonitors(String name, String description) {
         }
     }
 
-    @com.netflix.servo.annotations.Monitor(name = "count", type = DataSourceType.COUNTER)
```
joshgord, line 90: This seems like a curiously named metric "count", is there something that gets added with the servo Monitor that allows us to disambiguate such a generically named id @brharrington?

eureka-core/src/main/java/com/netflix/eureka/util/batcher/AcceptorExecutor.java
```diff
@@ -106,35 +104,38 @@ class AcceptorExecutor<ID, T> {
...
         this.acceptorThread.start();
 
-        final double[] percentiles = {50.0, 95.0, 99.0, 99.5};
-        final StatsConfig statsConfig = new StatsConfig.Builder()
-                .withSampleSize(1000)
-                .withPercentiles(percentiles)
-                .withPublishStdDev(true)
-                .build();
-        final MonitorConfig config = MonitorConfig.builder(METRIC_REPLICATION_PREFIX + "batchSize").build();
-        this.batchSizeMetric = new StatsTimer(config, statsConfig);
```
joshgord, line 116: I wasn't sure what to do with StatsTimer and the docs just say it's "not supported".  Seems like PercentileTimer is a decent replacement

https://netflix.github.io/atlas-docs/spectator/lang/java/servo-migration/#statstimer

eureka-core/build.gradle
```diff
@@ -10,7 +10,8 @@ dependencies {
     api "com.amazonaws:aws-java-sdk-route53:${awsVersion}"
     api "jakarta.servlet:jakarta.servlet-api:${servletVersion}"
     api 'jakarta.inject:jakarta.inject-api:2.0.1'
-    api 'com.thoughtworks.xstream:xstream:1.4.19'
+    api 'com.thoughtworks.xstream:xstream:1.4.20'
+    implementation 'com.google.guava:guava:33.0.0-jre'
```
joshgord, line 14: Since this PR is about servo->spectator I'm leaving the guava dependency in place (but updating to the latest version), I can follow up with another PR that will replace the guava usages

eureka-client/build.gradle
```diff
@@ -12,7 +12,10 @@ dependencies {
     api 'jakarta.ws.rs:jakarta.ws.rs-api:3.0.0'
     api 'jakarta.inject:jakarta.inject-api:2.0.1'
     api 'jakarta.annotation:jakarta.annotation-api:2.1.0'
-    api "com.netflix.servo:servo-core:${servoVersion}"
+    api "com.netflix.spectator:spectator-api:${spectatorVersion}"
+    api "org.slf4j:slf4j-api:${slf4jVersion}"
+    implementation "com.netflix.netflix-commons:netflix-eventbus:0.3.0"
```
brharrington, line 17: `netflix-commons` is also legacy stuff that brings in dependencies that can be problematic, e.g. [servlet-api](https://github.com/Netflix/netflix-commons/pull/24).

eureka-core/src/main/java/com/netflix/eureka/util/EurekaMonitors.java
brharrington: It would have a dimenion `class=EurekaMonitors`. See: https://github.com/Netflix/servo/blob/master/servo-core/src/main/java/com/netflix/servo/monitor/Monitors.java#L264

eureka-core/src/main/java/com/netflix/eureka/util/batcher/AcceptorExecutor.java
brharrington: If percentiles are really needed, then `PercentileTimer` or `PercentileDistributionSummary` would need to be used. Either way, it wouldn't be fully backwards compatible, so the user may need to tune their queries.

I'm not sure sure how this metric is used. The name is `batchSize` which doesn't seem like it would be measuring time, so `PercentileDistributionSummary` may be more appropriate. It was using a timer though.

brharrington (approved):

eureka-core/src/main/java/com/netflix/eureka/util/EurekaMonitors.java
joshgord: Got it, should I add this dimension to the metrics?  Will there be any gotchas about emitting that dimension manually?

brharrington: For better compatibility with the previous data it would be good to add it explicitly. There shouldn't be any issues with adding it directly.

joshgord: Will go ahead and do that to attempt to avoid any external changes in the metrics.  Thanks @brharrington

eureka-client/build.gradle
joshgord: That looks like an issue with the cherry pick from my old commit, this is already present in the build.gradle, I agree netflix-common is toxic but for now we'll leave that for future pull request 😄

joshgord: I changed it from api->implementation which should help a little bit for now

eureka-core/src/main/java/com/netflix/eureka/util/batcher/AcceptorExecutor.java
joshgord: Changed this one to a PercentileDistributionSummary but left another one as percentiletimer but noted that something is going to break if people are using this metric

eureka-client/src/main/java/com/netflix/discovery/util/ServoUtil.java
```diff
@@ -16,54 +16,58 @@
...
+    return timer.clock().monotonicTime();
+  }
+
+  public static void record(@Nonnull Timer timer, long startTime) {
+    timer.record(time(timer) - startTime, TimeUnit.NANOSECONDS);
+  }
 
-    private ServoUtil() {
-    }
+  public static <T> T monitoredValue(@Nonnull String name, @Nonnull T obj,
```
joshgord, line 46: Added this helper value that adds the class dimension to the metric by default

PR #1566, opened 2024-12-23
Use lock for instance status update.

OlgaMaciaszek:
This is related to https://github.com/spring-cloud/spring-cloud-netflix/issues/4094 as an attempted fix.


eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
```diff
@@ -1381,15 +1382,27 @@ void refreshInstanceInfo() {
         applicationInfoManager.refreshLeaseInfoIfRequired();
 
         InstanceStatus status;
-        try {
-            status = getHealthCheckHandler().getStatus(instanceInfo.getStatus());
-        } catch (Exception e) {
-            logger.warn("Exception from healthcheckHandler.getStatus, setting status to DOWN", e);
-            status = InstanceStatus.DOWN;
-        }
+        ReentrantLock instanceStatusUpdateLock = new ReentrantLock();
```
spencergibb, line 1385: Should probably be a field if I understand correctly.

OlgaMaciaszek: Right. No idea why I put it this way.

OlgaMaciaszek: Thanks for the catch.

eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
```diff
@@ -1381,15 +1383,26 @@ void refreshInstanceInfo() {
         applicationInfoManager.refreshLeaseInfoIfRequired();
 
         InstanceStatus status;
-        try {
-            status = getHealthCheckHandler().getStatus(instanceInfo.getStatus());
-        } catch (Exception e) {
-            logger.warn("Exception from healthcheckHandler.getStatus, setting status to DOWN", e);
-            status = InstanceStatus.DOWN;
-        }
+        if (instanceStatusUpdateLock.tryLock()) {
```
howardyuan, line 1386: I'm not sure if we need the fix here? Since both threads will call healthcheck to find latest status, unless healthcheck itself is flipping, the status would be the same? Maybe there is a race in update sequence, but the value would be consistent?

OlgaMaciaszek: Hi @howardyuan, thanks for the comment. You're right.  I've taken another look at it and I now think the approach to take could be to have a `compareAndSet` logic while changing  `InstanceInfo.InstanceStatus`  within `DiscoveryClient#refreshInstanceInfo` to prevent a situation where the status has been changed by the other thread between getting original status from the field and setting its transformed value there. Could be done by changing that field into an actual `AtomicReference or, to remain consistent with the current class implementation and avoid having to change/ add too many methods, keeping it as a volatile field and adding a synchronized method in that class that would first compare the expected field value with the actual one and calling that new method within DiscoveryClient#refreshInstanceInfo. Let me know what you think.

howardyuan: I think your suggestions of comparing before setting should be safe.

OlgaMaciaszek: Will work on it soon.

OlgaMaciaszek:
Closing in favour of https://github.com/Netflix/eureka/pull/1569.

PR #1609, opened 2026-01-12
perf: Jackson Codec allocation improvements

jasonk000:
Two changes, (1) the getText toUpperCase call gets a slight transform to avoid a lambda capture. More significantly, we (2) introduce a mutable ArrayCharBuffer and StringCharBuffer to avoid repeated allocations of the buffer themselves. The buffers are only used during intermediate calls, and, the class is clearly marked as not-thread-safe and is attached to independent ObjectReader.

### DIFF
--- a/eureka-client/src/main/java/com/netflix/discovery/util/DeserializerStringCache.java
+++ b/eureka-client/src/main/java/com/netflix/discovery/util/DeserializerStringCache.java
@@ -207,11 +207,12 @@ public class DeserializerStringCache implements Function<String, String> {
      * 
      * @param jp
      * @param cacheScope
+     * @param transform optional transform function applied to the parser text on cache miss
      * @return a possibly interned String
      * @throws IOException
      */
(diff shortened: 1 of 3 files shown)


eureka-client/src/main/java/com/netflix/discovery/util/DeserializerStringCache.java
```diff
@@ -483,4 +355,179 @@ public String consume(BiConsumer<CharBuffer, String> valueConsumer) {
...
+            return hash;
+        }
+
+        @Override
+        public boolean equals(Object other) {
+            if (other instanceof CharBuffer) {
+                CharBuffer otherBuffer = (CharBuffer) other;
+                if (otherBuffer.length() == length && otherBuffer.variant() == variant) {
+                    OfInt otherText = otherBuffer.chars();
+                    for (int i = offset; i < offset + length; i++) {
```
fvallenilla, line 403: Nice catch here

fvallenilla (approved):

PR #1616, opened 2026-01-20
backport - feat: Add metrics to count VIP address lookups in DiscoveryClient

jasonk000:
This adds simple counter metrics for the numerous discovery lookup calls a client may issue.

backport of #1621

### DIFF
--- a/eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
+++ b/eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
@@ -129,6 +129,16 @@ public class DiscoveryClient implements EurekaClient {
     private final Timer FETCH_REGISTRY_TIMER = SpectatorUtil.timer(PREFIX + "FetchRegistry", DiscoveryClient.class);
     private final Counter REREGISTER_COUNTER = SpectatorUtil.counter(PREFIX + "Reregister", DiscoveryClient.class);
 
+    // Lookup counters
+    private final Counter LOOKUP_GET_APPLICATION = SpectatorUtil.counter(PREFIX + "Lookup", "getApplication", DiscoveryClient.class);
+    private final Counter LOOKUP_GET_APPLICATIONS = SpectatorUtil.counter(PREFIX + "Lookup", "getApplications", DiscoveryClient.class);
+    private final Counter LOOKUP_GET_APPLICATIONS_FOR_A_REGION = SpectatorUtil.counter(PREFIX + "Lookup", "getApplicationsForARegion", DiscoveryClient.class);
+    private final Counter LOOKUP_GET_INSTANCES_BY_ID = SpectatorUtil.counter(PREFIX + "Lookup", "getInstancesById", DiscoveryClient.class);
+    private final Counter LOOKUP_GET_INSTANCES_BY_VIP = SpectatorUtil.counter(PREFIX + "Lookup", "getInstancesByVipAddress", DiscoveryClient.class);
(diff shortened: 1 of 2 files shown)


eureka-client/src/main/java/com/netflix/discovery/util/SpectatorUtil.java
```diff
@@ -100,6 +100,14 @@ public static Counter counter(@Nonnull String name, @Nullable String id,
     return Spectator.globalRegistry().counter(name, tags(id, clazz, extraTags));
   }
 
+  /**
+   * Creates a counter with varargs string tags (key-value pairs).
+   * Example: counter("myMetric", "key1", "value1", "key2", "value2")
+   */
+  public static Counter counter(@Nonnull String name, String... tags) {
```
fvallenilla: Why not use the existing method?

eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
```diff
@@ -696,17 +700,28 @@ public List<InstanceInfo> getInstancesByVipAddress(String vipAddress, boolean se
             if (null == applications) {
                 logger.debug("No applications are defined for region {}, so returning an empty instance list for vip "
                         + "address {}.", region, vipAddress);
+                recordVipAddressLookup(vipAddress, secure);
```
fvallenilla: There are a lot of methods to access the eureka client data, I'm not sure we want to instrument this narrowly

For example, the EVCache client fetches by application:
https://sourcegraph.netflix.net/github.netflix.net/corp/github.com_netflix_evcache@22ef2342a60a1cd059f7b509b729e63bdc708225/-/blob/evcache-client/src/main/java/com/netflix/evcache/pool/eureka/EurekaNodeListProvider.java?L68-71

Broader search:
https://sourcegraph.netflix.net/deepsearch/91754703-48d8-48f9-8736-4ea026ae4704

fvallenilla: An alternative could be to add a metric with a tag for each access method, and avoid tagging the vip/app that was queried

eg `name=lookup`, `tags={'id': 'getInstancesByVipAddress'}`

jasonk000: > There are a lot of methods to access the eureka client data, I'm not sure we want to instrument this narrowly

Good point, thanks, I'd totally missed that and assumed externals would be configured via vip.

> An alternative could be to add a metric with a tag for each access method, and avoid tagging the vip/app that was queried

A question here - do you see merit in adding a tag for the requested vip/app? It seems like we'd want that so we can discover which endpoints it's querying to hone in on a change. However if there's a better way to solve that problem, we would not need the extra detail and it would definitely simplify setup.

eureka-client/src/main/java/com/netflix/discovery/util/SpectatorUtil.java
jasonk000: It's a micro-opt that avoids the extra wrap/unwrap that `tags(..)` does. I can use the existing path and that likely makes sense. Nb we cannot cache these early-up unless we drop the 'vip' tag, which we can discuss in below comment.

eureka-client/src/main/java/com/netflix/discovery/DiscoveryClient.java
jasonk000: Ok, I have simplified this now to look more like the existing metrics, dropped the vip tag, and covered all of the entry points.

vilieva (approved):

PR #1619, opened 2026-02-01
Refactor InternalPrefixedConfig to use StringBuilder for namespace construction

### LINKED ISSUE #1606: Refactor namespace prefix construction to avoid String concatenation in loop
### Description

While reviewing `InternalPrefixedConfig`, I noticed that the namespace prefix is built using `String` concatenation inside a loop.

This works correctly, but repeated string concatenation can be inefficient and less readable, especially when the intent is to build a prefix string.

### Proposed Improvement

Refactor the constructor to:
- Avoid `String` concatenation inside loops
- Use `StringBuilder` or a stream-based approach
- Improve readability while keeping behavior unchanged

### Scope

- Internal refactor only
- No functional or behavioral changes

### Question

Would it be okay if I work on this refactoring and open a Pull Request?

omer-cengel:
### Summary
Refactored InternalPrefixedConfig to avoid String concatenation inside a loop by using StringBuilder.

### Problem
The constructor of InternalPrefixedConfig was building the namespace prefix using String concatenation (+=) inside a loop, which is less efficient due to repeated String object creation.

### Changes Made
- Replaced String concatenation with StringBuilder in the constructor.
- Preserved existing behavior for null, empty, and dotted namespace values.
- Added unit tests to cover edge cases and ensure identical behavior.

### Result
The code is more efficient, and clearer, with no behavioral changes.

### Fixes
Fixes #1606

### DIFF
--- a/eureka-client-archaius2/src/main/java/com/netflix/discovery/internal/util/InternalPrefixedConfig.java
+++ b/eureka-client-archaius2/src/main/java/com/netflix/discovery/internal/util/InternalPrefixedConfig.java
@@ -1,6 +1,7 @@
 package com.netflix.discovery.internal.util;
 
 import com.netflix.archaius.api.Config;
+import org.apache.commons.lang.StringUtils;
 
 import java.util.Iterator;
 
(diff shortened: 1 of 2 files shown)


marconi-prog (approved):
LGTM 👍

Clean refactor, behavior preserved, and tests added.

kimthienlang (approved):
👌

